PR #20, opened 2019-05-10
Save tapestry for a user in the database when any changes are made

### LINKED ISSUE #11: Save tapestry for a user in the database when any changes are made
**Tapestry Permissions**

* User must have the correct permission to carry out tasks
  * Admin / Editor: Full access to add, edit (any), delete (any)
  * Author: Access to add, edit (own), delete (own)
  * Others: Read only (if tapestry is public or user has specifically been given access)
* In the view page, only load tapestry if the user has the permission to view that tapestry
* In the edit page, determine whether to show editor tools or not based on permission of user for that given tapestry
* Ensure Wordpress permissions are set up correctly so that: 
  * Only users with add permission are allowed add content 
    * This includes authors, admins, and editors
  * Users only can list content that they have access to at least read
  * Users can only edit / delete a tapestry if they have the permission for it
* In the backend ajax functionality, check current user role before allowing CRUD operations

**Saving tapestries itself should user native WP functionality and database structure**

* New tapestries are saved as custom post types
* New nodes are saved as custom post types and a reference to them is saved in an array as the postmeta (named nodes). The structure will be as follows:
  * An array of postmeta IDs, each representing a tapestry-node relationship
  * Each tapestry-node postmeta will have:
    * post id (for the node)
    * node title (leave empty to keep the same as the title of the custom post type)
    * node coordinates in tapestry

anhduc130:
Fixes #11

Add controller implementation.


anhduc130:
The PR is ready for review. I have tested the endpoints for saving and retrieving the tapestry.

- As we discussed, each tapestry will contain an array of meta IDs that refers to another postmeta that represents a node as followed:
```
{'post_id': 1234, 'title': 'Some Title', 'coords': {'x': 100, 'y': 100}}
```
- Each tapestry node still has its own postmeta as before.

Sample data:
image

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,163 @@
...
+        return $nodes;
+    }
+
+    /**
+     * Retrieve Tapestry post
+     * 
+     * @param type @postId The postId to be retrieved
+     * @return type Tapestry post data
+     */
+    public function getTapestry($postId) {
```
wynnset: This function should be in the PR relating to loading tapestry, not here

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,163 @@
...
+            'post_id' => $nodePostId,
+            'title' => $node->title,
+            'coordinates' => (object) array(
+                'x' => $node->fx,
+                'y' => $node->fy
+            )
+        );
+    }
+    
+    private function getTapestryById($postId) {
```
wynnset: Similarly, this is for loading - not for saving, so it doesn't belong in this PR.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,163 @@
...
+        return wp_insert_post(array(
+            'post_type' => $postType,
+            'post_status' => $postStatus,
+            'post_title' => $postTitle
+        ), true);
+    }
+
+    // TODO: Remove this when we can build a tapestry from scratch
+    // HACK - create a new links array that works with new IDs
+    private function getNewLinks($oldLinks, $nodes) {
```
wynnset: Similarly, this should also be moved to the other PR.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
anhduc130: I removed it.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
anhduc130: I removed it.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
anhduc130: I removed it.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,110 @@
...
+     * @return WP_Error if postId is invalid
+     */
+    public function updateTapestryNodes($nodes, $postId = null) {
+        if (!isset($postId))
+            return $this->throwsError('INVALID_POST_ID');
+        $this->updateNodes($nodes, $postId);
+        return $nodes;
+    }
+
+    private function throwsError($code) {
```
wynnset: Please move all the helper functions to the bottom.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,110 @@
...
+        $ERROR = (object) self::ERRORS[$code];
+        return new WP_Error($code, $ERROR->MESSAGE, $ERROR->STATUS);
+    }
+
+    private function getNodeIds($nodes) {
```
wynnset: Please move all the helper functions to the bottom.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,110 @@
...
+        return array_map(function($node) {
+            return $node->id;
+        }, $nodes);
+    }
+
+    private function makeMetadata($node, $nodePostId) {
```
wynnset: Please move all the helper functions to the bottom.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,110 @@
...
+        foreach ($nodes as $node) {
+            if (!isset($node->id)) {
+                $nodePostId = $this->insertPost($node, 'tapestry_node');
+                $metadata = $this->makeMetadata($node, $nodePostId);
+                $node->id = add_post_meta($postId, 'tapestry_node', $metadata);
+            } else {
+                $metadata = get_metadata_by_mid('post', $node->id)->meta_value;
+                $nodePostId = $metadata->post_id;
+            }
+            update_post_meta($nodePostId, 'tapestry_node_data', $node);
```
wynnset: Don't you need to add in the tapestry ID somewhere in the meta key so we know this is the node in this tapstry?

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,110 @@
...
+    /**
+     * Update Tapestry nodes first then
+     * Update the existing Tapestry post if the postId is provided
+     * Otherwise, a new post will be created
+     * 
+     * @param type @post The post data
+     * @param type @postId The postId of the Tapestry
+     */
+    public function updateTapestry($tapestry, $postId = null) {
+        if (!isset($postId))
```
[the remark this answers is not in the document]
anhduc130: Thanks I'll fix it.

wp-content/plugins/tapestry/controller/permissions.php
```diff
@@ -0,0 +1,19 @@
+<?php
+/**
+ * Tapestry Endpoint Permissions
+ *
+ */
+
```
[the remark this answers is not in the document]
anhduc130: That's a good idea.

wp-content/plugins/tapestry/endpoints.php
```diff
@@ -3,3 +3,34 @@
...
+
+add_action( 'rest_api_init', function () {
+    register_rest_route( 'tapestry-tool/v1', '/tapestry-nodes', array(
+        'methods' => 'POST',
+        'callback' => 'updateTapestryNodes',
+        'permission_callback' => 'postTapestryNodePermissions'
+    ));
+});
+
+function updateTapestry($request) {
```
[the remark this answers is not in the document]
anhduc130: I see. I'll fix that.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,110 @@
...
+            $postId = $this->insertPost($tapestry, 'tapestry');
```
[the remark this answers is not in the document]
anhduc130: Sounds good. I'll put a comment to use `isJson()` once Kevin's PR is merged. I also agree that we should validate some tapestry elements at the upper level when receiving the data (in a separate task for sure).

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
anhduc130: I see. I updated.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
anhduc130: I updated.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
anhduc130: I updated.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
anhduc130: I don't think we need to add the ID to the meta key anymore. It's because we now have the `post_meta` of each node saved under the **tapestry**, which includes the node's `post_id` pointing to the node details.

In this way, we can enable sharing the node details among different tapestries. Only the node's `post_meta` saved under the **tapestry** is unique to a tapestry. And we use the `post_meta` saved under the **tapestry** to overwrite the node details such as title, coordinates etc. Does it make sense and align with what we discussed?

wynnset: Ok got it. I didn't realize this post meta refers to the node details that are shared.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,106 @@
...
+        if (!isset($tapestry->rootId)) {
+            $tapestry->rootId = $tapestry->nodes[0]->id;
+        }
+        $tapestry->nodes = $this->getNodeIds($tapestry->nodes);
+
+        update_post_meta($postId, 'tapestry', $tapestry);
+        return $tapestry;
+    }
+
+    private function updateNodes($nodes, $postId) {
```
wynnset: As a general rule, please prefix all private function names with an underscore. So for example this should be `_updateNodes`

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,106 @@
...
+                $node->id = add_post_meta($postId, 'tapestry_node', $metadata);
+            } else {
+                $metadata = get_metadata_by_mid('post', $node->id)->meta_value;
+                $nodePostId = $metadata->post_id;
+            }
+            update_post_meta($nodePostId, 'tapestry_node_data', $node);
+        }
+    }
+
+    private function updateWordPressPost($post, $type, $postId = null) {
```
wynnset: Please avoid using phrases that is the name of the application or plugin because it just makes the name longer. For example, this function should not have `WordPress` as part of it.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,106 @@
...
+        'TAPESTRY' => 'tapestry',
+        'TAPESTRY_NODE' => 'tapestry_node'
+    ];
+    const ERRORS = [
+        'INVALID_POST_ID' => [
+            'MESSAGE' => 'PostID is invalid',
+            'STATUS' => ['status' => 404]
+        ]
+    ];
+
```
wynnset: I recommend adding the postId as a class property that is optionally set in the constructor. That way, you don't have to pass it in as a parameter in your functions.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,106 @@
...
+     * @param  Object @tapestry The Tapestry data
+     * @param  Number @postId The postId of the Tapestry
+     * @return Object @tapestry
+     */
+    public function updateTapestry($tapestry, $postId = null) {
+        // TODO: use isValidPostID() utlility function
+        if (!isset($postId)) {
+            $postId = $this->updateWordPressPost($tapestry, 'tapestry');
+        }
+        $this->updateNodes($tapestry->nodes, $postId);
```
wynnset: Please add an empty line between each piece of functionality in your code. Essentially, you'd be grouping your code so it's more clear, even if you are not commenting what's happening in each part of your function. Right now, you are adding empty lines between some of your code, but it's a bit arbitrary and doesn't serve any purpose or make the code easier to read.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,106 @@
...
+            default:
+                $postType = $post->settings->type;
+                $postTitle = $post->settings->title;
+                $postStatus = $post->settings->status;
+                break;
+        }
+        if (isset($postId)) {
+            return wp_update_post(array(
+                'ID' => $postId,
+                'post_type' => $postType,
```
wynnset: Shouldn't this just be `$type`?

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,106 @@
...
+                break;
+            case self::POST_TYPES['TAPESTRY']:
+            default:
+                $postType = $post->settings->type;
+                $postTitle = $post->settings->title;
+                $postStatus = $post->settings->status;
+                break;
+        }
+        if (isset($postId)) {
+            return wp_update_post(array(
```
wynnset: You can use [`wp_insert_post`](https://developer.wordpress.org/reference/functions/wp_insert_post/) to do both update and insert, depending on whether an ID is passed it. I think that would be cleaner.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,106 @@
...
+        $tapestry->nodes = $this->getNodeIds($tapestry->nodes);
+
+        update_post_meta($postId, 'tapestry', $tapestry);
+        return $tapestry;
+    }
+
+    private function updateNodes($nodes, $postId) {
+        foreach ($nodes as $node) {
+            if (!isset($node->id)) {
+                $nodePostId = $this->updateWordPressPost($node, 'tapestry_node');
```
wynnset: How are you handling the case where an existing node is used in a different tapestry?

wp-content/plugins/tapestry/controller/class.tapestry-permissions.php
```diff
@@ -0,0 +1,14 @@
+<?php
+/**
+ * Tapestry Endpoint Permissions
+ *
+ */
+class TapestryPermissions {
+    static function postTapestryPermissions($request) {
```
wynnset: Since you are already in the `TapestryPermissions` class, I don't think it's needed to use the word `Permissions` in the name of your method here.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
anhduc130: We can include the `meta_id` of the post meta that points to this postId.

anhduc130: If we want to share the same node details but with a different title, x & y coordinates, permissions. We can just create a different post meta that points to this postId. The backend will grab the data in the post meta and overwrite the node details.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
anhduc130:
```
switch($type) {
    case self::POST_TYPES['TAPESTRY_NODE']:
        $postType = $post->type;
        $postTitle = $post->title;
        $postStatus = $post->status;
        break;
    case self::POST_TYPES['TAPESTRY']:
    default:
        $postType = $post->settings->type;
        $postTitle = $post->settings->title;
        $postStatus = $post->settings->status;
        break;
}
```
I think I named it $postType.

anhduc130: I consider `$type` param as optional. If the param `$type` is not provided, the switch case will fall into the default case, which generates the "tapestry" post. Does it help clarify my point?

anhduc130: Actually, I refactored the function a bit. Hope it's clear now.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,106 @@
...
+            'MESSAGE' => 'PostID is invalid',
+            'STATUS' => ['status' => 404]
+        ]
+    ];
+    private $postId;
+
+    /**
+     * Constructor
+     */
+    public function __construct($postId) {
```
wynnset: This parameter should default to 0 since that will allow it to be initialized without a parameter in order to create a new tapestry. See "ID" parameter description for [wp_insert_post](https://developer.wordpress.org/reference/functions/wp_insert_post/#parameters).

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,106 @@
...
+        $this->postId = $postId;
```
wynnset: You should validate `$postId` here to make sure it is a number, that it exists, and that it is a tapestry post type.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,106 @@
...
+     * @param  Object @tapestry The Tapestry data
+     * @return Object @tapestry
+     */
+    public function updateTapestry($tapestry) {
+        // TODO: use isValidPostID() utlility function
+        if (!isset($this->postId)) {
+            $this->postId = $this->_updatePost($tapestry, 'tapestry');
+        }
+
+        $this->_updateNodes($tapestry->nodes, $this->postId);
```
wynnset: You should never pass class properties as parameters in the class itself since they are accessible from your other methods.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,106 @@
...
+
+        if (!isset($tapestry->rootId)) {
+            $tapestry->rootId = $tapestry->nodes[0]->id;
```
wynnset: If there are no nodes passed in, this may throw a fatal error.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,106 @@
...
+            } else {
+                $metadata = get_metadata_by_mid('post', $node->id)->meta_value;
+                $nodePostId = $metadata->post_id;
+            }
+            update_post_meta($nodePostId, 'tapestry_node_data', $node);
+        }
+    }
+
+    private function _updatePost($post, $postType = 'tapestry', $postId = null) {
+        switch($postType) {
```
wynnset: I would suggest that if post type is `tapestry` and the class property `postId` is set already, then you should set it here so it saves it rather than create a new tapestry.

anhduc130:
I fixed the PR as requested. Please let me know if this is what we want.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
anhduc130: Interestingly, when I tested with no nodes passed it. It doesn't throw any internal server error. It just makes the `$tapestry->rootId` null. But you make a good point, we should check if the nodes array empty or not.

PR #21, opened 2019-05-10
10 Save node progress and h5p video settings

### LINKED ISSUE #10: Save node progress for each user in database instead of cookies
Save every 5 seconds

Should include amount viewed and also node locked / unlocked status

kevinthewebguy:
fixes #10


kevinthewebguy:
Ready for testing

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        if (!isset($postId)) {
+            throw new Exception('postId is invalid');
+        }
+    }
+
+    private function isJson($string) {
+        $test_json = json_decode($string);
+        if ($test_json) {
```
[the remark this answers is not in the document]
wynnset: See https://www.php.net/manual/en/function.json-decode.php#refsect1-function.json-decode-returnvalues

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,105 @@
+<?php
+/**
+ * Add/update/retrieve User progress
+ * 
+ */
+class UserController {
```
wynnset: You'll need to use a prefix when naming any classes or global variables and functions. Anything inside your class is fine since it's scope is limited to the class, but anything outside the class, including the class name itself should have a prefix. You can use `Tapestry` or `tapestry` as the prefix for everything. So for example here your controller should be called `TapestryUserController`. See https://developer.wordpress.org/plugins/plugin-basics/best-practices/#avoid-naming-collisions

wp-content/plugins/tapestry/endpoints.php
```diff
@@ -3,3 +3,73 @@
...
+add_action( 'rest_api_init', function () {
+    register_rest_route( 'tapestry-tool/v1', 'users/progress/', array(
+      'methods' => 'GET',
+      'callback' => 'get_user_progress_by_post_id',
+    ) );
+});
+
+// Get user progress on a tapestry page by user id and post id. Will need to pass these as query parameters
+// Example: http://localhost:8888/tapestry-wp/wp-json/tapestry-tool/v1/users/progress?userid=1&postid=42
+function get_user_progress_by_post_id($data) {
```
wynnset: As per my other comment for your controller class, please prefix this and any other global function or variable names with `tapestry_`.

wp-content/plugins/tapestry/endpoints.php
```diff
@@ -3,3 +3,73 @@
...
+    $userId = $data['userid'];
```
wynnset: Please always use underscores to separate the words so it's more readable. This should be `$data['user_id']` and the one below should be `$data['post_id']`. Please do this for your whole PR and keep this convention for the frontend work as well.

kevinthewebguy: changed to user_id on frontend and backend

wp-content/plugins/tapestry/endpoints.php
kevinthewebguy: Added tapestry_ prefix to related function names

wp-content/plugins/tapestry/controller/class.user-controller.php
kevinthewebguy: Changed to TapestryUserController

wp-content/plugins/tapestry/controller/class.user-controller.php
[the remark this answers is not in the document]
kevinthewebguy: Changed the if check to if ($test_json !== NULL)

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        // Not the current user. Comment out when testing the endpoints on Postman
+        if ($userId != apply_filters('determine_current_user', false)) {
+            throw new Exception('userId is invalid');
+        }
+
+        if (!isset($postId)) {
+            throw new Exception('postId is invalid');
+        }
```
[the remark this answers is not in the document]
kevinthewebguy: 1. Added a post type check.
2.  Currently this function is used for both UPDATE and GET. For the UPDATE situation, if postId does not exist in db, it is valid because it'll create the post ID. So I added a new function (postsExist) to check if postId exists in the db. I'll then use this function for the GET situations.

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,118 @@
...
+
+        if (!get_post_type($postId) == "tapestry") {
```
wynnset: Couple of things:
1. In your previous comment you said the postId might not yet exist in the db. If the postId doesn't exist, you wouldn't be able to check its type. Isn't that right? 
2. nit: please use `(get_post_type($postId) != "tapestry")` instead to be more clear.

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,118 @@
...
+ */
+class TapestryUserController {
+
+    /** 
+     * Update User's video progress for a tapestry post
+     * @param type @userId the user's id
+     * @param type @postId the post's ID
+     * @param type @progressData json data for progress of each node
+    */
+    public function updateProgress($userId, $postId, $progressData) {
```
wynnset: Rather than updating the whole progress data, I think we should store the progress _per node_. We'll be doing this save very often, therefore it would be much faster if the data load is smaller.

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,118 @@
...
+     * @param type @postId the post's ID
+    */
+    public function getH5PSetting($userId, $postId) {
+        $this->postExists($postId);
+        $this->checkUserAndPostId($userId, $postId);  
+        return $this->getUserH5PSetting($userId, $postId);
+    }
+
+    private function updateUserProgress($userId, $postId, $progressData) {
+        update_user_meta($userId, 'progress_data_' . $postId, $progressData);
```
wynnset: Please prefix this meta key and all other meta keys with `tapestry_` to avoid conflicts. You may want to use a format like `tapestry_{tapestryId}_progress_node_{nodeMetaId}` or like `tapestry_progress_node_{nodeMetaId}` to store the progress for a given node

wp-content/plugins/tapestry/endpoints.php
```diff
@@ -3,3 +3,73 @@
...
+// Get user's progress
+add_action( 'rest_api_init', function () {
+    register_rest_route( 'tapestry-tool/v1', 'users/progress/', array(
+      'methods' => 'GET',
+      'callback' => 'tapestry_get_user_progress_by_post_id',
+    ) );
+});
+
+// Get user progress on a tapestry page by user id and post id. Will need to pass these as query parameters
+// Example: http://localhost:8888/tapestry-wp/wp-json/tapestry-tool/v1/users/progress?userid=1&postid=42
```
wynnset: Don't forget to update the parameter names in your examples too.

kevinthewebguy: Updated parameter names of the examples.

wp-content/plugins/tapestry/controller/class.user-controller.php
kevinthewebguy: 1. Moved post type check together with checking if the post ID exists in a function called isPostValid. This will only be used for GET requests.
2. Changed if statement to (get_post_type($postId) != "tapestry")

wp-content/plugins/tapestry/endpoints.php
```diff
@@ -3,3 +3,73 @@
...
+    register_rest_route( 'tapestry-tool/v1', '/users/progress/', array(
+      'methods' => 'POST',
+      'callback' => 'tapestry_update_user_progress_by_post_id',
+    ) );
+});
+
+// Example: http://localhost:8888/tapestry-wp/wp-json/tapestry-tool/v1/users/progress?user_id=1&post_id=44&json={"32":0.5,"33":0,"34":0,"35":0,"36":0,"37":0,"38":0,"39":0,"40":0,"41":0}
+function tapestry_update_user_progress_by_post_id($data) {
+    $userController = new TapestryUserController;
+    $userId = $data['user_id'];
```
wynnset: I don't think we need to pass in the user_id here since we have access to it through `get_current_user()` function. I know you are validating that the user passed in is the same user as the current user, but I don't see why we need to pass it in in the first place.

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,120 @@
+<?php
+/**
+ * Add/update/retrieve User progress
+ * 
+ */
+class TapestryUserController {
+
```
wynnset: For errors, could you please follow what Andrew is doing in PR #20 as I find it a bit more elegant? So add in an array for error types like ```    const ERRORS = [
        'INVALID_POST_ID' => [
            'MESSAGE' => 'PostID is invalid',
            'STATUS' => ['status' => 404]
        ]
    ];```
and add in your own errors in here. Then also copy in the `throwsError` function in here and use that instead of the generic `throw new Exception`.

kevinthewebguy: That sounds good. Since him and I may have the same errors, should we create a new issue to merge this into a class to avoid duplication?

wp-content/plugins/tapestry/controller/class.user-controller.php
kevinthewebguy: Saved key as tapestry_{tapestryId}_progress_node_{nodeMetaId}

wp-content/plugins/tapestry/controller/class.user-controller.php
kevinthewebguy: Saving node progress one node at a time

wp-content/plugins/tapestry/endpoints.php
kevinthewebguy: get_current_user() seemed to always return 0 for me. Using apply_filters('determine_current_user', false) instead.

wp-content/plugins/tapestry/templates/single-tapestry.php
```diff
@@ -34,7 +34,9 @@
 			<script src="<?php echo plugin_dir_url(__FILE__) ?>tapestry-d3/libs/h5p-resizer.min.js" charset="UTF-8"></script>
 
             <script>
-            	var jsonUrl = "<?php echo plugin_dir_url(__FILE__) ?>tapestry-d3/tapestry.json";
+				var jsonUrl = "<?php echo plugin_dir_url(__FILE__) ?>tapestry-d3/tapestry.json";
```
wynnset: We shouldn't need this anymore, right? Since we are loading the json from the API we are building.

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,141 @@
...
+     * @param type @postId the post's Id
+    */
+    public function getH5PSetting($postId) {
+        $userId = apply_filters('determine_current_user', false);
+        $this->isPostValid($postId);
+        $this->checkUserAndPostId($userId, $postId);  
+        return $this->getUserH5PSetting($userId, $postId);
+    }
+
+    private function updateUserProgress($userId, $postId, $nodeId, $progressValue) {
```
wynnset: As a convention, please name all private functions starting with an underscore.

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,141 @@
...
+
+        $this->updateUserProgress($userId, $postId, $nodeId, $progressValue);
+    }
+
+    /** 
+     * Get User's video progress for a tapestry post
+     * @param type @postId the post's ID
+     * @param type @nodeIdArr is a list of ids currently in the tapestry
+    */
+    public function getProgress($postId, $nodeIdArr) {
```
wynnset: I think it's better if you get the array of nodes in the backend, rather than as a parameter here. Unless you have a reason why you need to have that passed in?

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,141 @@
...
+        return $this->getUserProgress($userId, $postId, $nodeIdArr);
+    }
+
+    /** 
+     * Update User's h5p video setting for a tapestry post
+     * @param type @postId the post's ID
+     * @param type @h5pSettingsData stores volume, playbackRate, quality of h5p video
+    */
+    public function updateH5PSetting($postId, $h5pSettingsData) {
+        $userId = apply_filters('determine_current_user', false);
```
wynnset: Please use the function [get_current_user_id()](https://developer.wordpress.org/reference/functions/get_current_user_id/) instead

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,141 @@
+<?php
+// TODO Change exceptions to using an ERROR class
+/**
+ * Add/update/retrieve User progress
+ * 
+ */
+class TapestryUserController {
+
```
wynnset: I think it would be better if you save the current user id to a class property when the class is constructed so you can use it anywhere in your class. You may want to declare functions as static if they don't need the class property.

wp-content/plugins/tapestry/controller/class.user-controller.php
kevinthewebguy: This would require me to get Andrew's new tapestrycontroller code. I would be blocked again.

wp-content/plugins/tapestry/templates/single-tapestry.php
kevinthewebguy: This will be removed in another PR that loads tapestry from the database.

wp-content/plugins/tapestry/endpoints.php
```diff
@@ -3,3 +3,71 @@
...
+
+// Get user's progress
+add_action( 'rest_api_init', function () {
+    register_rest_route( 'tapestry-tool/v1', 'users/progress/', array(
+      'methods' => 'GET',
+      'callback' => 'tapestry_get_user_progress_by_post_id',
+    ) );
+});
+
+// Get user progress on a tapestry page by user id and post id. Will need to pass these as query parameters
```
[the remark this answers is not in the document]
kevinthewebguy: Updated description.

wp-content/plugins/tapestry/controller/class.user-controller.php
kevinthewebguy: Updated to include underscore before function name for private functions.

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,141 @@
...
+            throw new Exception('userId is invalid');
+        }
+
+        if (!isset($postId)) {
+            throw new Exception('postId is invalid');
+        }
+
+    }
+
+    private function isPostValid ($postId) {
```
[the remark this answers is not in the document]
kevinthewebguy: changed function name to isvalidTapestryPost

wp-content/plugins/tapestry/controller/class.user-controller.php
kevinthewebguy: Added User Id as class property.

wp-content/plugins/tapestry/controller/class.user-controller.php
kevinthewebguy: get_current_user_id() is weird in that even if someone is logged in it returns 0. get_current_user_id might not be the best in this case as I think we need to do some preprocessing with init action. (I'm not too sure about the details of this). 

https://wordpress.stackexchange.com/questions/163407/get-current-user-id-returns-0

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,143 @@
...
+        if ($progressValue !== NULL) {
+            $progressValue = json_decode($progressValue);
+        }
+
+        // Value should be between 0 and 1
+        if ($progressValue < 0 || $progressValue > 1) {
+            throw new Exception('Invalid progress value');
+        }
+
+        $this->_updateUserProgress($this->userId, $postId, $nodeId, $progressValue);
```
wynnset: You shouldn't be passing class properties to your methods since they are already accessible from your methods. Please fix this everywhere.

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,143 @@
...
+
+        return $this->_getUserProgress($this->userId, $postId, $nodeIdArr);
+    }
+
+    /** 
+     * Update User's h5p video setting for a tapestry post
+     * @param type @postId the post's ID
+     * @param type @h5pSettingsData stores volume, playbackRate, quality of h5p video
+    */
+    public function updateH5PSetting($postId, $h5pSettingsData) {
```
wynnset: Please rename this to `updateH5PSettings` since it's multiple settings, not just one setting.

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,143 @@
...
+        $this->_isValidTapestryPost($postId);
+        $this->_checkUserAndPostId($this->userId, $postId);
+        return $this->_getUserH5PSetting($this->userId, $postId);
+    }
+
+    private function _updateUserProgress($userId, $postId, $nodeId, $progressValue) {
+        update_user_meta($userId, 'tapestry_' . $postId . '_progress_node_' . $nodeId, $progressValue);
+    }
+
+    private function _getUserProgress($userId, $postId, $nodeIdArr) {
```
wynnset: As mentioned above, you should find the node IDs from the backend, not receive them as parameters.

wp-content/plugins/tapestry/controller/class.user-controller.php
kevinthewebguy: renamed to updateH5PSettings

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,143 @@
...
+     * Update User's video progress for a tapestry post
+     * @param type @postId the post's ID
+     * @param type @nodeId the current node being viewed
+     * @param type @progressValue is how much the video was viewed, value should be between >= 0 and <= 1
+    */
+    public function updateProgress($postId, $nodeId, $progressValue) {
+        $this->_checkUserAndPostId($this->userId, $postId);
+
+        if ($progressValue !== NULL) {
+            $progressValue = json_decode($progressValue);
```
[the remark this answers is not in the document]
kevinthewebguy: It is a string and needs to be converted to float.using floatval() function

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    /** 
+     * Update User's video progress for a tapestry post
+     * @param type @postId the post's ID
+     * @param type @nodeId the current node being viewed
+     * @param type @progressValue is how much the video was viewed, value should be between >= 0 and <= 1
+    */
+    public function updateProgress($postId, $nodeId, $progressValue) {
+        $this->_checkUserAndPostId($this->userId, $postId);
```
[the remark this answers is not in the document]
kevinthewebguy: Removing passing in userId as parameter

wp-content/plugins/tapestry/controller/class.user-controller.php
kevinthewebguy: Removing passing userId

wp-content/plugins/tapestry/controller/class.user-controller.php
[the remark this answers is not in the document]
kevinthewebguy: We need to check for user because if we don't, anyone with the endpoint can view the data if they are using an application like Postman

wp-content/plugins/tapestry/controller/class.user-controller.php
kevinthewebguy: Yes, we use get_post_meta(). But since we split our pull requests, his tapestry class is not in my PR, and I do not have the functions to add nodes to the db to test that functionality. My current code is using the ids from tapestry.json to update the progress value in the user_meta. 

I could take his latest code in his PR, since his PR is not merged theres a possibility that the code is changed.

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,143 @@
...
+/**
+ * Add/update/retrieve User progress
+ * 
+ */
+class TapestryUserController {
+
+    private $userId = NULL;
+
+    public function __construct() {
+        $this->userId = apply_filters('determine_current_user', false);
```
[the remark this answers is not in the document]
kevinthewebguy: get_current_user_id() returns 0 regardless if someone has logged in or not. Tried debugging this and not sure how to fix it.

wp-content/plugins/tapestry/controller/class.user-controller.php
kevinthewebguy: Changed to receive node ids from backend.

wp-content/plugins/tapestry/controller/class.user-controller.php
kevinthewebguy: Aidin told me to leave it as is and not use get_current_user_id.

wp-content/plugins/tapestry/controller/class.user-controller.php
wynnset: Please use the `getTapestryNodeIds` function you added in the tapestry controller.

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,140 @@
...
+     * @param type @postId the post's Id
+    */
+    public function getH5PSettings($postId) {
+        $this->_isValidTapestryPost($postId);
+        $this->_checkUserAndPostId($postId);
+        return $this->_getUserH5PSettings($postId);
+    }
+
+    private function _updateUserProgress($postId, $nodeId, $progressValue) {
+        $userId = $this->userId;
```
wynnset: I would just use `$this->userId` directly in the function below

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,140 @@
...
+        update_user_meta($userId, 'tapestry_' . $postId . '_progress_node_' . $nodeId, $progressValue);
+    }
+
+    private function _getUserProgress($postId, $nodeIdArr) {
```
wynnset: Again, please use the `getTapestryNodeIds` function.

wp-content/plugins/tapestry/controller/class.user-controller.php
```diff
@@ -0,0 +1,140 @@
...
+            } else {
+                $progress->$nodeId = 0.0;
+            }
+        }
+
+        return json_encode($progress);
+    }
+
+    private function _updateUserH5PSettings($postId, $h5pSettingsData) {
+        $userId = $this->userId;
```
wynnset: I prefer you not declare new variables if you are using them only once or twice. It'd be easier to read your code if you just plug `$this->userId` directly where you need it. Same goes in other places in your code. Thanks!

wp-content/plugins/tapestry/endpoints.php
```diff
@@ -6,6 +6,76 @@
...
+// Get user's progress
+add_action( 'rest_api_init', function () {
+    register_rest_route( 'tapestry-tool/v1', 'users/progress/', array(
+      'methods' => 'GET',
+      'callback' => 'tapestry_get_user_progress_by_post_id',
+    ) );
+});
+
+// Get user progress on a tapestry page by post id. 
+// Example: http://localhost:8888/tapestry-wp/wp-json/tapestry-tool/v1/users/progress?post_id=44
```
wynnset: Could you move this comment block up to before `add_action` and remove any empty lines between this function declaration and the end of the previous code block? That way, it's all nicely grouped together and easier to read. Do the same for the rest of your endpoints too. Thanks!

wp-content/plugins/tapestry/controller/class.user-controller.php
kevinthewebguy: I prefer it this way for extendability in the future. But will change.

wp-content/plugins/tapestry/controller/class.user-controller.php
kevinthewebguy: What do you mean here? Not applicable to this code section. It is just taking the given node id and updating the usermeta.

wp-content/plugins/tapestry/controller/class.user-controller.php
kevinthewebguy: Can't seem to reply to the getTaptestryNodeIds comment. One where it says "Please use the getTapestryNodeIds function you added in the tapestry controller.". It is using it in the endpoint.

wp-content/plugins/tapestry/controller/class.user-controller.php
wynnset: Why are you passing in `$nodeIdArr` instead of using the `getTapestryNodeIds` function? What do you mean updating the usermeta? This is getting usermeta, not updating it from what I see.

wp-content/plugins/tapestry/endpoints.php
```diff
@@ -6,6 +6,71 @@
 
 require __DIR__ . '/controller/class.tapestry-permissions.php';
 require __DIR__ . '/controller/class.tapestry-controller.php';
+require __DIR__ . '/controller/class.user-controller.php';
+
+ // User endpoints
+
+// Get user progress on a tapestry page by post id. 
+// Example: http://localhost:8888/tapestry-wp/wp-json/tapestry-tool/v1/users/progress?post_id=44
```
wynnset: This is looking much better, thanks! Please just remove `http://localhost:8888/tapestry-wp/` from your comment as the address might be different somewhere else. Don't ever hardcode the website address in the plugin code, even in comments.

wp-content/plugins/tapestry/controller/class.user-controller.php
wynnset: I mean why are you passing in `$nodeIdArr` instead of just retrieving the node ids using the `getTapestryNodeIds` function?

wp-content/plugins/tapestry/endpoints.php
kevinthewebguy: Reformatted the code in the endpoints file.

wp-content/plugins/tapestry/endpoints.php
kevinthewebguy: Removed the http://localhost:8888/tapestry-wp from comments

controller/class.user-controller.php
```diff
@@ -0,0 +1,136 @@
...
+        update_user_meta($this->userId, 'tapestry_h5p_setting_' . $postId, $h5pSettingsData);
+    }
+
+    private function _getUserH5PSettings($postId) {
+        $settings = get_user_meta($this->userId, 'tapestry_h5p_setting_' . $postId, true);
+        return json_encode($settings);
+    }
+
+    // Helpers
+    private function _checkUserAndPostId($postId) {
```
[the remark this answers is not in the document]
wynnset: Leave the current comment and start another comment below to use as the doc comment block for the function.

endpoints.php
```diff
@@ -6,6 +6,71 @@
...
+
+// Get user h5p video setting on a tapestry page by post id. Will need to pass these as query parameters
+// Example: /wp-json/tapestry-tool/v1/users/h5psettings?post_id=42
+add_action( 'rest_api_init', function () {
+    register_rest_route( 'tapestry-tool/v1', 'users/h5psettings/', array(
+      'methods' => 'GET',
+      'callback' => 'tapestry_get_user_h5p_settings_by_post_id',
+    ) );
+});
+function tapestry_get_user_h5p_settings_by_post_id($data) {
```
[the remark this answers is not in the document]
kevinthewebguy: following convention of our code base to use lowercase.

endpoints.php
```diff
@@ -6,6 +6,71 @@
...
+
+    $userController = new TapestryUserController;
+    return $userController->getProgress($postId, $nodeIdArr);
+}
+
+// Update a single node progress by passing in node id, post id and progress value
+// Example: /wp-json/tapestry-tool/v1/users/progress?post_id=44&node_id=1&progress_value=0.2
+add_action( 'rest_api_init', function () {
+    register_rest_route( 'tapestry-tool/v1', '/users/progress/', array(
+      'methods' => 'POST',
```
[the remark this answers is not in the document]
kevinthewebguy: done

endpoints.php
```diff
@@ -6,6 +6,71 @@
...
+
+// Get user progress on a tapestry page by post id. 
+// Example: /wp-json/tapestry-tool/v1/users/progress?post_id=44
+add_action( 'rest_api_init', function () {
+    register_rest_route( 'tapestry-tool/v1', 'users/progress/', array(
+      'methods' => 'GET',
+      'callback' => 'tapestry_get_user_progress_by_post_id',
+    ) );
+});
+function tapestry_get_user_progress_by_post_id($data) {
```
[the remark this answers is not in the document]
kevinthewebguy: already in lower case (convention for our code)

controller/class.user-controller.php
```diff
@@ -0,0 +1,136 @@
...
+        if ($this->_isJson($h5pSettingsData)) {
+            $h5pSettingsData = json_decode($h5pSettingsData);
+        } else {
+            throw new Exception('Invalid json');
+        }
+
+        $this->_updateUserH5PSettings($postId, $h5pSettingsData);
+    }
+
+    /**
```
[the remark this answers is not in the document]
kevinthewebguy: Already there.

controller/class.user-controller.php
```diff
@@ -0,0 +1,136 @@
...
+        } else {
+            throw new Exception('Invalid json');
+        }
+
+        $this->_updateUserH5PSettings($postId, $h5pSettingsData);
+    }
+
+    /** 
+     * Get User's h5p video setting for a tapestry post
+     * @param type @postId the post's Id
```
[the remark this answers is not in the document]
kevinthewebguy: Already there.

controller/class.user-controller.php
```diff
@@ -0,0 +1,136 @@
...
+    */
```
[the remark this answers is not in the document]
kevinthewebguy: Done

controller/class.user-controller.php
```diff
@@ -0,0 +1,136 @@
...
+    public function getH5PSettings($postId) {
+        $this->_isValidTapestryPost($postId);
+        $this->_checkUserAndPostId($postId);
+        return $this->_getUserH5PSettings($postId);
+    }
+
+    private function _updateUserProgress($postId, $nodeId, $progressValue) {
```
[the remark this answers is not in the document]
kevinthewebguy: done

controller/class.user-controller.php
```diff
@@ -0,0 +1,136 @@
...
+        update_user_meta($this->userId, 'tapestry_' . $postId . '_progress_node_' . $nodeId, $progressValue);
+    }
+
+    private function _getUserProgress($postId, $nodeIdArr) {
```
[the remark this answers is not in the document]
kevinthewebguy: done

controller/class.user-controller.php
```diff
@@ -0,0 +1,136 @@
...
+        $progress = new stdClass();
+
+        // Build json object for frontend e.g. {0: 0.1, 1: 0.2} where 0 and 1 are the node IDs
+        foreach ($nodeIdArr as $nodeId) {
+            $progress_value = get_user_meta($this->userId, 'tapestry_' . $postId . '_progress_node_' . $nodeId, true);
+            if ($progress_value !== NULL) {
```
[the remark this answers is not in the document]
kevinthewebguy: made all NULLs lowercase.

controller/class.user-controller.php
```diff
@@ -0,0 +1,136 @@
...
+                $progress->$nodeId = $progress_value;
+            } else {
+                $progress->$nodeId = 0.0;
+            }
+        }
+
+        return json_encode($progress);
+    }
+
+    private function _updateUserH5PSettings($postId, $h5pSettingsData) {
```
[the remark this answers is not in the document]
kevinthewebguy: done

controller/class.user-controller.php
```diff
@@ -0,0 +1,136 @@
...
+        update_user_meta($this->userId, 'tapestry_h5p_setting_' . $postId, $h5pSettingsData);
+    }
+
+    private function _getUserH5PSettings($postId) {
```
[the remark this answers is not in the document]
kevinthewebguy: done

controller/class.user-controller.php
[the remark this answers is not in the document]
kevinthewebguy: done

controller/class.user-controller.php
```diff
@@ -0,0 +1,136 @@
...
+            throw new Exception('post id does not exist');
+        }
+
+        // Post type is correct
+        if (get_post_type($postId) != "tapestry") {
+            throw new Exception('post type is invalid');
+        }
+    }
+
+    private function _isJson($string) {
```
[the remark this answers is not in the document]
kevinthewebguy: done

controller/class.user-controller.php
```diff
@@ -0,0 +1,136 @@
...
+        $test_json = json_decode($string);
+        if ($test_json !== NULL) {
```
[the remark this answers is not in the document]
kevinthewebguy: done

controller/class.user-controller.php
```diff
@@ -0,0 +1,136 @@
...
+            throw new Exception('postId is invalid');
+        }
+
+        if (!isset($postId)) {
+            throw new Exception('postId is invalid');
+        }
+
+    }
+
+    private function _isValidTapestryPost ($postId) {
```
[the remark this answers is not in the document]
kevinthewebguy: fixed

endpoints.php
```diff
@@ -6,6 +6,71 @@
...
+    $userController = new TapestryUserController;
+    return $userController->getProgress($postId, $nodeIdArr);
+}
+
+// Update a single node progress by passing in node id, post id and progress value
+// Example: /wp-json/tapestry-tool/v1/users/progress?post_id=44&node_id=1&progress_value=0.2
+add_action( 'rest_api_init', function () {
+    register_rest_route( 'tapestry-tool/v1', '/users/progress/', array(
+      'methods' => 'POST',
+      'callback' => 'tapestry_update_user_progress_by_node_id',
```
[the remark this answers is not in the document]
kevinthewebguy: done

endpoints.php
```diff
@@ -30,7 +100,8 @@ function addTapestryNode($request) {
         'methods' => 'POST',
         'callback' => 'updateTapestry',
         'permission_callback' => 'TapestryPermissions::postTapestry'
-    ));
+    )
```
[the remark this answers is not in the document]
wynnset: This doesn't look right. Please either indent correctly or put it back the way it was.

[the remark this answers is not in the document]
kevinthewebguy: fixed this to use previous format.

kevinthewebguy:
> For your comment doc parameters, you are naming them with @ like @postid but it should be your actual variable name like $postId

Fixed to use $ instead of @

kevinthewebguy:
Fixed the spacing issues. The comments about the spaces for parameter names and type seems inconsistent, so I'll be skipping those.

controller/class.user-controller.php
```diff
@@ -0,0 +1,189 @@
...
+    private $_userId = null;
+
+    public function __construct() {
+        $this->_userId = apply_filters('determine_current_user', false); 
+    }
+
+    /**
+     * Update User's video progress for a tapestry post
+     *
+     * @param Integer $postId the post's ID
```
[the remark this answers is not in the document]
wynnset: The point of these comments is to align all the variable names and variable descriptions together like
```
@param Integer $postId               the post's ID
@param Integer $nodeId              the current node being viewed
@param Float     $progressValue percentage viewed (value should be between >= 0 and <= 1)
```

controller/class.user-controller.php
```diff
@@ -0,0 +1,177 @@
...
+    public function __construct() {
+        $this->userId = apply_filters('determine_current_user', false); 
+    }
+
+    /**
+      * Update User's video progress for a tapestry post
+      * @param Integer $postId the post's ID
+      * @param Integer $nodeId the current node being viewed
+      * @param Float $progressValue is how much the video was viewed, value should be between >= 0 and <= 1
+    */
```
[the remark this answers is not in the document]
wynnset: `@return` tag should be null

controller/class.user-controller.php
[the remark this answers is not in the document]
kevinthewebguy: Thanks, I have aligned my variable names and descriptions as described.

controller/class.user-controller.php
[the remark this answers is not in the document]
kevinthewebguy: Added return tag with null.

wynnset:
Please fix spacing issues and also rebase this branch, then assign to me for review.

kevinthewebguy:
Fixed merge conflict and spacing.

wynnset:
Still many hound items that make sense need to be fixed. Please review before assigning to me @kevinthewebguy

controller/class.user-controller.php
```diff
@@ -0,0 +1,189 @@
...
+        } else {
+            throw new Exception('Invalid json');
+        }
+
+        $this->_updateUserH5PSettings($postId, $h5pSettingsData);
+    }
+
+    /**
+     * Get User's h5p video setting for a tapestry post
+     * @param Integer $postId the post's Id
```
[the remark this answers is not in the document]
kevinthewebguy: There is a blank line

controller/class.tapestry-controller.php
```diff
@@ -188,7 +188,21 @@ function ($groupMetaId) {
         return $tapestry;
     }
 
-    private function _formNodeData($nodeData, $metadata)
+    /**
+     * Retrieve all node ids associated to a tapestry
+     *
+     */
+    public function getTapestryNodeIds()
```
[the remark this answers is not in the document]
kevinthewebguy: There is a return

controller/class.user-controller.php
```diff
@@ -0,0 +1,177 @@
...
+        $this->_updateUserProgress($postId, $nodeId, $progressValue);
+    }
+
+    /**
+      * Get User's video progress for a tapestry post
+      * @param Integer $postId the post's ID
+      * @param Array $nodeIdArr is a list of ids currently in the tapestry
+      *
+      * @return String progress of each node in json format
+    */
```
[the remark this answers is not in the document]
kevinthewebguy: added space now

controller/class.user-controller.php
```diff
@@ -0,0 +1,159 @@
...
+
+    private function _getUserH5PSettings($postId) 
+    {
+        $settings = get_user_meta($this->_userId, 'tapestry_h5p_setting_' . $postId, true);
+        return json_encode($settings);
+    }
+
+    // Helpers
+
+    private function _checkUserAndPostId($postId)
```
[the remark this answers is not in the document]
kevinthewebguy: Changed to /* now

controller/class.user-controller.php
```diff
@@ -0,0 +1,161 @@
...
+
+        return $this->_getUserProgress($postId, $nodeIdArr);
+    }
+
+    /**
+     * Update User's h5p video setting for a tapestry post
+     *
+     * @param Integer $postId          the post's ID
+     * @param String  $h5pSettingsData stores volume, playbackRate, quality of h5p video
+     */
```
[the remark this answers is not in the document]
kevinthewebguy: No need for @return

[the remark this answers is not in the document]
kevinthewebguy: Not needed for this function.

kevinthewebguy:
Fixed hound comments except for doc comments for private functions, number of characters exceeded in a line and @author, etc comment

endpoints.php
```diff
@@ -6,6 +6,71 @@
 
 require __DIR__ . '/controller/class.tapestry-permissions.php';
 require __DIR__ . '/controller/class.tapestry-controller.php';
+require __DIR__ . '/controller/class.user-controller.php';
+
+ // User endpoints
+
+// Get user progress on a tapestry page by post id. 
+// Example: /wp-json/tapestry-tool/v1/users/progress?post_id=44
+add_action( 'rest_api_init', function () {
```
anhduc130: @kevinthewebguy @wynnset  Shall we follow the pattern of creating an endpoint like other endpoints we have created here? Just for maintenance and consistency purposes.

kevinthewebguy:
I have conformed to the new style and fixed the hound requests.

PR #22, opened 2019-05-10
Load tapestries from a database instead of json files

### LINKED ISSUE #9: Load tapestries from a database instead of json files

kevinthewebguy:
fixes #9


wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,83 @@
...
+    }
+
+    private function _getTapestryById($postId) {
+        $tapestry = get_post_meta($postId, 'tapestry', true);
+
+        $metadatas = array_map(function($nodeId) {
+            return get_metadata_by_mid('post', $nodeId);
+        }, $tapestry->nodes);
+
+        $nodeDatas = array_map(function($metadata) {
```
wynnset: I would append the content of this function into the previous `array_map`, rather than doing `array_map` again here since you are not using `$metadatas` anywhere else. i.e. just have 1 `array_map` rather than 2.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,83 @@
...
+        }, $metadatas);
+
+        $tapestry->nodes = $nodeDatas;
+
+        // TODO: delete the below when being able to create tapestry from scratch
+        $tapestry->links = $this->_getNewLinks($tapestry->links, $nodeDatas);
+        return $tapestry;
+    }
+
+    private function _updateNodeData($nodeData, $metadata) {
```
wynnset: Please use a different word than `update` because "update" could imply something is getting updated in the db. Maybe `_formNodeData`

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -3,4 +3,83 @@
...
+        $nodeData->id = (int) $metadata->meta_id;
+        $nodeData->title = $metadata->meta_value->title;
+        $nodeData->fx = $metadata->meta_value->coordinates->x;
+        $nodeData->fy = $metadata->meta_value->coordinates->y;
+        return $nodeData;
+    }
+
+    // TODO: Remove this when we can build a tapestry from scratch
+    // HACK - create a new links array that works with new IDs
+    private function _getNewLinks($oldLinks, $nodes) {
```
wynnset: Instead of doing this, could you manually update the data in the database to contain the actual node IDs? This is too hacky to go into master.

anhduc130: The function is meant to support Kevin's work (I don't need this function at all). Yes, if he agrees to do that manually...

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
anhduc130: Ah right. Yeah, it's a good point!

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
anhduc130: Also, I think the data in the DB is serialized when inserted. It might be tricky to modify part of the data.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -55,6 +52,73 @@ public function updateTapestry($tapestry) {
...
+    /**
+     * Retrieve a Tapestry post
+     * 
+     * @return Object Tapestry
+     */
+    public function getTapestry() {
+        if (!$this->postId) {
+            return $this->_throwsError('INVALID_POST_ID');
+        }
+        return $this->_getTapestryById($postId);
```
wynnset: `$postId` is not defined here.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -55,6 +52,73 @@ public function updateTapestry($tapestry) {
...
+    }
+
+    private function _getTapestryById($postId) {
+        $tapestry = get_post_meta($postId, 'tapestry', true);
+
+        $tapestry->nodes = array_map(function($nodeMetaId) {
+            $metadata = get_metadata_by_mid('post', $nodeMetaId);
+            $postId = $metadata->meta_value->post_id;
```
wynnset: Suggest you rename this to `$nodePostId` so the reader doesn't confuse it with the function parameter.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
wynnset: Ok if it's too time-consuming, you don't have to do it. But if you wanted to do it, you can just give Kevin and I an export of your DB `posts` and `postmeta` tables. And changing it should not be difficult - you can just unserialize the data by googling "unseralize php" and using any of those tools and then seralizing it again before inserting it into the db. Again, if you think it's too time-consuming, then just leave it and we can remove it from the code later.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -55,6 +52,73 @@ public function updateTapestry($tapestry) {
...
+            $nodeData = get_post_meta($postId, 'tapestry_node_data', true);
+            return $this->_formNodeData($nodeData, $metadata);
+        }, $tapestry->nodes);
+
+        // TODO: delete the below when being able to create tapestry from scratch
+        $tapestry->links = $this->_getNewLinks($tapestry->links, $nodeDatas);
```
wynnset: `$nodeDatas` is undefined here.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
anhduc130: Thanks a lot for pointing it out. My bad on this, I didn't refactor carefully.

anhduc130:
I have fixed and tested the endpoint as requested.

PR #25, opened 2019-05-14
Add changing Tapestry settings 

### LINKED ISSUE #24: Add pop-up for changing Tapestry settings (admin only)

anhduc130:
Fixes #24


wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -55,6 +55,26 @@ public function updateTapestry($tapestry) {
...
+     * @param  Object @settings
+     * @return Object Tapestry settings
+     */
+    public function updateTapestrySettings($settings) {
+        if (!$this->postId) {
+            return $this->_throwsError('INVALID_POST_ID');
+        }
+
+        $tapestry = get_post_meta($this->postId, 'tapestry', true);
+        $tapestry->settings = $settings;
```
wynnset: Could you add a todo here to add some validation for the $settings that is being passed in?

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
```diff
@@ -55,6 +55,26 @@ public function updateTapestry($tapestry) {
...
+
+        $this->_updatePost($tapestry, 'tapestry');
```
wynnset: Is this doing anything? If not, could you please remove it? Thanks.

anhduc130: Yes, it is used to update the post title of the post.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
anhduc130: Should it be validated in the `endpoints.php`?

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
wynnset: Are you sure? The body of `_updatePost` function is setting the title of the post to `$post->title`, not `$post->settings->title` so either that function is wrong or this shouldn't be here.

wynnset: Oh sorry, I was looking at `TAPESTRY_NODE` .. I see that for `TAPESTRY` it is using settings.

wp-content/plugins/tapestry/controller/class.tapestry-controller.php
wynnset: I think doing the validation here makes more sense actually.

PR #27, opened 2019-05-15
Allow adding of new nodes 

### LINKED ISSUE #26: Allow adding of new nodes through the UI (admin only)
Each node will have a dangling plus sign (almost like a child node that is very close to the node). The admin user can click on that plus sign, which will cause a pop-up form to display where they can enter the node details.

Fields:

* Title
* Thumbnail URL
* Content Type
* Content Details
  * Video:
    * Video URLs (min. 1 required)
      * Video URL - 1080p (if available)
      * Video URL - 720p (if available)
      * Video URL - 480p (if available)
  * H5P
    * H5P Embed Code?

anhduc130:
Fixes #26


wp-content/plugins/tapestry/controller/class.tapestry-permissions.php
```diff
@@ -19,4 +19,12 @@ static function putTapestrySettings($request) {
         }
         return current_user_can('publish_posts');
     }
+
+    static function postTapestryNode($request) {
+        $postId = $request['id'];
```
wynnset: Doesn't this index also need to be changed to `tapestryPostId`?

anhduc130: Oh yes, I missed that. Will change it now.

PR #30, opened 2019-05-20
Adding Modal for adding new nodes

### LINKED ISSUE #29: Adding Modal for adding new nodes

kevinthewebguy:
fixes #29


templates/single-tapestry.php
```diff
@@ -29,29 +29,39 @@ function addNonceToScript()
 
 get_header(); ?>
 
+<?php require "utility/utility.php";?>
```
wynnset: I'd prefer if you didn't put one file in a directory, so just put utility.php into the templates directory maybe.

kevinthewebguy: moved to a single utility.php folder

templates/single-tapestry.php
```diff
@@ -29,9 +29,11 @@ function addNonceToScript()
 
 get_header(); ?>
 
+<?php require "utility.php";?>
 <div id="primary" class="content-area col-md-12">
     <main id="main" class="post-wrap" role="main">
-        <?php while (have_posts()) : the_post(); ?>
+            <?php while ( have_posts() ) : the_post(); ?>
```
wynnset: This is too indented. Indentation should be 4 spaces all the time.

templates/single-tapestry.php
```diff
@@ -29,9 +29,11 @@ function addNonceToScript()
...
+
 
             <?php get_template_part('content', 'page'); ?>
```
wynnset: This is the correct level of indentation since it's inside the while loop.

templates/single-tapestry.php
```diff
@@ -41,17 +43,25 @@ function addNonceToScript()
                 comments_template();
             endif;
             ?>
+        
+            <?php endwhile; // end of the loop. ?>
```
wynnset: This is too indented again.

templates/single-tapestry.php
```diff
@@ -41,17 +43,25 @@ function addNonceToScript()
...
 
-        <?php endwhile;
-    ?>
+        <div id="tapestry">
+            <!-- Don't render add root button when there's nodes -->
+            <?php if(!doesTapestryHaveNodes(get_the_ID())) : ?>
+                <div id="root-node-btn"><i class="fas fa-plus-circle fa-5x"></i></div>
```
wynnset: Could you please change this to have the text "Add Root Node" either inside the button or beside it (whatever looks good). I just want to highlight it more for new users.

kevinthewebguy: done

templates/single-tapestry.php
kevinthewebguy: fixed

templates/single-tapestry.php
kevinthewebguy: fixed

templates/single-tapestry.php
```diff
@@ -41,17 +43,28 @@ function addNonceToScript()
...
             endif;
             ?>
+        
+        <?php endwhile; // end of the loop. ?>
 
-        <?php endwhile;
-    ?>
+        <div id="tapestry">
+            <!-- Don't render add root button when there's nodes -->
+            <?php if(!doesTapestryHaveNodes(get_the_ID())) : ?>
```
wynnset: Actually could you move this into `tapestry.js` so it's dynamically added in instead? It's better like that since we're adding everything like that. Thanks!

templates/single-tapestry.php
```diff
@@ -41,17 +43,28 @@ function addNonceToScript()
...
+                <div id="root-node-container">
+                    <div id="root-node-btn"><i class="fas fa-plus-circle fa-5x"></i></div>
+                    <div id="root-node-label">Add Root Node</div>
+                <div>
+            <?php endif; ?>
+        </div>
+        
+        <?php require "modal/create-new-node-modal.php";?>
```
wynnset: Could you please pass the full URL to this template into tapestry.js and then load it in tapestry.js via an ajax call using the URL? This way, you don't have to include it here. You could also consider changing it to be an html file instead.

kevinthewebguy: done.

templates/single-tapestry.php
kevinthewebguy: done.

PR #40, opened 2019-05-28
9 load tapestries from db

kevinthewebguy:


wynnset:
@kevinthewebguy what is going on in this PR? There doesn't seem to be any changes other than 1 line being removed. I wouldn't want us to merge a branch that's essentially just 1 line changed. Is there something missing here?

kevinthewebguy:
It's the index to the d3 code and the removal of the jsonUrl. That is all for this issue.

wp-content/plugins/tapestry/templates/single-tapestry.php
```diff
@@ -34,7 +34,6 @@
 			<script src="<?php echo plugin_dir_url(__FILE__) ?>tapestry-d3/libs/h5p-resizer.min.js" charset="UTF-8"></script>
 
             <script>
-				var jsonUrl = "<?php echo plugin_dir_url(__FILE__) ?>tapestry-d3/tapestry.json";
```
wynnset: Please pass the URL into the JS the same way so that we don't have to hardcode it in. Something like 
`var apiUrl = "<?php get_rest_url(null, 'wp-json/tapestry-tool/v1'); ?>";` should work

kevinthewebguy: Good idea, using the get_rest_url function now.

PR #47, opened 2019-06-03
WIP - 23 Create Test data class

### LINKED ISSUE #23: Create class for seed data
Crate a class (TapestryTestData) to put seed data in to test for existing features such as unlock nodes, depth slider

kevinthewebguy:
fixes #23

### DIFF
--- a/controller/class.tapestry-test-data.php
+++ b/controller/class.tapestry-test-data.php
@@ -1,0 +1,321 @@
+<?php
+
+require __DIR__ . '/class.tapestry-controller.php';
+
+/**
+ * Tapestry Endpoint Permissions
+ *
+ */
+class TapestryTestData
+{
+    private $_testTapestryData = '{
+        "settings": {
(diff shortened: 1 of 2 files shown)


kevinthewebguy:
It looks like Andrew is removing the getLinks function (which was to convert hardcoded ids from tapestry.json to the ones from the database. He does not want to replace it with another function to set up the environment. So this issue is blocked until further notice.

kevinthewebguy:
Andrew has agreed to create a function to save tapestry to the database for seed data. Will need to wait for the endpoint to be finished.

wynnset:
As per my conversation with @kevinthewebguy we decided any functionalities related to adding data to the database should be in this class, not in other classes.

kevinthewebguy:
The refactoring of the backend has made the current changes out of date. Will need to redo this from scratch but it is not part of the current sprint.

PR #50, opened 2019-06-06
Create endpoint for linking nodes

anhduc130:


wynnset:
Please also rebase your branch.

controller/class.tapestry-controller.php
```diff
@@ -43,6 +43,10 @@ class TapestryController
         'INVALID_CHILD_NODE' => [
             'MESSAGE'   => 'Node is not a child of the tapestry',
             'STATUS'    => ['status' => 400]
+        ],
+        'INVALID_TAPESTRY' => [
```
wynnset: The key here is not descriptive of the error. Maybe `INVALID_NEW_TAPESTRY` would be better.

controller/class.tapestry-controller.php
```diff
@@ -59,16 +63,20 @@ public function __construct($postId = 0)
...
      */
-    public function updateTapestry($tapestry)
+    public function addTapestry($tapestry)
     {
+        if (isset($tapestry->links) || count($tapestry->nodes) > 1) {
+            return $this->_throwsError('INVALID_TAPESTRY');
+        }
         if (!$this->postId) {
             $this->postId = $this->_updatePost($tapestry, 'tapestry');
         }
```
wynnset: What about the case that there is already a tapestry saved with the passed in ID? You would be adding _additional_ nodes, which is unexpected. Maybe you should check to make sure the tapestry is empty if `$this->postId` is passed in, otherwise return an error.

controller/class.tapestry-controller.php
```diff
@@ -333,32 +339,23 @@ private function _isValidTapestryGroup($groupMetaId)
...
-            }
+            $nodePostId = $this->_updatePost($node, 'tapestry_node');
+            $metadata = $this->_makeMetadata($node, $nodePostId);
+            $node->id = add_post_meta($this->postId, 'tapestry_node', $metadata);
             update_post_meta($nodePostId, 'tapestry_node_data', $node);
         }
     }
 
-    private function _updateGroups($groups)
+    private function _addGroups($groups)
```
wynnset: It would look cleaner if you change this function to `_addGroup` and accept a single group instead and do the loop over groups in the caller when needed.

controller/class.tapestry-controller.php
anhduc130: Yeah, I just realize that there will be no such scenario that `$this->postId` would be set when this method is called (the function in `endpoints.php` didn't initialize the postId when calling this method). I'd say it's safe to remove this check.

wynnset: Then, I suggest adding a check to make sure `$this->postId` is not passed in.

controller/class.tapestry-controller.php
```diff
@@ -43,54 +43,86 @@ class TapestryController
...
             if (!isset($node->permissions)) {
                 $node->permissions = (object)self::NODE_PERMISSIONS['DEFAULT'];
             }
         }
 
-        $this->_updateNodes($tapestry->nodes);
-        $this->_updateGroups($tapestry->groups);
+        foreach ($tapestry->nodes as $node) {
+            if ($this->_isValidTapestryNode($node->id)) {
+                return $this->_throwsError('NODE_ALREADY_EXISTS');
```
wynnset: It's not a good idea to return an error after you have made changes to the database. I suggest moving the validation for both nodes and groups up before you call `_updatePost`

controller/class.tapestry-controller.php
```diff
@@ -43,54 +43,86 @@ class TapestryController
...
+            }
+            $this->_addNode($node);
+        }
+
+        foreach ($tapestry->groups as $group) {
+            if ($this->_isValidTapestryGroup($group->id)) {
+                return $this->_throwsError('GROUP_ALREADY_EXISTS');
```
wynnset: As per the previous comment, move the validation up.

controller/class.tapestry-controller.php
```diff
@@ -43,54 +43,86 @@ class TapestryController
...
+            'STATUS'    => ['status' => 500]
         ]
     ];
     private $postId;
 
     /**
      * Constructor
      */
-    public function __construct($postId = 0)
+    public function __construct($postId)
```
wynnset: Why did you remove the default? This means this argument will be mandatory, so even new tapestries will need to pass in 0 to work.

controller/class.tapestry-controller.php
```diff
@@ -43,54 +43,86 @@ class TapestryController
...
         }
         if (!isset($tapestry->nodes)) {
             $tapestry->nodes = [];
         }
         if (!isset($tapestry->groups)) {
             $tapestry->groups = [];
         }
+
+        $this->postId = $this->_updatePost($tapestry, 'tapestry');
+
```
wynnset: I suggest you dedicate this function solely to adding a new tapestry, not its nodes, groups, etc.

controller/class.tapestry-controller.php
anhduc130: What I am trying to do is that if there is no `$postId` passed in or `$postId` is valid, I would set it to NULL. The methods will then check for the $postId.

controller/class.tapestry-controller.php
anhduc130: Yeah. I guess we are going to remove this anyway.

controller/class.tapestry-controller.php
anhduc130: Set the parameter default to NULL.

controller/class.tapestry-controller.php
```diff
@@ -43,56 +43,76 @@ class TapestryController
...
+            'STATUS'    => ['status' => 400]
+        ],
+        'LINKS_EXIST_IN_NEW_TAPESTRY' => [
+            'MESSAGE'   => 'Links exist when creating a new Tapestry',
+            'STATUS'    => ['status' => 400]
+        ],
+        'POST_ID_ALREADY_SET' => [
+            'MESSAGE'   => 'PostID is set when creating a new Tapestry',
+            'STATUS'    => ['status' => 500]
         ]
```
wynnset: Could you please phrase your error messages to tell them what needs to be changed for their request to be successful? So "Settings are required to created a new Tapestry", "Nodes should not be passed in when creating a new Tapestry", etc.

controller/class.tapestry-controller.php
```diff
@@ -43,56 +43,76 @@ class TapestryController
...
     ];
     private $postId;
 
     /**
      * Constructor
      */
-    public function __construct($postId = 0)
+    public function __construct($postId = null)
```
wynnset: Documentation for [wp_insert_post](https://developer.wordpress.org/reference/functions/wp_insert_post/) states that it requires postId to be 0 to create a new post. I suggest you change this to 0 and also do the same below since you may be passing this to the wp_insert_post function at some point.

controller/class.tapestry-controller.php
```diff
@@ -43,56 +43,76 @@ class TapestryController
...
     {
-        if (!$this->postId) {
-            $this->postId = $this->_updatePost($tapestry, 'tapestry');
+        if ($this->postId) {
+            return $this->_throwsError('POST_ID_ALREADY_SET');
         }
-        if (!isset($tapestry->nodes)) {
-            $tapestry->nodes = [];
+        if (empty($tapestry->settings)) {
+            return $this->_throwsError('INVALID_NEW_TAPESTRY');
```
wynnset: Again, I suggest you change this error index to be more descriptive like `'SETTINGS_MISSING_IN_NEW_TAPESTRY'` or something

controller/class.tapestry-controller.php
```diff
@@ -211,7 +235,36 @@ public function updateTapestrySettings($settings)
...
+    }
+
+    /**
+     * Add A Tapestry Link
+     * 
+     * @param   Object  $link   Tapestry link
+     * 
+     * @return  Object  $link 
+     */
+    public function addTapestryLink($link)
```
wynnset: Could you please move this function before the `updateTapestrySettings` function so it's a bit more organized? All the add functions should be beside eachother as much as possible.

controller/class.tapestry-controller.php
wynnset: Please update the message for the last item `'POST_ID_ALREADY_SET'` as well.

PR #63, opened 2019-06-26
d3 reference for bug fix

kevinthewebguy:
Also includes stuff from depth slider and nonce


kevinthewebguy:
PR #30 will fix the bug indtead.

PR #76, opened 2019-07-05
Refactor into separate classes that are abstracted

### LINKED ISSUE #75: Refactor into separate classes that are abstracted and use non-static methods
**Refactor as follows:**
- Separate nodes, links, and groups into their own controller function
- For each new controller class, create private properties to keep track of the data for that class
- Public classes should be abstract and not use any Wordpress functionality
- Wordpress functions should only be called in private functions

anhduc130:
Fixes #75, #28


controllers/class.tapestry-controller.php
```diff
@@ -0,0 +1,337 @@
...
+require_once dirname(__FILE__) . "/../interfaces/interface.tapestry-controller.php";
+
+/**
+ * Add/update/retrieve a Tapestry
+ * 
+ */
+class TapestryController implements ITapestryController
+{
+    private $postId;
+
```
wynnset: Missing other private properties for nodes, links, groups, and settings (including rootId).

controllers/class.tapestry-controller.php
```diff
@@ -0,0 +1,337 @@
...
+        return new TapestrySettingController($this->postId);
+    }
+
+    private function _addTapestry($tapestry)
+    {
+        $this->postId = TapestryHelpers::updatePost($tapestry, 'tapestry');
+
+        $tapestry->links = [];
+        $tapestry->nodes = [];
+        $tapestry->groups = [];
```
wynnset: These should be initialized in the constructor

controllers/class.tapestry-controller.php
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    /**
+     * Add A Tapestry
+     * 
+     * @param   Object  $tapestry   Tapestry
+     * 
+     * @return  Object  $tapestry
+     */
+    public function save($tapestry)
```
wynnset: Please create a new method called `set($tapestry)` for setting the tapestry content. You could potentially move the validation in there as well.

controllers/class.tapestry-controller.php
```diff
@@ -0,0 +1,337 @@
...
+    {
+        if (empty($tapestry->settings)) {
```
wynnset: These validation should be put in an if statement `$this->postId == 0`

controllers/class.tapestry-controller.php
```diff
@@ -0,0 +1,337 @@
...
+     * 
+     * @return  Object  $tapestry
+     */
+    public function get()
+    {
+        if (!$this->postId) {
+            return TapestryErrors::throwsError('INVALID_POST_ID');
+        }
+
+        return $this->_getTapestry();
```
wynnset: The tapestry should automatically be loaded in the constructor. Here, you should not need to do any database operations. Rather, you should just be able to return the tapestry using the private properties of this class that were already set in the constructor.

controllers/class.tapestry-controller.php
```diff
@@ -0,0 +1,337 @@
...
+        if (is_wp_error($node)) {
+            return $node;
+        }
+
+        $tapestry = get_post_meta($this->postId, 'tapestry', true);
+
+        if (!isset($tapestry->nodes)) {
+            $tapestry->nodes = [];
+        }
+
```
wynnset: Lines 83-87 should not be needed if you initialize and load the tapestry in the constructor.

controllers/class.tapestry-controller.php
```diff
@@ -0,0 +1,337 @@
...
+        array_push($tapestry->nodes, $node->id);
```
wynnset: Here you should be able to push to `$this->nodes` instead.

controllers/class.tapestry-controller.php
```diff
@@ -0,0 +1,337 @@
...
+
+        if (empty($tapestry->rootId)) {
+            $tapestry->rootId = $tapestry->nodes[0];
```
wynnset: Again, this should be saved to the class private property.

controllers/class.tapestry-controller.php
```diff
@@ -0,0 +1,337 @@
...
+     * 
+     * @param  Object   $link   Tapestry link
+     * 
+     * @return  Object  $link   Tapestry link
+     */
+    public function addLink($link)
+    {
+        $tapestryLinkController = new TapestryLinkController($this->postId);
+        $link = $tapestryLinkController->save($link);
+
```
wynnset: Also push the link to `$this->links`

controllers/class.tapestry-controller.php
```diff
@@ -0,0 +1,337 @@
...
+    /**
+     * Add a new link
+     * 
+     * @param  Object   $link   Tapestry link
+     * 
+     * @return  Object  $link   Tapestry link
+     */
+    public function addLink($link)
+    {
+        $tapestryLinkController = new TapestryLinkController($this->postId);
```
wynnset: As discussed, we probably don't need a class for the links.

controllers/class.tapestry-controller.php
```diff
@@ -0,0 +1,337 @@
...
+        $tapestry = get_post_meta($this->postId, 'tapestry', true);
+
+        if (!isset($tapestry->groups)) {
+            $tapestry->groups = [];
+        }
+
+        array_push($tapestry->groups, $group->id);
+
+        update_post_meta($this->postId, 'tapestry', $tapestry);
+
```
wynnset: The same comments from `addNode()` apply here.

controllers/class.tapestry-controller.php
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    /**
+     * Get the group controller
+     * 
+     * @return  Object  $group  group controller
+     */
+    public function getGroups()
+    {
+        return new TapestryGroupController($this->postId);
```
wynnset: An instance of TapestryGroup class should be a representation of 1 group. So if you want to get groups for this tapestry, please iterate over the groups as per `$this->groups` and load them one by one into an array that you will then return.

controllers/class.tapestry-controller.php
```diff
@@ -0,0 +1,337 @@
...
+     * Get the setting controller
+     * 
+     * @return  Object  $setting   setting controller
+     */
+    public function getSettings()
+    {
+        return new TapestrySettingController($this->postId);
+    }
+
+    private function _addTapestry($tapestry)
```
wynnset: Change this class to be called `saveToDatabase` instead. And you shouldn't need to pass in any arguments.

controllers/class.tapestry-controller.php
```diff
@@ -0,0 +1,337 @@
...
+    {
+        $this->postId = TapestryHelpers::updatePost($tapestry, 'tapestry');
+
+        $tapestry->links = [];
+        $tapestry->nodes = [];
+        $tapestry->groups = [];
+
+        update_post_meta($this->postId, 'tapestry', $tapestry);
```
wynnset: I recommend creating a private method for gathering all the info (from the class private properties) needed to go into this `$tapestry`.

controllers/class.tapestry-controller.php
```diff
@@ -0,0 +1,337 @@
...
+
+        return $tapestry;
+    }
+
+    private function _formNodeData($nodeData, $nodeMetadata)
```
wynnset: This should be move to the TapestryNode class instead

controllers/class.tapestry-controller.php
```diff
@@ -0,0 +1,337 @@
...
+        if (isset($nodeMetadata->meta_value->imageURL)) {
+            $nodeData->imageURL = $nodeMetadata->meta_value->imageURL;
+        }
+        if (isset($nodeMetadata->meta_value->unlocked)) {
+            $nodeData->unlocked = $nodeMetadata->meta_value->unlocked;
+        }
+        return $nodeData;
+    }
+
+    private function _getTapestry()
```
wynnset: I would change the name of this function to `_loadFromDatabase` and then make sure you are saving the info you are retrieving to the class private properties. You can then call this function from your constructor.

controllers/class.tapestry-controller.php
anhduc130: I don't think we can do this. It's because the `nodes` and `groups` are stored as arrays of IDs in the `tapestry`. When I load the data for the two properties, they will be arrays of IDs and not the actual data. Additionally, I would need to check various permissions to filter what to return to the client. This would require further further data operations in order to load the actual data after filtering.

PR #79, opened 2019-07-12
Fix for saving and retrieving unlocked node per user

wynnset:
- Remove tapestry node unlocked functionality/endpoints since it should be related to the user
- Add endpoint and related functions for setting user node to unlocked
- Retrieve and return user node unlocked status with node progress

### DIFF
--- a/controller/class.user-controller.php
+++ b/controller/class.user-controller.php
@@ -131,9 +131,9 @@ class TapestryUserController
             $default_unlocked_status = $nodeMetadata->unlocked ? true : false;
             $unlocked_value = get_user_meta($this->_userId, 'tapestry_' . $postId . '_node_unlocked_' . $nodeId, true);
             if ($unlocked_value !== null) {
-                $progress->$nodeId->$unlocked = $unlocked_value;
+                $progress->$nodeId->unlocked = $unlocked_value;
             } else {
-                $progress->$nodeId->$unlocked = $default_unlocked_status;
+                $progress->$nodeId->unlocked = $default_unlocked_status;
             }
         }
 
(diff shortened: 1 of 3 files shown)


controller/class.user-controller.php
```diff
@@ -94,17 +108,32 @@ private function _updateUserProgress($postId, $nodeId, $progressValue)
...
+            }
+
+            $nodeMetadata = get_metadata_by_mid('post', $nodeId)->meta_value;
+            $default_unlocked_status = $nodeMetadata->unlocked ? true : false;
+            $unlocked_value = get_user_meta($this->_userId, 'tapestry_' . $postId . '_node_unlocked_' . $nodeId, true);
+            if ($unlocked_value !== null) {
+                $progress->$nodeId->$unlocked = $unlocked_value;
             } else {
-                $progress->$nodeId = 0.0;
+                $progress->$nodeId->$unlocked = $default_unlocked_status;
```
anhduc130: Should this be `$progress->$nodeId->unlocked`???

endpoints.php
```diff
@@ -119,6 +111,13 @@
             'callback'              => 'updateProgressByNodeId',
         ]
     ],
+    'UPDATE_TAPESTRY_UNLOCKED' => (object)[
+        'ROUTE'     => 'users/unlocked',
```
anhduc130: Should this be `users/(?P<userId>[\d]+)/progress`?

wynnset:
@fordat please create a new branch from master and do this again from that branch since the code has been completely refactored. It's not worth it to do a merge because of the amount of things that have changed.

PR #92, opened 2019-07-30
Refactor frontend using Vue JS

anhduc130:


templates/tapestry-d3-vue/src/components/NodeModal.vue
```diff
@@ -0,0 +1,242 @@
...
+          <div class="modal-header">
+            <h4 class="modal-title" id="createNewNodeModalLabel"></h4>
+          </div>
+          <div class="modal-body" id="createNewNodeModalBody">
+            <div id="add-node-error-msg"></div>
+            <form class="form-container">
+              <label>
+                Title
+                <input
+                  id="add-node-title-input"
```
wynnset: Please remove all IDs from inputs and bind them to a model instead.

templates/tapestry-d3-vue/src/components/NodeModal.vue
```diff
@@ -0,0 +1,242 @@
...
+                <div class="dropdown">
+                  <select id="mediaType" name="mediaType">
+                    <option value="default">Select type:</option>
+                    <option value="video" selected>Video</option>
+                    <!--<option value="image">image</option>-->
+                  </select>
+                </div>
+              </label>
+              <label>
+                Media Format
```
wynnset: This has been removed in master. Please update.

templates/tapestry-d3-vue/src/components/NodeModal.vue
```diff
@@ -0,0 +1,242 @@
...
+                  <select id="mediaFormat" name="mediaFormat" v-model="selectedMediaFormat">
+                    <option value>Select format:</option>
+                    <option value="mp4">MP4</option>
+                    <option value="h5p">H5P</option>
+                    <!--<option value="jpeg">JPEG</option>-->
+                  </select>
+                </div>
+              </label>
+              <div
+                id="contents-details"
```
wynnset: If ID is not needed, please remove.

templates/tapestry-d3-vue/src/components/NodeModal.vue
```diff
@@ -0,0 +1,242 @@
...
+                class="content-details"
+                style="display: none"
+                v-show="showContentDetails"
```
wynnset: Please use actual logic here instead of calling a function. Or otherwise create a `calculated` value that can be called if the logic is very complex.

templates/tapestry-d3-vue/src/components/NodeModal.vue
```diff
@@ -0,0 +1,242 @@
...
+              >
+                <h3>Content Details</h3>
+                <div id="mp4-content" class="mp4-content" v-show="showMp4Content">
```
wynnset: Instead of "showMp4Content", please use actual logic (e.g. formData.mediaType == "mp4" or something like that after you bind all your form elements to models). This comment applies similarly to the rest of your code as well.

templates/tapestry-d3-vue/src/components/NodeModal.vue
```diff
@@ -0,0 +1,242 @@
...
+  data() {
+    return {
+      selectedMediaFormat: '',
+      showMp4Content: false,
+      showH5pContent: false,
+      showContentDetails: false
+    }
+  },
+  watch: {
+    selectedMediaFormat(value) {
```
wynnset: You shouldn't need these if you bind your inputs to models

templates/tapestry-d3-vue/src/components/RootNodeButton.vue
```diff
@@ -0,0 +1,52 @@
...
+  name: 'root-node-button',
+  data() {
+    return {}
+  },
+  methods: {
+    showModal() {
+      console.log('here')
+      // Populate title
+      $('#createNewNodeModalLabel').text("Add root node")
+      $("#submit-add-new-node").hide()
```
wynnset: Please do not use this hide/show logic as it's very long and complicated. Bind your elements to models and then they should automatically be filled in / updated.

templates/tapestry-d3-vue/src/components/RootNodeButton.vue
```diff
@@ -0,0 +1,52 @@
...
+
+      // Show the modal
+      $("#createNewNodeModal").modal()
+    }
+  }
+}
+</script>
+
+<style scoped>
+#root-node-btn {
```
wynnset: If you absolutely have to use IDs, please prefix all of them with`tapestry-`. I recommend having one wrapper with an ID and then use classes for the nested elements. That way, you can do something like `#tapestry-add-root-node-wrapper > .btn`, `#tapestry-add-root-node-wrapper > label`, etc.

templates/tapestry-d3-vue/src/components/NodeModal.vue
anhduc130: I found this being used in tapestry.css. Maybe it's for styling purposes?

templates/tapestry-d3-vue/src/components/RootNodeButton.vue
anhduc130: Could we please leave this for later battle when we refactor to get rid of all the jQuery scripts? Ideally, I think there should be one single button that does all the edit/add rootNode/nodes in the modal.

templates/tapestry-d3-vue/src/components/RootNodeButton.vue
anhduc130: `#root-node-btn` is now also used for styling in tapestry.css. Could we leave this for later as well?

templates/tapestry-d3-vue/src/components/RootNodeButton.vue
wynnset: What is the point of using VueJS if we don't use models? It's very weird for me to not use models as they are very easy to use and make life a whole lot simpler. I would really prefer we use v-model instead of manually having to fill in fields like this (unless you are out of time, in which case I'll let you prioritize)

templates/tapestry-d3-vue/src/components/RootNodeButton.vue
wynnset: Yes, you can leave this for later - just make sure you make a list of the things that still need to be done.

wynnset:
Please make sure we prefix all the event names with "tapestry-" since it's global.

wynnset:
Closing PR since all pending PRs were now moved to be merged into master

PR #101, opened 2019-08-14
100 settings modal wp

### LINKED ISSUE #100: Add settings for the whole tapestry
We want to give the author to control certain aspects of their Tapestry. We need to add a setting button in the interface and clicking that should pop-up a modal with the settings:
- Auto-layout
- Background URL (string)

Auto-layout is currently a variable in tapestry.js. We need to remove that variable and instead use this setting instead
Background URL should be applied to the body. Currently this feature doesn't exist

kevinthewebguy:
fixes #100


kevinthewebguy:
Migrating code to vue. closing PR

PR #131, opened 2019-09-23
Fixed root node being very large on first load

### LINKED ISSUE #122: The root node is very large after being created
Steps to reproduce:
- Create an empty tapestry
- Add root node

Result:
- The root node is very large and almost covers the entire screen

Workaround:
- Refresh the page

narendrasss:
Fixes #122 

Somewhat hacky solution.

### DIFF
--- a/templates/tapestry.js
+++ b/templates/tapestry.js
@@ -1897,6 +1897,8 @@ function tapestryTool(config){
         };
     }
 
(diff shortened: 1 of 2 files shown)


narendrasss:
I think Andrew's solution #130 solves the core issue.

PR #141, opened 2019-09-24
140 Media type of H5P is shown as 'video'

anhduc130:


wynnset:
Going to close this as per our conversation. We need to take a different approach to fixing this and make sure it works well with tapestry.js

PR #151, opened 2019-10-01
Added tyde menu

narendrasss:
Implements #149


templates/vue/src/components/Tyde.vue
```diff
@@ -1,20 +1,97 @@
...
 </script>
 
-<style scoped>
-
+<style>
+:root {
+  --gray: #404040;
+  --light-gray: #787878;
+  --med-gray: #535353;
+  --light-blue: #1eade1;
```
wynnset: If these are global variables, we might want to prefix them with something like `tyde-` or `tapestry-` so it doesn't cause conflicts with potentially other things naming the variables the same thing.

narendrasss:
Waiting on tyde #160

narendrasss:
@wynnset I fixed the logs, but I need to update the way the menu handles lightboxes since it is now out of `tapestry.js`.

wynnset:
Has this been tested and working with mp4 videos, h5p videos, and also other content types?

wynnset:
Needs to be tested with H5P to make sure it's working. Otherwise good to go.

wynnset:
Error: When I try to change the option "Allow skipping video if user has not watched at least once", the application throws console errors and fails to complete the operation.

PR #156, opened 2019-10-07
Refactored lightbox to Vue

narendrasss:
Implements #153


narendrasss:
TODO: Update media icons, analytics and view mode

narendrasss:
BUG: Video duration does not get saved if the user exits/refreshes the tab while the lightbox is still open.

templates/tapestry.js
```diff
@@ -1374,329 +1384,28 @@ function tapestryTool(config){
...
-                updateMediaIcon(id, mediaType);
-            }
-        }, 1000);
-    
-        if (mediaType === "video") {
-            var loadEvent = 'load';
-            if (mediaFormat == "mp4") {
-                loadEvent = "loadstart";
+
+    this.updateChildren = function(id, video) {
```
wynnset: Please leave a comment to explain what this function does. Also, if it's possible, I prefer to have functions in their own components rather than in this file as much as possible. We want to eventually get rid of this altogether, so if you could move this function into the component that uses it, it would be better (potentially put into a service and consume it in the component). You can always `setDataset()` from the component to push changes to `tapestry.js`.

templates/tapestry.js
```diff
@@ -1374,329 +1384,28 @@ function tapestryTool(config){
...
-                var para = document.createElement("p");
-                para.setAttribute("id", "text-light-box-paragraph-text");
-                para.innerHTML = paragraphArray[i].replace('\n','<br>');
-                paraDiv.appendChild(para);
-                paragraphSection.appendChild(paraDiv);
-            }
-            lightboxContent.appendChild(paragraphSection);
-        }
-        return $(lightboxContent);
+    this.saveVideoProgress = function(id, currentTime, duration) {
```
wynnset: Same as my previous comment, please see if you could move this out of `tapestry.js` since it's not used anywhere else in this file.

templates/tapestry.js
```diff
@@ -1374,329 +1384,28 @@ function tapestryTool(config){
...
+        updateViewedValue(id, currentTime, duration);
+        updateViewedProgress();
     }
+
+    this.recordAnalyticsEvent = recordAnalyticsEvent;
     
+    this.exitViewMode = exitViewMode;
     // Builds the view mode, including functionality to
-    function changeToViewMode(lightboxDimensions) {
+    this.changeToViewMode = function changeToViewMode(lightboxDimensions) {
```
wynnset: Same as my previous comments, please see if it's possible to move these out of here and to view instead where they fit logically.

templates/vue/src/components/Lightbox.vue
```diff
@@ -0,0 +1,249 @@
...
+      }
+    },
+  },
+  async mounted() {
+    const node = await this.tapestryApiClient.getNode(this.nodeId)
+    const meta = await this.tapestryApiClient.getNodeProgress(this.nodeId)
+    node.typeData.progress[0].value = meta.progress
+    node.typeData.progress[1].value = 1.0 - meta.progress
+
+    const settings = await this.tapestryApiClient.getH5pSettings()
```
wynnset: I believe this is a tapestry-wide setting that should be loaded at the beginning when the tapestry loads, so I don't think we should be calling it every time for a singe node as it will always give us the same data. This should be loaded when the tapestry is initially loaded and then made available to this component.

templates/vue/src/components/Lightbox.vue
```diff
@@ -0,0 +1,249 @@
...
+  background: #eee;
+  color: #333;
+  padding: 1em;
+}
+
+.content-text .media-wrapper {
+  height: 100%;
+  overflow: scroll;
+  background: transparent;
+}
```
wynnset: Couldn't this be moved into the TextMedia component?

templates/tapestry.js
narendrasss: Unfortunately this works with d3 so it has to be kept in tapestry.js until we eventually move tapestry into a vue component.

templates/tapestry.js
narendrasss: Same as above, changeToViewMode does some stuff with d3.

templates/vue/src/components/Lightbox.vue
narendrasss: Unfortunately no because it deals with the styles in the outer lightbox container. I'll see if I can make any structural changes so that I can move this around.

templates/tapestry.js
narendrasss: I looked into this a little bit more and to me I feel like I would keep it there for now at least until this file becomes its own component. For now I added a comment that describes what this function does.

While this function **is** only used by the `VideoMedia` component, it does manipulate the entire Tapestry. So we have 2 options, which is to pass in the entire tapestry to `VideoMedia`, or move the functionality to `Tapestry.vue`. I feel that passing the entire tapestry is wasteful, and `Tapestry.vue` is bloated as it is.

But let me know what you think.

wynnset: Ok, I see your point. We can wait until this becomes a component to make that change. Thanks

templates/tapestry.js
wynnset: Only `updateViewedProgress()` works with d3, so I think it's better if we move the `updateViewedValue` function out of tapestry.js and into TapestryAPI and then use it in Vue directly. We can make the `updateViewedProgress()` accessible from outside so that we can just call that function from Vue. 

Again, I'm trying to make sure we move as much stuff out of this file as possible so that it becomes more realistic to refactor this into a component, otherwise it will remain too big for any one person to attempt to turn it into a component.

Issue #118, opened 2019-09-18
Add linter


narendrasss:
Add linter + automatic styling for vscode

PR #158, opened 2019-10-09
Linted Master

narendrasss:
Added config files from #118 and linted all js files (except tapestry.js)


templates/vue/src/App.vue
```diff
@@ -5,12 +5,12 @@
...
 <script>
-import Tapestry from './components/Tapestry'
+import Tapestry from "./components/Tapestry"
 
 export default {
-  name: 'app',
+  name: "app",
   components: {
-    Tapestry,
+    Tapestry
```
wynnset: Is there any way we can keep these commas? Is there a setting for this? The reason is I want to minimize potential merge conflicts with the TYDE branch and this is a culprit of those.

templates/vue/src/components/NodeModal.vue
```diff
@@ -1,74 +1,107 @@
...
                 v-model="node.typeData.mediaURL"
+                placeholder="Enter URL for MP4 Video"
                 required
               />
             </b-form-group>
-            <b-form-group label="Video Duration" v-show="node.mediaType === 'video'">
+            <b-form-group
+              v-show="node.mediaType === 'video'"
+              label="Video Duration"
+            >
```
wynnset: How come the first b-form-group element is in one line and the second one is in 2 lines? Anyway to keep this consistent? I'm guessing it's due to the length of the line, but I am still curious to see if it's possible to keep it consistent somehow.

templates/vue/src/components/NodeModal.vue
```diff
@@ -77,16 +110,16 @@
         <b-tab title="Appearance">
           <div id="modal-appearance">
             <b-form-group>
-              <b-form-checkbox
-                v-model="addThumbnail"
-              >Add a thumbnail</b-form-checkbox>
+              <b-form-checkbox v-model="addThumbnail"
+                >Add a thumbnail</b-form-checkbox
+              >
```
wynnset: This looks really weird. Either put it all on one line or put "Add a thumbnail" on its own line with the opening tag and closing tag also on their own line.

templates/vue/src/components/NodeModal.vue
```diff
@@ -166,156 +225,186 @@
...
-        { value: 'video', text: 'Video' },
-        { value: 'h5p', text: 'H5P' },
-        { value: 'url-embed', text: 'URL Embed' },
-      ],
-      formErrors: '',
-      maxDescriptionLength: 250,
-      addThumbnail: false,
-    }
-  },
+  name: "NodeModal",
```
wynnset: This name change doesn't break anything?

templates/vue/src/components/SettingsModal.vue
```diff
@@ -1,83 +1,105 @@
...
 export default {
-  name: 'settings-modal',
-  data() {
-    return {
-      currentSettings: {},
-      backgroundUrl: '',
-      autoLayout: false,
-    }
-  },
+  name: "settings-modal",
```
wynnset: How come this was changed to camelCase in the other modal, but not here? I thought it was a lint setting.. Let's keep it consistent.

templates/vue/src/components/Tapestry.vue
```diff
@@ -1,17 +1,42 @@
...
+    <RootNodeButton
+      v-show="tapestryLoaded && !tapestry.rootId"
+      @add-root-node="addRootNode"
+    />
     <NodeModal
       :node="populatedNode"
-      :modalType="modalType"
-      :rootNodeTitle="selectedNode.title"
+      :modal-type="modalType"
+      :root-node-title="selectedNode.title"
```
wynnset: This doesn't break anything?

narendrasss: I didn't know this either, but apparently not.

See the [vue docs here](https://vuejs.org/v2/guide/components-props.html#Prop-Casing-camelCase-vs-kebab-case).

PR #162, opened 2019-10-17
Improve overall performance

narendrasss:
Rewrote `getChildren` and `addDepthToNode` as well as several changes to their respective callers to improve overall tapestry performance.

Load time for North American LGTBQ+ **before** PR: ~200s
Load time for North American LGTBQ+ **after** PR: ~4s

Possible further optimizations:

* If a node has no children at depth `n`, then it cannot have any children at depth `n+1`. Currently `getChildren` still tries to check for children at all depths `> n`.

### DIFF
--- a/templates/tapestry.js
+++ b/templates/tapestry.js
@@ -2119,7 +2119,7 @@ function tapestryTool(config){
             node.depth = depth;
         }
 
-        const children = getChildren(id, 1);
+        const children = getChildren(id, 0);
         children.forEach(child => {
             if (visited.includes(child)) {
                 const childNode = tapestry.dataset.nodes[findNodeIndex(child)];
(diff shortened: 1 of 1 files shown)


templates/tapestry.js
```diff
@@ -94,6 +94,34 @@ function tapestryTool(config){
...
+//                          (tapestry.dataset.links[i].target == 4550 && tapestry.dataset.links[i].source != 4190)
+//                        || tapestry.dataset.links[i].source == 4542
+                        ) {
+                        tapestry.dataset.links.splice(i,1);
+                    }
+                }
+                */
+            }
+        }
+        // END TEMPORARY FIX
```
wynnset: Remember not to add in any dev code when you commit :)

PR #165, opened 2019-10-21
Add authenticated permission

### LINKED ISSUE #164: Add authenticated permission
Add a new type of permission under public called authenticated users.

narendrasss:
Fixes #164


wynnset:
This is not working as per my test with an existing node that already had permissions set. The rows are not sorted correctly (authenticated appears last instead of after "public") and the "read" permission that is set for public does not apply to individual users.

PR #173, opened 2019-10-25
172 do not allow icons to move with drag

### LINKED ISSUE #172: Do not allow icons to move with click-drag
Allow `administrator` to set if the users are allowed to drag the nodes.

anhduc130:
Fixes #172


templates/single-tapestry.php
```diff
@@ -3,6 +3,40 @@
 /*
  Template Name: Tapestry Page Template
  */
+/**
```
anhduc130: I basically swap the order of function calls here so that Vue.js could use the `wpApiSettings` data provided by Wordpress.

anhduc130:
Synced up the branch with master and tested. I will merge this PR in.

PR #175, opened 2019-10-27
174 Delete associated links when deleting a node

### LINKED ISSUE #174: When deleting a node, delete associated links too

anhduc130:
Fixes #174


templates/tapestry.js
```diff
@@ -541,7 +541,6 @@ function tapestryTool(config){
                             url: apiUrl + "/tapestries/" + config.wpPostId + "/nodes/" + nodeId,
                             method: API_DELETE_METHOD,
                             success: function() {
-                                deleteLink(tapestry.dataset.links[linkToBeDeleted].source.id, tapestry.dataset.links[linkToBeDeleted].target.id, true, spliceIndex);
```
wynnset: I see there is a for loop here that is going over the links. Is this loop still necessary??

anhduc130: I am not sure what `loop` you are referring to here. Do you mean the `for loop` in the `deleteLink()`?

PR #178, opened 2019-10-29
Add completed flag

### LINKED ISSUE #171: Add 'completed' flag for nodes

narendrasss:
Fixes #171


templates/vue/src/components/Lightbox.vue
```diff
@@ -91,6 +90,12 @@ export default {
   computed: {
     ...mapState(["h5pSettings"]),
     ...mapGetters(["getNode"]),
+    node() {
+      return this.getNode(this.nodeId)
+    },
+    canSkip() {
+      return this.node.completed || this.node.skippable
```
wynnset: I think this.node.skippable would return false because it would be undefined for existing nodes. We want to make sure that all existing nodes (and new nodes) are skippable unless the user turns it off and makes it non-skippable. This should be changed to !== false

templates/tapestry.js
```diff
@@ -151,6 +151,7 @@ function tapestryTool(config){
                 jQuery.get(USER_NODE_PROGRESS_URL, { "post_id": config.wpPostId }, function(retrievedUserProgress) {
 
                     if (retrievedUserProgress && !isEmptyObject(retrievedUserProgress)) {
+                        console.log(retrievedUserProgress)
```
wynnset: Please remove console log

PR #190, opened 2019-11-06
Add module and stage components

narendrasss:
Implements the module and stage component functionalities as described by the asana tasks:

- [Module](https://app.asana.com/0/1126491658233864/1146730222278364)
- [Stage](https://app.asana.com/0/1126491658233864/1146730222278365)

Styling is not yet implemented.

**Questions**

- What should we display if we're at the last stage of the module? Should the next button be shown? If it is, where would the next button go to?


templates/vue/src/components/tyde/TydeStage.vue
```diff
@@ -0,0 +1,188 @@
...
+    <lightbox v-if="isLightboxOpen" :node-id="lightboxId" @close="closeLightbox" />
+  </div>
+</template>
+
+<script>
+import { mapGetters } from "vuex"
+import Lightbox from "../Lightbox"
+import Helpers from "../../utils/Helpers"
+
+import StageBackground from "../../assets/stage-bg.png"
```
wynnset: This background should come from the thumbnail of the stage node, not hard coded

templates/vue/src/components/tyde/TydeStage.vue
```diff
@@ -0,0 +1,188 @@
...
+import ButtonOne from "../../assets/tyde-button-1.png"
+import ButtonTwo from "../../assets/tyde-button-2.png"
+import ButtonThree from "../../assets/tyde-button-3.png"
+import ButtonFour from "../../assets/tyde-button-4.png"
```
wynnset: These buttons should come from the thumbnail of each question node (i.e. the child nodes of the stage node), not hardcoded. Please remove all these binaries as well.

templates/vue/src/components/tyde/TydeStage.vue
narendrasss: Thanks, fixed.

templates/vue/src/components/tyde/TydeStage.vue
narendrasss: Thanks, fixed.

wynnset:
@narendrasss thanks for that update. I found another bug: when you are in the stage, if you open the ESC menu and then close it, it plays the video in the background (I can only hear the sound of the video).

PR #197, opened 2019-11-13
Add Quiz Progress

narendrasss:
**Changes**

* Adds a completed flag to quizzes to track whether the given user has completed it.
* Adds endpoint in backend to update a quiz's completion flag given a quiz `id`.

PR #199, opened 2019-11-13
Add End-of-Video Question Set

narendrasss:
Alternative to PR #197.

**Changes**
Compared to the model proposed in Asana, in this model:

* Each node only has _one_ quiz, where each quiz has multiple _questions_.
* Each question has a question _text_ which describes what the question is asking.
* Each question has a series of _answer types_, which could be either `text`, `audio`, or `checklist`. These correspond to either gravity form ids (for text or checklist) or H5P ids (for audio).

I believe that this model better represents the mockups that was given by Linnea, compared to the model discussed in Asana.


classes/class.tapestry-user-progress.php
```diff
@@ -89,6 +89,12 @@ public function complete()
         $this->_complete();
     }
 
+    public function completeQuestion($questionId)
+    {
+        $this->_checkUserAndPostId();
+        $this->completeQuestion($questionId);
```
anhduc130: Shouldn't it be `_completeQuestion`? Seems like a deadly infinite loop.

narendrasss: Thanks, fixed.

templates/vue/src/components/node-modal/QuizModal.vue
```diff
@@ -0,0 +1,132 @@
...
+              @click="deleteQuestion(question.id)"
+            >
+              Delete
+            </b-button>
+          </b-row>
+          <b-form-group label="Question Text">
+            <b-form-input v-model="question.text" />
+          </b-form-group>
+          <b-form-group label="Question Answer Types">
+            <b-form-group label="Textbox ID">
```
wynnset: This should read "Textbox Gravity Form ID"

templates/vue/src/components/node-modal/QuizModal.vue
```diff
@@ -0,0 +1,132 @@
...
+              <b-form-input v-model="question.answers.textId" />
+            </b-form-group>
+            <b-form-group label="Audio ID">
```
wynnset: This should read "H5P Audio Recorder ID"

templates/vue/src/components/node-modal/QuizModal.vue
```diff
@@ -0,0 +1,132 @@
...
+              <b-form-input v-model="question.answers.audioId" />
+            </b-form-group>
+            <b-form-group label="Checklist ID">
```
wynnset: This should read "Checklist Gravity Form ID"

templates/tapestry.js
```diff
@@ -206,7 +206,7 @@ function tapestryTool(config){
         }
 
         this.dataset.nodes = this.dataset.nodes.map(node => {
-            const updatedNode = fillEmptyFields(node, { skippable: true, behaviour: "embed", completed: false })
+            const updatedNode = fillEmptyFields(node, { skippable: true, behaviour: "embed", completed: false, quiz: [] })
```
wynnset: Please do not use ES6 in tapestry.js as it is not using webpack

templates/tapestry.js
```diff
@@ -1880,6 +1880,7 @@ function tapestryTool(config){
             var amountViewed = progressObj[id].progress;
             var amountUnviewed = 1.00 - amountViewed;
             var unlocked = progressObj[id].unlocked;
+            const quizCompletionInfo = progressObj[id].quiz;
```
wynnset: Please do not use ES6 in tapestry.js as it is not using webpack

templates/tapestry.js
```diff
@@ -1888,6 +1889,14 @@ function tapestryTool(config){
                 tapestry.dataset.nodes[index].typeData.progress[0].value = amountViewed;
                 tapestry.dataset.nodes[index].typeData.progress[1].value = amountUnviewed;
                 tapestry.dataset.nodes[index].unlocked = unlocked ? true : false;
+
+                const questions = tapestry.dataset.nodes[index].quiz;
+                if (quizCompletionInfo) {
+                    Object.entries(quizCompletionInfo).forEach(([questionId, isCompleted]) => {
+                        const question = questions.find(question => question.id === questionId);
+                        question.completed = isCompleted;
+                    })
```
wynnset: Please do not use ES6 in tapestry.js as it is not using webpack

wynnset:
**Problem:** Seems like the form is not actually getting submitted when the user submits the form right now. We need to make sure it gets submitted with AJAX.

wynnset:
@narendrasss PR #211 replaces this PR, correct? We can close this one now that the TYDE version has been merged in? If yes, please close this PR.

narendrasss:
@wynnset yup, looks like it. I'll close it now.

PR #200, opened 2019-11-14
Add quiz screen

narendrasss:


anhduc130:
The gravity form should be embedded in the quiz screen.

narendrasss:
Thanks @anhduc130. That means this PR is now ready for QA.

templates/vue/src/components/lightbox/quiz/Question.vue
```diff
@@ -0,0 +1,174 @@
...
+            .querySelector(`form#gform_${id}`)
+            .addEventListener("submit", (event) => {
+              this.formOpened = false
+              this.$emit("form-submitted")
+              this.disableButton = false;
+            })
+
+          this.$refs["form-container"].appendChild(gravityForm)
+          this.formOpened = true
+          this.$emit("form-opened")
```
narendrasss: We should avoid doing any direct manipulation of the DOM. Vue has a `v-html` attribute that you can use to inject html, and events "bubble" so you can attach the submit handler to the `form-container` element and it should work fine.

anhduc130: Please feel free to modify the code. Thanks.

templates/vue/src/App.vue
```diff
@@ -22,37 +22,45 @@ export default {
 
 html {
   font-size: 100%;
+}
```
wynnset: It seems like this is reverting the SASS changes. Could you please make sure this is written is SCSS instead and not revert that change?

templates/vue/src/components/lightbox/quiz/Question.vue
wynnset: Yes, I agree with Narendra here. We should avoid doing things through plain JS when there is a vue way of doing it. Please fix this.

templates/vue/src/App.vue
narendrasss: Sorry about that, looks like a problem from doing the merge. Fixed.

PR #204, opened 2019-11-14
Add SASS, update webpack, and utilize it in all components

wynnset:


narendrasss:
Looks good. The next step is to convert the css code to sass so it's more compact.

anhduc130:
Looks good to me.

PR #209, opened 2019-11-19
Fixed node disconnecting after edit

narendrasss:


templates/vue/src/store/mutations.js
```diff
@@ -37,12 +37,9 @@ export function addNode(state, node) {
...
   const nodeIndex = Helpers.findNodeIndex(payload.id, state)
   const oldNode = state.nodes[nodeIndex]
-
-  const newNode = { ...oldNode }
   Object.entries(payload.newNode).forEach(([key, value]) => {
-    newNode[key] = value
+    oldNode[key] = value
   })
-  state.nodes[nodeIndex] = newNode
   state.nodes = [...state.nodes]
```
wynnset: is this required?

templates/vue/src/store/mutations.js
```diff
@@ -37,12 +37,9 @@ export function addNode(state, node) {
 export function updateNode(state, payload) {
   const nodeIndex = Helpers.findNodeIndex(payload.id, state)
   const oldNode = state.nodes[nodeIndex]
```
wynnset: can we rename this to just `thisNode` or something since "old" doesn't make sense anymore?

PR #211, opened 2019-11-19
Add End Screen and Quiz (generic)

narendrasss:


templates/vue/src/components/lightbox/quiz/QuizScreen.vue
```diff
@@ -0,0 +1,166 @@
...
+    </footer>
+  </div>
+</template>
+
+<script>
+import { mapActions } from "vuex"
+import Question from "./Question"
+import Loading from "../../Loading"
+import Helpers from "../../../utils/Helpers"
+import BackgroundImg from "../../../assets/11-18-QuestionScreen.png"
```
wynnset: Shouldn't this be removed?

narendrasss: Thanks, fixed.

PR #232, opened 2019-11-29
Stop commit when pre-commit hook fails to allow programmer to fix errors

### LINKED ISSUE #231: Remove fix flag from lint command

narendrasss:
Fixes #231

This PR changes the pre-commit hook to lint the files only and not attempt to fix the linting errors.

**Reasons for change**

* When the hook is run and the files get fixed, the files are modified _after_ the commit. This means I have to do `git commit --amend` every time a file gets fixed by the linter.


.githooks/pre-commit
```diff
@@ -1,5 +1,3 @@
 #!/bin/sh
 
-echo "We are now linting on every commit. You can skip this with the '--no-verify' flag."
-
-cd templates/vue && npm run lint -- --fix
\ No newline at end of file
+cd templates/vue && node lint.js
```
wynnset: Can we actually add a flag to allow user to skip linting when committing?

narendrasss: Users can do this already with the `--no-verify` flag. See [git docs](https://git-scm.com/docs/githooks#_pre_commit).

PR #233, opened 2019-12-01
Select H5P and Gravity Forms from a searchable dropdown instead of typing in ID

### LINKED ISSUE #219: List H5P content when selecting "H5P" content type

narendrasss:
Fixes #219

Changes the "h5p embed link" input into a combobox that lists all available h5p content in the server.


wynnset:
@narendrasss the only thing remaining on this task is to use the database table to retrieve the actual title of the H5P content instead of using its url.

narendrasss:
@wynnset I addressed the changes. From my research it seems the only way to interact with it is through pure SQL queries, which is what I've done. Please take a look.

classes/class.tapestry-h5p.php
```diff
@@ -0,0 +1,19 @@
...
+class TapestryH5P implements ITapestryH5P
+{
+  private $postId;
+
+  public function __construct($postId = null)
+  {
+    $this->postId = $postId;
+  }
+
+  public function get()
```
wynnset: All public PHP functions must have a comment block including description, parameters, and return

classes/class.tapestry-h5p.php
```diff
@@ -0,0 +1,19 @@
...
+  {
+    global $wpdb;
+    $content = $wpdb->get_results("select id, title from wp_h5p_contents");
```
wynnset: The table prefix `wp_` is not always the same on all sites, so please get the prefix dynamically using `$wpdb->prefix`.

PR #244, opened 2019-12-03
237 Mark answer types answered

### LINKED ISSUE #237: Mark answer types as answered when answered

anhduc130:
Fixes #237 
- This PR is technically done. However, it relies on some new changes coming from `tyde` to function properly, which is done via #256 .


templates/vue/src/components/lightbox/quiz/Question.vue
```diff
@@ -24,19 +24,22 @@
         <div class="button-container">
           <answer-button
             v-if="hasId('textId')"
+            :completed="true"
```
anhduc130: Mocking data here only. Will need to pass in the real value. So this PR is still WIP.

anhduc130:
This PR is now done and updated with the latest version of `master`. Just a headsup that if the `gravity form id` is shared among multiple nodes, the check mark will appear on all the nodes provided that the form is already answered by one of the shared nodes.

PR #246, opened 2019-12-05
Fix audio recorder

anhduc130:


anhduc130:
As per our conversation, the issue happened because the user was not logged in, leading to having no valid `entries` field. I added a check for that and it should be good now.

PR #248, opened 2019-12-13
Fix H5P end screen behaviour

narendrasss:
- Fix bugs recorded on Asana task
- Add loading screen that shows before the h5p iframe loads


templates/vue/src/components/lightbox/H5PMedia.vue
```diff
@@ -116,6 +122,7 @@ export default {
 
       if (this.node.mediaType === "video") {
         const h5pVideo = h5pObj.instances[0].video
+        console.log(h5pVideo)
```
wynnset: Please remove the console log or leave a comment as to why it needs to be here.

PR #251, opened 2019-12-14
Add TYDE node types

narendrasss:
**Changes**

- Add four node types as per specifications on Asana
- Update authoring modals to allow users to select from a predefined list of types
  - It also changes the possible options based on the parent of the current node

**Todo**

- ~~Add validation check on back-end to make sure users are not able to submit an invalid node~~
- Enforce new types on old tapestries
  - e.g. if an old tapestry has a module, modify the types of the children to be stages/question sets
- ~~Remove the add button for question sets so users are not able to add children~~

**Notes**

There's some edge cases that I think we should talk about.

- ~~What happens if a user changes a node type from module to regular? All of the children nodes will immediately become invalid because the parent would no longer be a module.~~ **Don't allow users to change the node type until module has no children**.
- ~~What happens if a user deletes a module or stage?~~ **Don't allow user to delete module or stages until their children are deleted as well**.


classes/class.tapestry.php
```diff
@@ -267,6 +290,47 @@ public function isEmpty()
         return empty($this->rootId);
     }
 
+    public function validate($node, $parent = null)
+    {
+        $type = $node->tydeType;
```
wynnset: It would be more clear if you name this variable `$tydeType`

classes/class.tapestry.php
```diff
@@ -267,6 +290,47 @@ public function isEmpty()
...
+
+        $parentType = $parent->tydeType;
+        if ($parentType == "Module") {
+            return $type == "Stage";
+        } else if ($parentType == "Stage") {
+            return $type == "Question set";
+        } else if ($parentType == "Regular") {
+            return $type == "Module" || $type == "Regular";
+        } else {
+            // otherwise parent is a question set, so we shouldn't be able
```
wynnset: What if `$parentType` is not set? It should pass, no? (For backwards compatibility)

classes/class.tapestry.php
```diff
@@ -267,6 +290,47 @@ public function isEmpty()
...
+            // to get here in the first place.
+            return false;
+        }
+    }
+
+    public function getParent($nodeId)
```
wynnset: Please rename this to `getNodeParent` for more clarity.

interfaces/interface.tapestry.php
```diff
@@ -45,12 +45,16 @@ public function getNodeIds();
 
     /**
      * Add a new node
+     * Note: (Narendra) I changed the signature of this function to
+     * allow the Tapestry class to validate the node before it gets
+     * saved to the DB.
```
wynnset: Please remove comment or form it into something that is not PR-specific.

templates/vue/src/utils/constants.js
```diff
@@ -0,0 +1,6 @@
+export const nodeTypes = {
```
wynnset: This should be called `tydeTypes`

classes/class.tapestry.php
```diff
@@ -131,8 +131,19 @@ public function getNodeIds()
...
      * @return  Object  $node   Tapestry node
      */
-    public function addNode($node)
+    public function addNode($node, $parentId = null)
     {
+        $parent = null;
+
+        if (isset($parentId)) {
+            $controller = $this->getNode($parentId);
+            $parent = $controller->get();
```
wynnset: Can you combine these two lines into one?

classes/class.tapestry.php
```diff
@@ -196,6 +207,18 @@ public function deleteNodeFromTapestry($nodeId)
      */
     public function addLink($link)
     {
+        $controller = $this->getNode($link->source);
+        $parent = $controller->get();
```
wynnset: Can you combine these two lines into one?

classes/class.tapestry.php
```diff
@@ -196,6 +207,18 @@ public function deleteNodeFromTapestry($nodeId)
...
+
+        $controller = $this->getNode($link->target);
+        $child = $controller->get();
```
wynnset: Can you combine these two lines into one?

templates/tapestry.js
```diff
@@ -204,7 +204,14 @@ function tapestryTool(config){
         }
 
         this.dataset.nodes = this.dataset.nodes.map(node => {
-            var updatedNode = fillEmptyFields(node, { skippable: true, behaviour: "embed", completed: false, quiz: [], showInBackpack: true })
+            var updatedNode = fillEmptyFields(node, {
+                tydeType: "Regular",
```
wynnset: Could we have it so when you don't pass the tydeType it sets it to Regular automatically? (Related to the other related comment)

templates/vue/src/components/NodeModal.vue
```diff
@@ -270,22 +276,31 @@
 
 <script>
 import Helpers from "../utils/Helpers"
-import Combobox from "./Combobox"
+import { tydeTypes } from "../utils/constants"
 import QuizModal from "./node-modal/QuizModal"
+import TydeTypeInput from "./node-modal/TydeTypeInput"
+import { mapGetters } from "vuex"
+import Combobox from "./Combobox"
```
wynnset: To minimize merge conflicts with master, please don't move lines if not necessary (unless you also move it on master)

templates/vue/src/components/NodeModal.vue
```diff
@@ -270,22 +276,31 @@
 
 <script>
 import Helpers from "../utils/Helpers"
-import Combobox from "./Combobox"
+import { tydeTypes } from "../utils/constants"
```
wynnset: To minimize merge conflicts with master, can we put all the tyde-related lines close to eachother if possible?

templates/vue/src/components/Tapestry.vue
```diff
@@ -215,7 +235,10 @@ export default {
         group: 1,
         typeData: {
           linkMetadata: null,
-          progress: [{ group: "viewed", value: 0 }, { group: "unviewed", value: 1 }],
+          progress: [
+            { group: "viewed", value: 0 },
+            { group: "unviewed", value: 1 },
+          ],
```
wynnset: If you're doing this here, make sure you also make the same change on master to minimize conflicts.

PR #252, opened 2019-12-16
Refactor h5p iframe on master

### LINKED ISSUE #242: Refactor H5PIframe into a component as per #217
Add the new H5PIframe component and simplify same as in tyde #217 (there may be other changes that also need to be brought over that apply to master)

anhduc130:
Fixes #242


anhduc130:
This branch should be ready for a review. Tested with the latest version of master.

PR #277, opened 2019-12-25
Add gravity form as a content type

### LINKED ISSUE #270: Add gravity form as a content type
Make sure to reuse the same code that we are using now.

narendrasss:
Fixes #270 

**Changes**

* Add gravity form as a content type in both the authoring tool and as a media type in the lightbox.
* ~~Refactor the `GravityForm` vue component so that it is more independent; it now only requires the form id and does everything else on its own (i.e. fetching html, get entries, submission, etc.).~~ **Moved to PR #279**.
* ~~Refactor the `Question` and `QuizScreen` components to conform to the new `GravityForm` API.~~ **Moved to PR #279**
* ~~Add a get gravity form entry endpoint to `TapestryForm`.~~ **Moved to PR #279**

**Notes**

* ~~In the lightbox, when a user submits a form the user is given no feedback if the submission is successful. This is intentional, as I'm planning to add the confirmation screen as part of a separate PR.~~ **Done as part of PR #280**


templates/vue/.vscode/settings.json
```diff
@@ -11,5 +11,8 @@
   ],
   "eslint.autoFixOnSave": true,
   "editor.formatOnSave": false,
-  "vetur.validation.template": false
+  "vetur.validation.template": false,
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": true
+  }
```
narendrasss: This was an automated change due to the old setting (`eslint.autoFixOnSave`) being deprecated.

PR #282, opened 2019-12-28
Add vue router

### LINKED ISSUE #281: Add vue router to allow linking of lightboxes

narendrasss:
**Changes**

* Add vue router
* Change lightbox functionality so that state is entirely dependent on current url
* Remove all obsolete lightbox code

The lightbox url is at `/nodes/:nodeId`, e.g. `/nodes/123` will open the lightbox for the node with ID 123.

Fixes #281


templates/vue/src/components/Lightbox.vue
```diff
@@ -1,18 +1,19 @@
 <template>
   <div
+    v-if="isTapestryLoaded"
```
wynnset: Can we change this to `tapestryLoaded` or `tapestryIsLoaded` instead since it will read nicer: "if tapestry is loaded"

narendrasss: Done, thanks.

narendrasss:
@wynnset Everything is resolved except for the part about navigating directly to a particular gravity form/audio recorder. You mentioned last time that I can leave it as is, but let me know if that changes and that feature is necessary.

PR #286, opened 2020-01-03
Add automated testing

narendrasss:
Add automated end-to-end testing using [cypress](https://www.cypress.io/).

- `author.spec.js` contains all tests related to the authoring tool as laid out in the testing checklist.
- `user.spec.js` contains all tests related to user use.

Some tests from the checklist are yet to be included here due to limitations in cypress itself:

- Dragging seems to be difficult to replicate given the way our nodes are implemented so that test simply asserts that all nodes are visible to the user.


templates/vue/cypress/support/roles.js
```diff
@@ -0,0 +1,10 @@
+export default {
+  admin: {
+    username: "admin",
+    password: "Tapestry123!",
+  },
+  subscriber: {
+    username: "narendras",
+    password: ")%AbNZgppS)uNvxT5)OdgzhX",
```
wynnset: If these require set up from the user's side, please add it to the README file. Please do not use your name, but instead a more generic username

templates/vue/src/components/NodeModal.vue
```diff
@@ -104,6 +110,7 @@
               <b-form-input
                 id="node-h5p-media-duration"
                 v-model="node.mediaDuration"
+                data-testid="node-h5pDuration"
```
wynnset: Can you not use the id attribute instead of the test-specific data attribute? Or the class attribute if it's repeatable? Would be much cleaner.

templates/vue/cypress/support/utils.js
```diff
@@ -0,0 +1,48 @@
+// TODO: Add node generator
+// TODO: Add link generator
+
+export const SITE_URL = "http://localhost:8888"
```
wynnset: This might be different on other machines - please make it customizable somehow - maybe through an environment file that you would add to .gitignore? And then include instructions to add it in the README file.

narendrasss:
@wynnset I changed all of the constants to cypress's environment file and I added instructions to the README. For the testId, I think it's good for us to keep it this way because we immediately know that the id is used only for testing purposes (and won't be changed in the future).

templates/vue/README.md
```diff
@@ -33,3 +33,37 @@ You need to adjust your git config so the automatic linting process works correc
...
+## Testing Setup
+
+The testing framework that we use, Cypress, needs to be configured with some environment variables. In the `cypress.json` file, you should see something like:
+
+```json
+{
+    "env": {
+        "DEV_URL": "http://localhost:8888",
+        "TEST_TAPESTRY_NAME": "testing",
+        "USER_TAPESTRY_NAME": "testing-user",
```
wynnset: Can we remove these? Just name them whatever - the user doesn't care about the name of the tapestries used in testing (right?)

templates/vue/README.md
```diff
@@ -33,3 +33,37 @@ You need to adjust your git config so the automatic linting process works correc
...
+        "ADMIN_USERNAME": "admin",
+        "ADMIN_PASSWORD": "Tapestry123!",
+        "SUBSCRIBER_USERNAME": "narendras",
+        "SUBSCRIBER_PASSWORD": ")%AbNZgppS)uNvxT5)OdgzhX"
```
wynnset: Can you use admin_user / admin_pass and subscriber_user / subscriber_pass instead of the values you have set here as examples, please?

templates/vue/README.md
```diff
@@ -33,3 +33,37 @@ You need to adjust your git config so the automatic linting process works correc
...
+    }
+}
+```
+
+Change the variables to match your own settings.
+
+* `DEV_URL` is the url of your local site.
+* `TEST_TAPESTRY_NAME` is the name of a new tapestry that is automatically made (and deleted) during the authoring tests.
+* `USER_TAPESTRY_NAME` is the name of the tapestry that is used for the user-side tests.
```
wynnset: As per my previous comment, let's remove these?

templates/vue/cypress.json
```diff
@@ -0,0 +1,11 @@
...
+    "env": {
+        "DEV_URL": "http://localhost:8888",
+        "TEST_TAPESTRY_NAME": "testing",
+        "USER_TAPESTRY_NAME": "testing-user",
+        "ADMIN_USERNAME": "admin",
+        "ADMIN_PASSWORD": "Tapestry123!",
+        "SUBSCRIBER_USERNAME": "narendras",
+        "SUBSCRIBER_PASSWORD": ")%AbNZgppS)uNvxT5)OdgzhX"
+    }
+}
```
wynnset: I think it's better we remove this file and add it to gitignore and just update the README to let the user know they need to add this file in this format. Otherwise, users might accidentally push their own settings to the repo, which we don't want.

templates/vue/README.md
narendrasss: By the user do you mean the developers? The `TEST_TAPESTRY_NAME` doesn't matter, but the `USER_TAPESTRY_NAME` matters because it should refer to a test tapestry that was (previously) manually created.

templates/vue/README.md
narendrasss: Please see my other comment.

templates/vue/cypress.json
narendrasss: Sure, I'll add a `cypress-sample.json` and update the README

templates/vue/src/components/NodeModal.vue
wynnset: Hey @narendrasss still waiting for a response here.

wynnset: Oops, just saw your response - ignore.

PR #292, opened 2020-01-14
Changed Font Awesome 'play' to "Aa" string

liam-armstrong:
Removed function call to helper that adds Font Awesome play icon
Added String "Aa" between tags to render that text as an icon

### DIFF
--- a/templates/tapestry.js
+++ b/templates/tapestry.js
@@ -1218,7 +1218,7 @@ function tapestryTool(config){
             .append("svg:foreignObject")
             .html(function (d) {
                 return '<i id="mediaButtonIcon' + d.id + '"' + 
-                    ' class="' + getIconClass(d.mediaType, 'play') + ' mediaButtonIcon"' +
+                    ' class="' + ' mediaButtonIcon"' +
                     ' data-id="' + d.id + '"' + 
                     ' data-format="' + d.mediaFormat + '"' + 
                     ' data-media-type="' + d.mediaType + '"' + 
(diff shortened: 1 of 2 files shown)


templates/tapestry.js
```diff
@@ -1218,15 +1218,15 @@ function tapestryTool(config){
             .append("svg:foreignObject")
             .html(function (d) {
                 return '<i id="mediaButtonIcon' + d.id + '"' + 
-                    ' class="' + getIconClass(d.mediaType, 'play') + ' mediaButtonIcon"' +
+                    ' class="' + ' mediaButtonIcon"' +
```
wynnset: This doesn't work. We only want the icon to change for text nodes (i.e. nodes that have their node type set as "Text"). You'll need to keep this function and instead change the implementation of the function and maybe add some CSS.

templates/tapestry.js
```diff
@@ -1218,15 +1218,15 @@ function tapestryTool(config){
...
                     ' data-id="' + d.id + '"' + 
                     ' data-format="' + d.mediaFormat + '"' + 
                     ' data-media-type="' + d.mediaType + '"' + 
                     ' data-thumb="' + d.imageURL + '"' +
                     ' data-fullscreen="' + d.fullscreen + '"' +
                     ' data-url="' + (d.typeData.mediaURL ? d.typeData.mediaURL : '') + '"' +
                     ' data-media-width="' + d.typeData.mediaWidth + '"' + 
-                    ' data-media-height="' + d.typeData.mediaHeight + '"><\/i>';
+                    ' data-media-height="' + d.typeData.mediaHeight + '">Aa<\/i>';
```
wynnset: I recommend using CSS pseudo-code `:after` to add the `Aa` text in using CSS so that you don't have to add anything in here and it's consistent with how FontAwesome works.

PR #294, opened 2020-01-14
Turn off eslint line break rule

narendrasss:


.gitignore
```diff
@@ -0,0 +1,3 @@
+package-lock.json
+.DS_Store
+./templates/.DS_Store
```
wynnset: Can you change this so all .DS_Store files are excluded, regardless of the directory they are in?

PR #295, opened 2020-01-14
Disabled Autocomplete for Gravity Forms

liam-armstrong:
Added a function to GravityForm.vue that iterates through all form inputs on page, filters to select only the GravityForm fields and manually sets the autocomplete tag to "off" for each


wynnset:
@narendrasss please do the code review for this and "Squash and Merge" when it's ready.

PR #297, opened 2020-01-15
Explicitly show scrollbar

ItsAdel:
Fix will force show a custom scrollbar for windows and mac

Fixing the issue to explicitly show for mac:
http://simurai.com/blog/2011/07/26/webkit-scrollbar

image

### DIFF
--- a/templates/vue/src/components/Lightbox.vue
+++ b/templates/vue/src/components/Lightbox.vue
@@ -202,12 +202,12 @@ export default {
     box-shadow: 0 0 100px -40px #000;
     border-radius: 15px;
     scrollbar-color: rgba(0,0,0,.5) rgba(255,255,255,.5) ;
-    scrollbar-width: 7px;
+    scrollbar-width: auto;
     -ms-overflow-style:scrollbar;
 
     ::-webkit-scrollbar {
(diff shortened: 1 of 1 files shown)


ItsAdel:
From conducting some more research online, I realized that we can only increase the width for the scrollbar for webkit based browsers. The -ms-overflow-style property should work for MSEdge but tapestry won't load on my MSEdge browser, so I can't confirm. More info: https://developer.mozilla.org/en-US/docs/Web/CSS/-ms-overflow-style.  For firefox, the scrollbar-width property is used to adjust the width but it only has a few options to choose from: https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-width

wynnset:
@ItsAdel we still need to put this feature in place, so have to find a way. This seems to work for me: https://grsmto.github.io/simplebar/  - Not sure if there's other solutions, but I recommend you try to find something that works

narendrasss:
Should we close this @wynnset? It's on hold, but it's nevertheless made obsolete by #347

wynnset:
Obsolete PR and doesn't seem like we can find a cross-browser solution to this.

PR #299, opened 2020-01-15
Disabled scroll when lightbox is open

ItsAdel:


templates/vue/src/components/Lightbox.vue
```diff
@@ -141,9 +141,13 @@ export default {
     this.applyDimensions()
     thisTapestryTool.selectNode(Number(this.nodeId))
     thisTapestryTool.changeToViewMode(this.lightboxDimensions)
+    let elem = document.querySelector('body')
+    elem.style.overflow = "hidden"
   },
   beforeDestroy() {
     thisTapestryTool.exitViewMode()
+    let elem = document.querySelector('body')
+    elem.style.overflow = "auto"
```
wynnset: We generally try not to declare variables if they are only used once. Please change these to `document.querySelector('body').style.overflow = "hidden"` and `document.querySelector('body').style.overflow = "auto"` respectively.

PR #305, opened 2020-01-17
Add accordion node type

### LINKED ISSUE #291: Add accordion content type

narendrasss:
Fixes #291

**Tasks**

- ~~Refactor lightbox to allow reuse of media components~~
- ~~Add accordion type to authoring tool and database~~
- ~~Read rows from children of accordion node~~
- ~~Hide accordion rows from users who cannot edit the tapestry~~
- ~~Add confirmation screen when a user is finished with a particular row~~
- ~~Add "back to top" button~~
- ~~Add accordion icon to node~~
- ~~Add locking feature so locked rows cannot be opened~~
- ~~Update progress bar so progress represents number of rows completed~~
- ~~Fix up styling~~

### DIFF
--- a/templates/tapestry.js
+++ b/templates/tapestry.js
@@ -212,13 +212,19 @@ function tapestryTool(config){
             updatedNode.permissionsOrder = reorderPermissions(updatedNode.permissions);
 
             if (node.mediaType === "accordion") {
+                const accordionProgress = []
                 const directChildren = this.dataset.links.filter(link => {
                     return link.source == node.id
                 }).map(link => link.target)
                 directChildren.forEach(childId => {
                     const child = this.dataset.nodes[findNodeIndex(childId)]
                     child.presentationStyle = "accordion-row"
+                    
+                    if (child.completed) {
+                        accordionProgress.push(childId)
+                    }
                 })
+                updatedNode.accordionProgress = accordionProgress
             }
 
             return updatedNode
--- a/templates/tapestry.js
+++ b/templates/tapestry.js
@@ -212,12 +212,12 @@ function tapestryTool(config){
             updatedNode.permissionsOrder = reorderPermissions(updatedNode.permissions);
 
             if (node.mediaType === "accordion") {
-                const directChildren = this.dataset.links.filter(link => {
-                    return link.source == node.id
-                }).map(link => link.target)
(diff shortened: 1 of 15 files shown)


templates/tapestry.js
```diff
@@ -210,6 +210,23 @@ function tapestryTool(config){
...
+                directChildren.forEach(childId => {
+                    const child = this.dataset.nodes[findNodeIndex(childId)]
+                    child.presentationStyle = "accordion-row"
+                    
+                    if (child.completed) {
+                        accordionProgress.push(childId)
+                    }
+                })
+                updatedNode.accordionProgress = accordionProgress
+            }
```
wynnset: Progress update is done further up in the code using a different function that handles everything. Can you please put it there instead to keep the code clean and readable?

templates/vue/src/components/TapestryModal.vue
```diff
@@ -0,0 +1,131 @@
+<template>
```
wynnset: Should this file not go into the lightbox folder instead? Not sure - just checking.

templates/tapestry.js
narendrasss: Fixed, thanks

templates/vue/src/components/TapestryModal.vue
narendrasss: I left it here because it might be used outside of a lightbox component. I'm not sure, but I think we can leave it here for the time being.

wynnset:
A few changes required after user feedback:
- Accordions first video sometimes starts not at the beginning when you open the window
- Do not automatically play video when accordion rows open - instead add a big play button on video thumbnail and let them play
- Change full-screen functionality to be available to all nodes

narendrasss:
@wynnset I added an autoplay prop and set it to false for accordion nodes so the videos don't play automatically. That also made the first problem obsolete. For the full-screen functionality, I'll open a separate PR.

wynnset:
Hey @narendrasss this is looking really good! Sorry I missed one other thing they asked for: When the user presses the "Finished" button at the bottom of each accordion row, we currently show them the pop-up. We want to change this so that the pop-up only shows up when pressing the Finished button on the last accordion row, not for every row.

narendrasss:
@wynnset Fixed.

PR #314, opened 2020-01-30
Change module start event to look for module type

narendrasss:


wynnset:
This is already merged into tyde along with the latest master merge

Issue #306, opened 2020-01-18
Improve GF Image Choices UI


liam-armstrong:
Currently, it's not fully clear what the difference is between a choice that has the focus (whether it's selected or not) and a choice that is selected. 

Let's change it so that the selected items maybe have a checkmark in the corner. Focused items must not have the black border, but instead a much more subtle difference with a choice in its regular state.

Their website has some good UIs we can try to use: https://jetsloth.com/gravity-forms-image-choices - just need to add the relevant CSS. 

Also let's left-align the checkboxes (see 3rd screenshot from M1S1 - "In School & Community")

The CSS can be added to the GravityForm component probably (not scoped)

PR #315, opened 2020-01-31
#306 Gravity Forms Image UI Styling

liam-armstrong:
Added:
- gf-image-ui.css
Stylesheet for forms targeting whole Gravity Forms html box, placing it in GravityForm.vue was not high enough on the scope
- InjectImageStyling() method to GravityForms.vue:
Iterates through the relevant HTML elements and adds event listeners for triggering class changes, activating CSS animation
- Load GravityForms JS and GravityForm Image Choice JS from singleTapestry.php
Loading the plugins CSS and JS keeps our in modal quizzes consistent with Gravity Forms posts

### DIFF
--- a/templates/libs/gf-image-ui.css
+++ b/templates/libs/gf-image-ui.css
@@ -1,0 +1,155 @@
+.gform_wrapper {
+    width: 50%;
+}
+
+.gfield_checkbox {
+    display: flex;
+    justify-content: space-evenly;
+    align-items: center;
+}
+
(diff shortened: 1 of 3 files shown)


wynnset:
As per our discussion, I'm going to assign this back to you to look into the alternative way we discussed. I think it might be worthwhile for you to create a separate branch / PR for that alternative solution and keep this one intact in case that one doesn't end up working for any reason @liam-armstrong / @narendrasss

PR #320, opened 2020-02-01
Simplify file upload

ItsAdel:
Created a new reusable file upload component for easy drag and drop or upload button for external links and thumbnails. Users can still manually enter links. Added file upload progress bar at the bottom for better UI indicator

External Link:
image

Thumbnail:
image


templates/vue/src/components/FileUpload.vue
```diff
@@ -0,0 +1,135 @@
...
+              'Content-Type': 'multipart/form-data'
+            },
+            onUploadProgress: function( progressEvent ) {
+              this.uploadPercentage = parseInt( Math.round( ( progressEvent.loaded / progressEvent.total ) * 100 ))
+            }.bind(this)
+          }
+        )
+        .then(response => {
+          this.$emit('input', response.data.data.url)
+        })
```
narendrasss: Let's keep the code styling consistent. Make sure the linter runs before you commit. If it doesn't, let me know and we can figure out why it's not running automatically.

templates/vue/src/components/FileUpload.vue
```diff
@@ -0,0 +1,135 @@
...
+
+  data() {
+    return{
+      uploadPercentage: 0
+    }
+  },
+
+  methods: {
+    uploadFile(event) {
+      var formData = new FormData()
```
narendrasss: Try to use `const` over `var` whenever possible. `var` is not block scoped so other places may counter-intuitively have access to it.

templates/vue/src/components/NodeModal.vue
```diff
@@ -144,17 +144,16 @@
...
+              v-if="node.mediaType === 'url-embed'"
               label="External Link"
             >
-              <b-form-input
-                id="node-embed-media-duration"
+              <file-upload
                 v-model="node.typeData.mediaURL"
-                placeholder="Enter embed link (starting with http)"
-                required
+                input-id="node-embed-media-duration"
```
wynnset: We can probably remove this and the `inputId` prop altogether from the component.

templates/vue/src/components/NodeModal.vue
```diff
@@ -144,17 +144,16 @@
...
+                placeholder-text="Enter embed link (starting with http)"
```
wynnset: Is it not possible to just call this prop `placeholder`?

templates/vue/src/components/NodeModal.vue
```diff
@@ -173,11 +172,10 @@
               </b-form-checkbox>
             </b-form-group>
             <b-form-group v-if="addThumbnail">
-              <b-form-input
-                id="node-image-url"
+              <file-upload
                 v-model="node.imageURL"
-                placeholder="Enter the URL for the thumbnail"
-                required
+                input-id="node-image-url"
```
wynnset: Again, I don't think this prop is necessary and can be removed.

templates/vue/src/components/NodeModal.vue
```diff
@@ -173,11 +172,10 @@
...
+                placeholder-text="Enter the URL for the thumbnail"
```
wynnset: Ditto as per comment on previous instance of this tag.

templates/vue/src/components/FileUpload.vue
```diff
@@ -0,0 +1,138 @@
...
+<script>
+import axios from "axios"
+
+export default {
+  name: "file-upload",
+  props: {
+    inputId: {
+      type: String,
+      required: true,
+    },
```
wynnset: We can probably get rid of this prop.

templates/vue/src/components/FileUpload.vue
```diff
@@ -0,0 +1,138 @@
...
+    placeholderText: {
```
wynnset: Can we change this to just `placeholder`?

templates/vue/src/components/FileUpload.vue
```diff
@@ -0,0 +1,138 @@
...
+  display: inline-block;
+  overflow: hidden;
+  vertical-align: middle;
+  width: 100%;
+  position: absolute;
+  left: 0;
+  padding-left: 30px;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+}
```
wynnset: Is this necessary? I tried removing all of this and it worked just fine. Let's try and keep the code minimal.

templates/vue/src/components/FileUpload.vue
```diff
@@ -0,0 +1,138 @@
...
+  padding: 10px;
+}
+
+progress[value] {
+  /* Reset the default appearance */
+  -webkit-appearance: none;
+  appearance: none;
+
+  width: 250px;
+  height: 20px;
```
wynnset: Can you reduce the height to 5px and move the progress bar closer to the file input field please?

PR #337, opened 2020-02-08
Don't allow scroll when stage is open

ItsAdel:


templates/vue/src/components/tyde/TydeModule.vue
```diff
@@ -54,9 +54,11 @@ export default {
     ...mapMutations(["openLightbox", "closeLightbox"]),
     openLightbox(id) {
       this.$router.push(`/nodes/${id}`)
+      document.querySelector("body").style.overflow = "hidden"
     },
     closeLightbox() {
       this.$router.push(`/`)
+      document.querySelector("body").style.overflow = "auto"
```
wynnset: This is not working because it's attached to the lightbox which is different than a stage.

templates/vue/src/components/Lightbox.vue
```diff
@@ -201,6 +201,12 @@ export default {
 }
 </script>
 
+<style>
+.lightbox-open {
```
wynnset: To avoid conflicts, please add `tapestry` at the beginning of your class name. i.e. `tapestry-lightbox-open` instead of `lightbox-open`. Even better, put `body.tapestry-lightbox-open` instead so it only applies to the body.

templates/vue/src/components/tyde/TydeStage.vue
```diff
@@ -74,6 +80,10 @@ export default {
   --tyde-green: #205a27;
   --tyde-border-green: #39b54a;
 }
+
+.stage-open {
```
wynnset: To avoid conflicts, please add `tapestry` at the beginning of your class name. i.e. `tapestry-stage-open` instead of `stage-open`. Even better, put `body.tapestry-stage-open` instead so it only applies to the body.

PR #347, opened 2020-02-15
Add custom and enlarged scrollbar to lightboxes

### LINKED ISSUE #296: Make the scrollbar larger and always visible on all browsers

narendrasss:
Resolves #296 

TODO:

- [X] Minor styling fixes
- [X] Allow scrollbar to be dragged
- [x] Add arrow up/arrow down listeners so scroll mimic real scroll
- [x] Hide native scrollbar on **all** browsers (currently only hidden in Firefox)


templates/vue/src/components/lightbox/WpPostMedia.vue
```diff
@@ -26,6 +29,8 @@ export default {
       loading: true,
       title: "",
       content: "",
+      scrollHeight: 0,
+      clientHeight: 0,
```
narendrasss: This has to be added to the `data` field of the parent component.

templates/vue/src/components/lightbox/WpPostMedia.vue
```diff
@@ -39,6 +44,13 @@ export default {
     this.title = post.title
     this.content = post.content
     this.$emit("complete")
+    this.$emit("load")
+
+    this.$nextTick(() => {
+      const post = this.$refs.post
+      this.scrollHeight = post.scrollHeight
+      this.clientHeight = post.clientHeight
```
narendrasss: This has to be done anytime the lightbox's height changes. In places like the accordion that can change often, this gets slightly complex.

templates/vue/src/components/lightbox/H5PIframe.vue
```diff
@@ -203,7 +201,7 @@ export default {
         })
       }
       $(window).resize(setIframeDimensions)
-      setIframeDimensions()
+      setIframeDimensions() */
```
narendrasss: I'm not super sure what this block of code does. I commented it out because it was affecting the positioning of the scrollbar, and there didn't seem to be any differences.

wynnset: This was added in to automatically size the iframe for the H5P to fit into the lightbox. Are you sure it works without it?

templates/vue/src/components/lightbox/WpPostMedia.vue
wynnset: Can you expand? Is there an action you'd like me to take on this?

templates/vue/src/components/lightbox/WpPostMedia.vue
wynnset: Ok, but this is in `wpPostMedia` - should we add this to Accordion Media as well? And fire it off everytime the container is resized? I know in jQuery you have a `resize` event on elements - not sure what's the Vue equivalent - or we can just use vanilla JS

narendrasss:
> Tested this and I had several issues:
> 
>     * H5P media is not loading
> 
>     * No scrollbar displays on WP Post (didn't try other node types)
> 
> 
> Note that these issues I mentioned do not exist on master.

@wynnset Can confirm that H5P media nodes are loading properly on my end. What kind of H5P are you attempting to load?

templates/vue/src/components/lightbox/WpPostMedia.vue
narendrasss: Sorry if I wasn't clear. I meant that this same code is copied and reused in all media types that require a scrollbar. I highlighted this to show the complexity that implementing this custom scrollbar introduces - if (in your opinion) this added code isn't a problem, then I wouldn't worry about it.

templates/vue/src/components/lightbox/WpPostMedia.vue
narendrasss: No - please see comment below.

templates/vue/src/components/lightbox/H5PIframe.vue
narendrasss: Did some testing and can confirm that it works properly.

wynnset:
The scrollbar is not draggable for me. Also, we should be hiding the native scrollbar if we are displaying our own scrollbar (it currently shows both).

wynnset:
Obsolete PR and doesn't seem like we can find a cross-browser solution to this.

PR #348, opened 2020-02-17
Add unlock conditions

### LINKED ISSUE #343: Add condition field to node and authoring tool

narendrasss:
Resolves #343 

* Add `conditions` array to nodes that determines the necessary conditions for the node to be unlocked. Each element is an object of the following shape:

```
{
  type: condition_type (only node completion at the moment)
  value: node_id
  fulfilled: bool
}
```

* Add new "Unlock Conditions" tab to authoring modal
* Remove `unlocked` field from nodes (in database)
  * `unlocked` field is now computed on initial load based on the node's conditions

The only thing left to do now is to ensure the appropriate conditions are getting fulfilled when a node gets completed. This is done by:

- [x] Computing a list of `dependents` for each node
- [x] Updating said list of dependents when the node gets completed

I believe @ItsAdel wanted to work on this part, so I'll leave it to him for now.


templates/tapestry.js
```diff
@@ -2021,20 +2017,30 @@ function tapestryTool(config){
...
+        }
+
+        const { nodes } = tapestry.dataset
+        nodes.forEach(node => {
+            const { conditions } = node
+            conditions.forEach(condition => {
+                const dependency = nodes[findNodeIndex(condition.value)]
+                let nodeDependents = dependency.dependents
+                nodeDependents.push({id: node.id, condition: condition})
+                dependency.dependents = nodeDependents
```
narendrasss: @ItsAdel I think I know what's causing the duplicate conditions. `setUnlocked` is called every time the tapestry rerenders, and every time the nodeDependents list gets updated. Maybe add an if statement before to ensure the condition doesn't exist already? Or change line to:

```suggestion
                let nodeDependents = []
                nodeDependents.push({id: node.id, condition: condition})
                dependency.dependents = nodeDependents
```
Although this would cause a new array to be created every time, which might be a bit inefficient.

narendrasss: Nevermind disregard this - I found a workaround that doesn't involve reinitializing the entire tapestry, so this doesn't happen anymore.

templates/vue/src/components/node-modal/ConditionsForm.vue
```diff
@@ -0,0 +1,108 @@
...
+
+.condition-close-button {
+  background: none;
+  margin: 0;
+  padding: 0;
+  position: absolute;
+  top: 8px;
+  right: 12px;
+  color: rgba(0, 0, 0, 0.5);
+}
```
wynnset: This should be nested under `.condition-container` to be more aligned with the actual HTML.

templates/tapestry.js
```diff
@@ -1071,11 +1026,104 @@ function tapestryTool(config){
...
+            const listItem = document.createElement("li");
+            listItem.innerText = "Complete this parent.";
+            wrapper.appendChild(listItem);
+        } else {
+            node.conditions.forEach(cond => {
+                if (!cond.fulfilled) {
+                    const listItem = document.createElement("li");
+                    switch (cond.type) {
+                        case conditionTypes.NODE_COMPLETED: {
+                            const node = getNodeById(cond.value);
```
wynnset: Why is the node ID called "Value"? Should it not be called "Node ID" so it's more clear what it is?

narendrasss: @ItsAdel can you look at the above bug? I think this was the thing that we discovered a while back. I thought it was fixed.

PR #353, opened 2020-02-21
Add styling changes for locked nodes

narendrasss:
Screen Shot 2020-02-21 at 13 40 52

- [X] Update locked nodes to appear greyed out and show a locked icon
- [x] Add tooltip to indicate to users what they need to do unlock a certain node
- [x] Fix tooltip positioning


wynnset:
Thanks for the great work on here. I just pushed a small reorganization of the authoring tool to include the Lock Conditions under the same tab as "Permissions" (now renamed to "Access"). The authoring part now looks good.

wynnset:
I'm seeing some minor issues with the placement of the tooltip. The first time I loaded it (or if I clear my cache and refresh), the tooltip shows in a weird place: 
image

After that point, the tooltip shows a bit higher than it should:
image

wynnset:
Also found another issue where if I make my browser smaller to shrink the tapestry, it changes the vertical positioning of the tooltip.

wynnset:
It might be easier and better to create the tooltip as an SVG element using the d3 structure so it can be positioned automatically and then show and hide it using the same way you are doing it.

wynnset:
As the author, when I select the node, the tooltip is covered by the media button: 
image

PR #361, opened 2020-03-05
Added authoring tools for adding spaceship parts

ItsAdel:
Getting linter errors on this. #360


templates/vue/src/components/NodeModal.vue
```diff
@@ -528,6 +629,14 @@ export default {
         { name: "fullscreen", value: this.node.fullscreen },
         { name: "tydeType", value: this.node.tydeType },
         { name: "showInBackpack", value: this.node.showInBackpack },
+        { name: "planetViewNotEarnedIconUrl", value: this.node.typeData.planetViewNotEarnedIconUrl },
+        { name: "planetViewEarnedIconUrl", value: this.node.typeData.planetViewEarnedIconUrl },
+        { name: "spaceshipPartNotEarnedIconUrl", value: this.node.typeData.spaceshipPartNotEarnedIconUrl },
+        { name: "spaceShipPartEarnedIconUrl", value: this.node.typeData.spaceShipPartEarnedIconUrl },
```
wynnset: How come `spaceShipPartHoverUrl` is everywhere else, but not here?

templates/vue/src/components/NodeModal.vue
```diff
@@ -334,6 +334,106 @@
         >
           <quiz-modal :node="node" />
         </b-tab>
+        <b-tab
+          v-if="node.tydeType === tydeTypes.MODULE"
+          title="Spaceship Icons"
```
wynnset: This should read "Spaceship Part"

templates/vue/src/components/NodeModal.vue
```diff
@@ -334,6 +334,106 @@
...
+              label-cols-lg="2"
+              label="Height: "
+              label-for="node-spaceship-part-height"
+            >
+              <b-form-input
+                id="node-spaceship-part-height"
+                v-model="node.typeData.spaceshipPartHeight"
+                placeholder="In pixels"
+              />
+            </b-form-group>
```
wynnset: It would be nice if we organize this the same way as the coordinates so they are both in the same row (called "Dimensions" or something)

templates/vue/src/components/NodeModal.vue
```diff
@@ -437,6 +537,7 @@ export default {
       formErrors: "",
       maxDescriptionLength: 250,
       addThumbnail: false,
+      tydeTypes: tydeTypes,
```
wynnset: This was not previously needed. Are you sure you need to add this?

templates/vue/src/components/NodeModal.vue
```diff
@@ -805,5 +914,9 @@ table {
       margin-bottom: 0;
     }
   }
+
+  #node-shapeship-coordinate-text {
+    padding-top: 6px;
+  }
```
wynnset: Please use bootstrap classes for minor spacing adjustments instead of your own CSS. See https://getbootstrap.com/docs/4.0/utilities/spacing/ (e.g. `pt-2` or something like that)

templates/vue/src/components/NodeModal.vue
```diff
@@ -334,6 +334,106 @@
...
+                placeholder="Enter link (starting with http)"
+              />
+            </b-form-group>
+            <b-form-group label="Planet View Icon - earned">
+              <file-upload
+                v-model="node.typeData.planetViewEarnedIconUrl"
+                placeholder="Enter link (starting with http)"
+              />
+            </b-form-group>
+            <b-form-group label="Spaceship Part - not earned">
```
wynnset: Please add headings to group items belonging to "Planet View Icon" and "Spaceship Cockpit Image" separately. See the Appearance tab to see how the headings are implemented and replicate that here.

templates/vue/src/components/NodeModal.vue
wynnset: (also you've misspelled spaceship here it says "shapeship")

templates/vue/src/components/NodeModal.vue
ItsAdel: We need this because I use it in the template for checking tydeTypes. Previous it was only used in the script tag and this is imported from a different file.

templates/vue/src/components/NodeModal.vue
```diff
@@ -528,6 +640,15 @@ export default {
         { name: "fullscreen", value: this.node.fullscreen },
         { name: "tydeType", value: this.node.tydeType },
         { name: "showInBackpack", value: this.node.showInBackpack },
+        { name: "planetViewNotEarnedIconUrl", value: this.node.typeData.planetViewNotEarnedIconUrl },
+        { name: "planetViewEarnedIconUrl", value: this.node.typeData.planetViewEarnedIconUrl },
+        { name: "spaceshipPartNotEarnedIconUrl", value: this.node.typeData.spaceshipPartNotEarnedIconUrl },
+        { name: "spaceShipPartEarnedIconUrl", value: this.node.typeData.spaceShipPartEarnedIconUrl },
+        { name: "spaceShipPartHoverUrl", value: this.node.typeData.spaceShipPartHoverUrl },
```
wynnset: All the other images finish with `IconUrl` but this one ends with just `Url`. Can you please keep this consistent by making it end with `IconUrl` as well?

templates/vue/src/components/Tapestry.vue
```diff
@@ -348,6 +357,33 @@ export default {
...
+          case "spaceShipPartEarnedIconUrl":
+            newNodeEntry.typeData.spaceShipPartEarnedIconUrl = fieldValue
+            break
+          case "spaceShipPartHoverUrl":
+            newNodeEntry.typeData.spaceShipPartHoverUrl = fieldValue
+            break
+          case "spaceshipPartX":
+            newNodeEntry.typeData.spaceshipPartX = fieldValue
+            break
+          case "spaceshipPartX":
```
wynnset: This should be `"spaceshipPartY"`, not X.

PR #367, opened 2020-03-13
Create spaceship cockpit

ItsAdel:
Did not run linter yet.
Implementation for rendering all the spaceships parts done.


templates/vue/src/components/tyde/TydeCockpit.vue
```diff
@@ -1,42 +1,29 @@
...
-        v-for="log in visibleLogs"
-        :key="log.name"
-        :log="log"
-        @close="$emit('close')"
-      />
-    </ul>
+  <section id="tyde-cockpit" :style="cockpitImage">
+    <div v-for="node in this.nodes" :key="node.id">
+      <div
+        v-if="isNodeModuleType(node)"
```
narendrasss: `v-for` with `v-if` sometimes doesn't work as expected. I would suggest adding a computed property that filters this out, e.g.

```js
modules() {
  return this.nodes.filter(isNodeModuleType)
}
```

templates/vue/src/components/tyde/TydeCockpit.vue
```diff
@@ -65,6 +53,12 @@ export default {
       }
       return this.logsWithAnswers
     },
+    cockpitImage() {
+      return {
+        backgroundImage: `url(${this.settings.spaceshipBackgroundUrl})`
+      }
```
narendrasss: I'm not able to change this setting anywhere in the UI. Not sure if that's intended. Can you clarify @wynnset ?

templates/vue/src/components/tyde/TydeSpaceship.vue
```diff
@@ -55,9 +56,9 @@ export default {
   },
   computed: {
     ...mapState(["lightbox"]),
-    backpackUrl() {
-      return `${wpData.vue_uri}/${BackpackIcon.split("dist")[1]}`
-    },
+    spaceshipIconUrl() {
+      return `${wpData.vue_uri}/${SpaceshipIcon.split("dist")[1]}`
+    }
```
narendrasss: Please use the `getImagePath` function available in the `Helpers` class instead.

templates/vue/src/components/tyde/TydeCockpit.vue
wynnset: You can change the setting by going to the Tapestry settings (the gear icon where you also can change the background for the Tapestry or turn auto-layout on/off) and then going to the TYDE tab.

templates/vue/src/components/tyde/TydeCockpit.vue
```diff
@@ -126,4 +147,5 @@ export default {
     overflow-y: scroll;
   }
 }
+
 </style>
```
wynnset: Please keep the original TydeMenuHome component. We may want to use it still in the future.

templates/vue/src/components/tyde/TydeCockpit.vue
```diff
@@ -105,12 +99,39 @@ export default {
...
 }
 </script>
 
 <style lang="scss" scoped>
-#tyde-menu-home {
+#tyde-cockpit {
+  height: 85vh;
   nav ul {
     display: flex;
     font-size: 16px;
```
wynnset: `nav ul` and `.logs` is not used in the TydeCockpit component. Please remove.

templates/vue/src/components/tyde/TydeCockpit.vue
```diff
@@ -105,12 +100,45 @@ export default {
...
+    },
+    isNodeModuleType(node) {
+      return node.tydeType === tydeTypes.MODULE
+    },
+    nodeMouseOverHandler(node) {
+      document.getElementById("tyde-cockpit-"+node.id).style.backgroundImage = `url(${node.typeData.spaceShipPartHoverIconUrl})`
+    },
+    nodeMouseLeaveHandler(node) {
+      document.getElementById("tyde-cockpit-"+node.id).style.backgroundImage = `url(${this.moduleImage(node)})`
+    },
```
wynnset: This is not the right way of doing things in Vue. I recommend creating a component for a spaceship part. Inside that component, keep a state to determine which image should be used. Update this state when the user hovers over or away, and tie the background image to the state. Let me know if you want to chat about this to either myself or Narendra.

Issue #362, opened 2020-03-07
Fix the progress for module, stage, and accordion


liam-armstrong:
The progress for these node types equals the percentage of their children that are marked as completed.

This should get updated when children are marked as completed or when a child is added or deleted.

PR #369, opened 2020-03-14
362 fix progress

liam-armstrong:
Closes issue #362 
Added:
- tydeProgress attribute to node (real number, 0 to 1)
- Mutation to update tydeProgress on stages and modules
- Calls to tydeProgress on tapestry change and node progress update

### DIFF
--- a/templates/vue/src/store/mutations.js
+++ b/templates/vue/src/store/mutations.js
@@ -86,13 +86,12 @@ export function updateTydeProgress(state, {parentId, isParentModule}){
   if(isParentModule){ // parentNode is a module, and all children are stages
     const childProgress = childNodes.map(stage => stage.tydeProgress)
     const reducer = (accumulator, progress) => accumulator + progress
-    var curProgress = childProgress.reduce(reducer, 0) / childNodes.length
+    parentNode.tydeProgress = childProgress.reduce(reducer, 0) / childNodes.length
   } else { // parentNode is a stage, and all children are nodes (topics)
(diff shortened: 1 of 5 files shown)


templates/vue/src/store/mutations.js
```diff
@@ -78,3 +78,36 @@ export function updateEntry(state, { answerType, entry, nodeId, questionId }) {
...
+  const childNodeIds = getChildIds(state, parentId)
+  const childNodes = childNodeIds.map(id => state.nodes[Helpers.findNodeIndex(id, state)])
+  if(isParentModule){ // parentNode is a module, and all children are stages
+    const childProgress = childNodes.map(stage => stage.tydeProgress)
+    const reducer = (accumulator, progress) => accumulator + progress
+    var curProgress = childProgress.reduce(reducer, 0) / childNodes.length
+  } else { // parentNode is a stage, and all children are nodes (topics)
+    var childProgress = childNodes.map(topic => topic.completed)
+    const reducer = (accumulator, completed) => accumulator + (completed === true ? 1 : 0)
+    var curProgress = childProgress.reduce(reducer, 0) / childNodes.length
```
wynnset: This line seems to be repeated in the `if` and `else` statements. Please move it out of that conditional. You actually don't even need to assign it to `curProgress`, rather directly assign it to `parentNode.tydeProgress` below.

wynnset: You forgot to move it out of the if/else statement, but no worries - I will make this change and commit it before merging this in.

PR #376, opened 2020-03-18
Update accordion progress to be computed based on rows

### LINKED ISSUE #375: Calculate accordion progress on load
The accordion progress should be computed rather than being saved on the db

narendrasss:
Closes #375

* Accordion progress now computed based on row completion
* Recompute progress when a row gets completed/added/deleted


wynnset:
@narendrasss please fix the conflicts here as soon as you get a chance and then reassign this back to me.

PR #378, opened 2020-03-18
Improve accordion ux (master)

narendrasss:
- [x] Limit video height to <= lightbox height
- [x] Scroll to accordion row when completed


wynnset:
Requires a fix as per latest comment on #383

PR #383, opened 2020-03-19
Improve accordion ux (tyde)

### LINKED ISSUE #377: Improve accordion UX
- [x] Make header non-sticky in TYDE
- [x] Limit video height to <= lightbox height
- [x] Scroll to accordion row when completed

narendrasss:
Closes #377

### DIFF
--- a/templates/vue/src/components/AccordionRow.vue
+++ b/templates/vue/src/components/AccordionRow.vue
@@ -28,10 +28,6 @@ export default {
   &:last-child {
     margin-bottom: 0;
   }
-
-  &:nth-child(2) {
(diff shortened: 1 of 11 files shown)


wynnset:
Hey @narendrasss just testing this - it seems that the accordion correctly now jumps down to the right part, but the video is still too tall and it flows out of my view. You can try zooming in on your browser maybe (Cmd +) and test to make sure the video always resizes to fit in.

image

narendrasss:
@wynnset Can confirm the video fits for me just fine, even after zooming in several times. Is this a video or an h5p? If it's an h5p, there might be something else happening.

PR #385, opened 2020-03-19
Allow authors to follow up to previous activities

### LINKED ISSUE #384: Show answer from a previous activity

narendrasss:
Closes #384

- [X] Add follow up to authoring tool
- [x] Display previous answer above activity
- [x] Fix entry not existing in question after the question gets completed


wynnset:
@narendrasss Audio answers are not appearing correctly. Similar to how they were appearing incorrectly in the summary activity. You can see here it's just printing "10" instead of showing an audio player: 
image

narendrasss:
New look:
Screen Shot 2020-04-02 at 13 22 07

wynnset:
Looks much better thank you. Just one more request please - Can we remove the original *question* and also allow the author to customize the text that is currently "Previously, you said:" and also make it a little bigger in size?

PR #388, opened 2020-03-24
Show module progress and planet view icon for modules

ItsAdel:
#379


ItsAdel:
> I think the styling should be updated to better reflect the mockup on Asana.

@wynnset Should I revert back to the original design?

PR #398, opened 2020-03-27
Add copilot role

narendrasss:
Closes #396 
- Add "copilot" as a custom role in WP
- Add "Teen ID" label to user profile if user is a copilot

PR #399, opened 2020-03-27
Show teen's progress if user is a copilot

### LINKED ISSUE #395: Show the teen's progress in a dyad tapestry
When the dyad opens a tapestry, they should see the progress of their teen for **all** nodes except for dyad-only nodes.

**Dyad-Only Nodes**
* A dyad-only node is a node that contains the dyad role permission.

**Other Nodes**
* The lock status should be ignored like how it is for authors.
* Any activity done should not be saved in the backend.

narendrasss:
Includes PRs #398, #394
Closes #395

**Steps to Test**

- From the admin dashboard, create a new user with the role of "copilot".
- Login with the copilot user, and in your profile scroll down to the "Teen ID" field. Type in "1", which is (usually) the ID of the admin user. The copilot is now linked to the admin user.
- To create a copilot only node, uncheck all of the permissions in the node except for "copilot".

**Features to Test**

- For copilot only nodes, the copilot should be able to normally update their progress.
- For non-copilot nodes, the copilot should not be able to update progress.
  - This includes quizzes.
- Copilots should be able to navigate and view locked nodes like the admin/author.
  - This should apply to locked accordion rows as well.


ItsAdel:
I tested this earlier, everything worked for me

narendrasss:
> I tested this earlier, everything worked for me

Thanks, I'll assign this back to me to fix the merge conflicts then assign it back to @wynnset after for code review.

narendrasss:
@wynnset Implemented the changes to the module summary as requested.

wynnset:
I think you may have missed a part of this task - Module progress bar and stage progress bar were showing 0 progress when logged in as co-pilot when it showed some progress when logged in as the teen. The idea here is that parents can view their teen's progress in the planet view and also if they view a module / stage. Progress bars should work and show the teen's progress.

Also, in the spaceship cockpit:
- Some modules were not showing up if they were not completed
- When clicking on a spaceship part, if the teen has not completed any activities, when I clicked on the "Activities" filter, it seemed to throw an error and display a weird very short box.

templates/tapestry.js
```diff
@@ -2581,17 +2583,17 @@ function tapestryTool(config){
...
         if (node.presentationStyle === "accordion-row" && !config.wpCanEditTapestry) return false;
 
-        // CHECK 5: Hide stage and question set nodes unless user is an editor - this check is for TYDE only
+        // CHECK 4: Hide stage and question set nodes unless user is an editor - this check is for TYDE only
         if ((node.tydeType === "Stage" || node.tydeType === "Question set") && !config.wpCanEditTapestry) {
             return false;
         }
+
+        // CHECK 5: If node is the root node or is currently selected
+        if (node.nodeType === "root" || (node.id == tapestry.dataset.rootId && node.nodeType !== "")) return true;
```
wynnset, line 2596: What is the point of having this here when the next statement after this returns true anyways? The original point of this was to return `true` before other things returned `false`.

narendrasss: This was causing a bug where when a copilot clicks on a topic to view in the module summary, that topic becomes viewable afterwards because it becomes the root. I see your point though, I'm not sure if just removing this would cause any issues but I don't remember experiencing any.

narendrasss:
The teen's activities are showing up for me (albeit it says 'You answered' rather than something like 'teen answered'). Can you list out what you did to make this not happen?

narendrasss:
Nevermind, I think you meant inside a gravity form. This makes sense, working on it now.

narendrasss:
@wynnset I've made the changes. The only thing is that only the button is disabled but if the user were to inspect element and submit it anyway, the back-end will allow it to go through.

I am able to prevent the node from getting completed in this case but the entry will still be written because it's handled by Gravity Forms on their end. I can work on this if you think it's necessary, but it might take time because I'm personally unfamiliar with how GF handles their form submission.

PR #403, opened 2020-04-01
Add profile tab

ItsAdel:
Added new profile tab in tapestry setting with the ability to add activities to be displayed in the backpack. Created a new backpack profile tab where activities are displayed. Fixed issue with logs iterating over null question entry

Testing: (Tyde)
- Create nodes and set the content type to activity
- Create a few questions
- Go to tapestry settings, then profile tab and be able to add activities.
- Should be able to see the activities (if answered) when you go to backpack, and then click the profile tab.
- The order should be maintained (how the questions are listed in tapestry settings and backpack)


wynnset:
As discussed, please refactor activities to be retrieved the same way as in the module summary and remove any unneeded code after that.

wynnset:
@ItsAdel can you please fix the merge conflicts on this and also not assign this to me since it's not been tested by anyone yet?

PR #419, opened 2020-04-07
Add nodes to favourites

ItsAdel:
Added button on the modal to favourite a node. Added hearts in accordion and subaccordion row to favourite them.

Steps to recreate:
Be logged in as a user
View any node
Should have a heart button next to the close button
Adding should toggle the button to unfavourite and vice versa
Refresh pages after adding a few nodes to favourites to ensure they are favourited when you view them again

Create accordions and multiple sub accordions.
In the accordion view, for every accordion row, notice there is a white heart. On click, should turn into a red heart representing it has been favourited.
Check that you can favourite and unfavourite accordion rows. (Should retain favourites upon refreshing the page)
Check that locked accordions don't allow or show the heart, so you can't favourite.


liam-armstrong:
image

I believe it should be spelled `favorites`
Also the text and button are slightly out of line, might want to make it a flex container and align them

Otherwise everything seems to be working well!

wynnset:
The Add node to favourites button is being placed out of bounds for nodes that open in full-screen. It also does not show in an accordion row or a subaccordion row (a user must be able to add a specific accordion row or a subaccordion row to their favorites). This needs to get fixed. You can have it appear at the bottom perhaps. For the accordion/subaccordions, perhaps you can place a heart icon where the lock icon goes (i.e. the right-hand side of the accordion row header). Of course it shouldn't appear there unless the row is unlocked.

The text should just read "Add to favorites" (this is the spelling outside of Canada and more international, so let's go with that).

PR #423, opened 2020-04-09
Changed cockpit modules to use percentage instead of pixels

ItsAdel:
#421


liam-armstrong:
Tested and working as specified on my machine. 
However, are we sure that we want to size images based on percentage? I understand positioning but managing size that way is a little counter-intuitive - to me at least.

PR #420, opened 2020-04-08
Allow authors to select and move multiple nodes together

narendrasss:
Closes #411

- Add node selection using `cmd`, `ctrl` and `shift` keys
- Add node selection using a selection box using the `DragSelect` library
- Select all nodes using `cmd+A`
- Clear selection by either clicking outside a node or pressing the `esc` key
- Selected nodes are currently highlighted orange, but can be easily changed if need be

**Testing**

- Go to a tapestry that you've previously made
- Select around using the commands above

PR #432, opened 2020-04-14
Allow authors to select and move multiple nodes together (#420)

wynnset:
* Drag and release to select nodes
* Ctrl/Cmd+A to select all nodes
* Ctrl/Cmd + click to select current node plus every other selected node and this one


templates/single-tapestry.php
```diff
@@ -114,6 +114,7 @@ function enqueue_vue_app_build()
         <script src="<?php echo plugin_dir_url(__FILE__) ?>libs/jscookie.js" type="application/javascript"></script>
         <script src="<?php echo plugin_dir_url(__FILE__) ?>libs/d3.v5.min.js" type="application/javascript"></script>
         <script src="<?php echo plugin_dir_url(__FILE__) ?>libs/h5p-resizer.min.js" charset="UTF-8"></script>
+        <script src="<?php echo plugin_dir_url(__FILE__) ?>libs/dragselect.min.js"></script>
```
wynnset, line 117: I've fixed this one - you had this point to a github repo. It's always better to import the script locally instead or at minimum an external CDN that is reliable.

templates/tapestry.js
```diff
@@ -44,7 +44,7 @@ function tapestryTool(config){
         simulation,                                             // Force
         tapestrySlug, 
         saveProgress = true,                                    // Saving Progress
-        tapestryDimensionsBeforeDrag, nodeBeforeDrag,
+        tapestryDimensionsBeforeDrag, nodesBeforeDrag,
```
wynnset, line 47: Doesn't seem like this is used at all.

templates/tapestry.js
```diff
@@ -299,6 +299,8 @@ function tapestryTool(config){
             $("#" + TAPESTRY_CONTAINER_ID + " > svg").prepend(nodeLinkLine);
             recordAnalyticsEvent('app', 'load', 'tapestry', tapestrySlug);
         }
+
+        initializeSelection();
```
wynnset, line 303: Can you call this `initializeDragSelect` instead?

templates/tapestry.js
```diff
@@ -637,72 +639,6 @@ function tapestryTool(config){
             });
     }
 
-    // D3 DRAGGING FUNCTIONS
```
wynnset, line 640: Can you please move this back up here? It shouldn't be down there as it is out of place. Also hard to review the changes like this.

templates/tapestry.js
```diff
@@ -901,6 +837,78 @@ function tapestryTool(config){
...
+            },
+            onElementUnselect: node => {
+                const id = node.id.split("node-")[1]
+                selection.delete(tapestry.dataset.nodes[findNodeIndex(id)])
+            }
+        });
+    }
+
+    const selection = createSelection();
+    let isMultiSelect = false;
```
wynnset, line 861: Variables are all declared at the beginning of this file.

templates/tapestry.js
```diff
@@ -901,6 +837,78 @@ function tapestryTool(config){
...
+                data.forEach(d => {
+                    selection.delete(d);
+                })
+            },
+            forEach(fn) {
+                data.forEach(fn);
+            }
+        }
+        return selection;
+    }
```
wynnset, line 911: Can you please move this into where it says "ADD EDITOR ELEMENTS" and put the functions declarations all after the event listener additions?

templates/tapestry.js
```diff
@@ -1052,12 +1117,94 @@ function tapestryTool(config){
...
+            if (canEditNode(nd) && !autoLayout) {
+                $.ajax({
+                    url: config.apiUrl + "/tapestries/" + config.wpPostId + "/nodes/" + nd.id + "/coordinates",
+                    method: API_PUT_METHOD,
+                    data: JSON.stringify({x: nd.x, y: nd.y}),
+                    error: function(e) {
+                        alert("Sorry, there was an error saving the coordinates of this node!");
+                        console.error(e);
+                        //nd[xORfx] = nodesBeforeDrag.x;
+                        //nd[yORfy] = nodesBeforeDrag.y;
```
wynnset, line 1194: Please do not leave commented out code without a comment why they are commented out. Please either remove or leave a comment as to why it's commented out.

templates/tapestry.js
narendrasss: Not sure how you got that idea - both variables seem to be used to me.

templates/tapestry.js
narendrasss: I understand the difficulty in reviewing changes, but I feel it's better suited down below since it's only being used in one place only, i.e. where we're setting up node dragging. I feel it's more difficult having to jump to different places in the file when working on the drag functionality.

templates/tapestry.js
narendrasss: Why should this be the case? For consistency?

I left it here because it's only being used in one place.

templates/tapestry.js
wynnset: `nodesBeforeDrag` is not used. Can you point out where it's used to me?

wynnset: Sorry I see you are now using it. Before it was not being used (only somewhere commented out).

narendrasss:
> When you press cmd/ctrl and click on a node, the current node also gets selected. This is unexpected behavior that shouldn't be happening

Is this unexpected? On most things (e.g. file systems) when something is already selected cmd includes whatever is currently selected as well. I can change this of course.

wynnset:
> Is this unexpected? On most things (e.g. file systems) when something is already selected cmd includes whatever is currently selected as well. I can change this of course.

Hey, to be clear, I don't mean when a node is selected to move, but when a node is the current node being viewed (i.e. the node is appearing larger in d3). If a node is not highlighted, it should not be assumed that it's selected. Reach me on Slack if you're still not clear on this.

PR #436, opened 2020-04-16
435 Planet View Updates

### LINKED ISSUE #435: Planet View Updates
- Add checkmark when a planet is complete
- Hide progress bar if module not started or complete
- Hide planet view icon if no icon present

liam-armstrong:
Closes #435 
Adds:
 - Check-mark icon to completed modules in planet view
Changes:
 - Hides planet view progress bar if progress is 0
 - Hides planet view icon if no image URL exists

Steps to test:
 - Create a TYDE tapestry with a module, stage and 2 topics
 - Leave both planet view icon links empty for the module
 - Verify the progress bar is not visible before completing any topics
 - Complete one topic and verify the progress bar has appeared, but without the icon
 - Complete the other topic and verify there is a check-mark in the bottom corner of the planet
 - Add a Planet View Completed icon link to the module and verify the icon appears
 - Add a new topic to the stage
 - Add a Planet View Incomplete icon link to the module and verify the icon appears


ItsAdel:
For the "Complete the other topic and verify there is a check-mark in the bottom corner of the planet", the checkmark doesn't appear right away. It seems to only appear after a page refresh

wynnset:
On Asana, the requirement is "Hide progress bar if module not started or complete". In this PR, the progress bar is only being hidden if progress is 0. Not sure why the other part of this was not implemented. It needs to hide if progress = 1 as well.

wynnset:
Ok I fixed that issue so the progress bar is now being hidden for completed modules. There is still the problem that Adel reported. I just tested it and it's not fixed yet. Checkmark must show up as soon as the module gets completed without requiring any further action. This is currently not the case.

PR #448, opened 2020-04-22
Fix not being able to delete combobox values

### LINKED ISSUE #438: "question" - when I add a form I cannot delete it
When adding a question to one of the nodes, I accidentally inputted a gravity form number in to the "textbox gravity form" input box.

But, this question did not require any textbox gravity form. So, I want to remove this text. But, I cannot remove the text -- it keeps autofilling with the gravity form number I inputted.

narendrasss:
Fixes #438

### DIFF
--- a/templates/vue/src/components/Combobox.vue
+++ b/templates/vue/src/components/Combobox.vue
@@ -46,8 +46,9 @@ export default {
       required: true,
     },
     value: {
-      type: [Object, String],
-      required: true,
+      type: [Object, String, null],
+      required: false,
+      default: null,
     },
     emptyMessage: {
       type: String,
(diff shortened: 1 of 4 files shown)


templates/vue/src/components/Combobox.vue
```diff
@@ -46,8 +46,9 @@ export default {
       required: true,
     },
     value: {
-      type: [Object, String],
-      required: true,
+      type: [Object, String, null],
```
wynnset, line 49: @narendrasss is adding `null` necessary here? I noticed that in the TYDE version of this PR you didn't include this and so was wondering which one is correct (I think since `required` is false, `null` is not necessary to be added in the `types` array, but wanted to confirm)

narendrasss: @wynnset Oh yeah I looked it up and it’s not necessary, forgot to update it on the base branch.

wynnset: Np I'll update it thanks

PR #450, opened 2020-04-22
Fix accordion not auto scrolling for some content types

### LINKED ISSUE #449: Subaccordion not auto-scrolling for some content types

narendrasss:
Fixes #449 

- Also fix iframes and h5p videos sometimes rendering too small within the subaccordions
- Remove h5p resizer lib
- Remove scrollbars on videos and h5p content types


templates/vue/src/components/lightbox/H5PIframe.vue
```diff
@@ -175,36 +174,6 @@ export default {
...
-              .height(width * $(this).data("ratio"))
-          } else {
-            $(this)
-              .height(height)
-              .width(height / $(this).data("ratio"))
-          }
-        })
-      }
-      $(window).resize(setIframeDimensions)
-      setIframeDimensions()
```
narendrasss, line 206: I removed this code because it was causing the `h5p` video to resize to the size of the subaccordion row, making it too small. This didn't seem to have any visible effects to regular h5p videos.

wynnset: Ok. I tested it a bit as well and it seems to work fine except when you view the H5P video in "fullscreen", the video sometimes gets cut off at the bottom. That might be something we have to address at a different time.

templates/vue/src/components/lightbox/H5PIframe.vue
```diff
@@ -175,35 +168,25 @@ export default {
...
+        const setIframeDimensions = () => {
+          const { width, height } = frame.getBoundingClientRect()
+          const parentHeight = frame.parentNode.getBoundingClientRect().height
+          if (height > parentHeight) {
+            const ratio = height / width
+            if (ratio < 1) {
+              const newWidth = width * (parentHeight / height)
+              frame.style.width = newWidth + "px"
+              frame.style.height = parentHeight + "px"
+            }
```
narendrasss, line 184: This works, but I don't necessarily like that I'm going into the DOM for this. I feel that there is a more elegant solution, but would require changing the html in other places of the lightbox. I was already going over 2 hours working on this so I thought I would leave it as is for now.

wynnset:
Sorry @narendrasss this is still not working for me :( It seemed to not be cut off the first time I opened a fullscreen video, but then I cleared my cache and refresh and after that it was cut off at the bottom again.

wynnset:
Message me if you want me to send you the H5P video I used (too big to attach here)

wynnset:
The H5P video is appearing super short in the sub-accordion...
image

wynnset:
Nevermind, I just realized you fixed this specific issue in #470

PR #452, opened 2020-04-24
429 star celebration

ItsAdel:
Modified accordion completion modal to show module progress at the top. Added star celebration modal when a stage is completed.

Testing
Create a module and stage with a question set
Question set should be an accordion
Finish the accordion to see the accordion finish modal with a progress bar on top which is the same as the modal progress bar.
If you create a regular accordion outside of stage and the accordion finish modal pops up, we shouldn't see a progress bar

In you stage, if you finish all the question sets, see a star celebration pop up with music playing.
Should only happen when a stage is completed and the lightbox of the node(which completed the stage) closing triggers it.


liam-armstrong:
Accordion celebration appears as expected
Module celebration does not, however. This appears to be a result of the tydeProgress not being updated, and so the module isn't "complete" when I close it.
Take this with a grain of salt because I've been having issues with my local dev environment recently; so someone else may want to test this

liam-armstrong:
> 
> 
> Accordion celebration appears as expected
> Module celebration does not, however. This appears to be a result of the tydeProgress not being updated, and so the module isn't "complete" when I close it.
> Take this with a grain of salt because I've been having issues with my local dev environment recently; so someone else may want to test this

This is fixed after correcting accordion progress in branch #459 (PR #460)

PR #456, opened 2020-04-24
Add nodes to favourites (TYDE)

ItsAdel:


wynnset:
Found some issues with buttons not showing up and also with the favourite button mistakenly toggling an accordion. I've fixed both of these, but @ItsAdel please make sure to test your PR more before you put it into "ready for review" in the future.

PR #460, opened 2020-04-27
Fix accordion completion not triggering stage tydeProgress update

### LINKED ISSUE #459: Accordion TydeProgress not Updating
Completing an accordion does not trigger a tydeProgress update on parent stages without refresh

liam-armstrong:
Closes #459 
Moves logic from TapestryMedia to Lightbox

Steps to test:
Create a module, stage, accordion and child text node in order
Refresh page to ensure accordion progress is correct
Ensure module is not complete before child text node is 
Launch module, see stage screen, open accordion
Complete text node
Verify stage and module become complete


ItsAdel:
Everything seems to work

wynnset:
Closing this PR as it is already being merged in #452

PR #461, opened 2020-04-28
Show special icon for activity as well in accordion row

ItsAdel:
#458 

Testing:
Create an accordion node - node1
Create a child accordion node, child text node and child activity node from node 1
When you view accordion node1, check that special icon appears for both rows representing the accordion and activity but not the text node

### DIFF
--- a/templates/vue/src/components/lightbox/AccordionMedia.vue
+++ b/templates/vue/src/components/lightbox/AccordionMedia.vue
@@ -396,6 +396,7 @@ button[disabled] {
 
   &-activity {
     width: 52px;
+    margin-right: 10px;
   }
 }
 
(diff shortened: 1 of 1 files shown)


liam-armstrong:
image

Icon appears for activity but not sub-accordion. No clues in the console

narendrasss:
Hey @liam-armstrong did you end up working on this? Is it ready for review?

liam-armstrong:
> Hey @liam-armstrong did you end up working on this? Is it ready for review?

It is ready for review, as the actual Asana task doesn't specify an icon for sub-accordions. However, Aidin also mentioned that having the activity icon appear in sub-accordion rows would be good, but it's not a requirement

PR #463, opened 2020-04-29
If a parent node has a certain permission, its children should be limited to that permission too

ItsAdel:
Fixed issue with children nodes of parent nodes without permission to appear on-screen appearing
#462 

Testing:
Create node1 and multiple child nodes
Edit node 1 and in access tab change, uncheck read

In incognito mode (or log out of your account), check that the child nodes don't appear when refreshed


liam-armstrong:
Tested and working for me!

PR #465, opened 2020-04-29
Fixed issue with unchecking lock nodes checklist

ItsAdel:
#464 

Testing:
Create new nodes
Make nodeA dependent on nodeB through the access tab and locking it after checking "Prevent Access Until Specified Conditions Are Met"
Check nodeA is locked
Go back to authoring tool and uncheck "Prevent Access Until Specified Conditions Are Met". Hit submit

Ensure that the lock disappears


liam-armstrong:
Tested and working

PR #466, opened 2020-04-29
Improve algorithms for finding node parent & more

wynnset:
- Use a smarter algorithm for finding parent of a node
- Take into account the tydeType and child ordering when finding parents
- Update getParent algorithm everywhere
- Improve node type validation
- Remove obsolete code

### DIFF
--- a/classes/class.tapestry.php
+++ b/classes/class.tapestry.php
@@ -140,9 +140,7 @@ class Tapestry implements ITapestry
             $parent = $this->getNode($parentId)->get();
         }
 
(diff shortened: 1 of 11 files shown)


liam-armstrong:
@wynnset Are there any specific features to test for this PR?

wynnset:
@wynnset to fix merge conflicts and then assign to @narendrasss for testing and code review

wynnset:
This is for an obsolete version of Tapestry/TYDE and is no longer necessary. Tapestry 3.0 is going to vastly change all of these algorithms as well, so this is no longer needed.

PR #470, opened 2020-04-30
Fix videos getting clipped in some cases

### LINKED ISSUE #469: Videos sometimes get clipped

narendrasss:
Fixes #469 

Testing steps:

- Create a video node and choose a video that is wider than it is tall
- Set the node appearance to fullscreen
- Resize your browser so it is much wider than it is tall
- The video should fit the browser and be centered


ItsAdel:
I tested it using the listed testing steps. However, after making the browser wider than tall and opening the video. It didn't resize the video and the image is getting clipped off

PR #474, opened 2020-05-01
Default Node Permissions and Access Tab Control

### LINKED ISSUE #473: Allow author to set default permissions
Authoring side:
If the "Add" permission is checked (for any rows) in the Access tab:
    Add a new section under the permissions called "Default permissions for new added nodes" and allow author to define those. The default here should be what the default is currently ("Read" for "public")
    Also add a switch that says "Show/Hide Access tab" (default: Show)

User side:
When a user is adding a new node, set the default permissions based on what the author set and show/hide the access tab also as per the author

liam-armstrong:
Closes #473 
#### Adds:
- Access tab to settings modal
- `defaultPermissions` settings field
- `viewAccess` settings field
- Node initialization with default permissions in both front and backend
- Access tab viewing control based on edit permissions and settings 
____
#### Steps to test:
Default Permissions:
 - Create a new node and verify the permissions, without modification, are set to `public: read` and `authenticated: read`.
 - Go to the settings modal and modify the default permissions in the Access tab
 - Create another node and verify the default permissions have changed

Show/Hide Access Tab:
 - Give public edit/add permission to a node
 - Open an incognito window and click to edit the node
 - Verify you can see the Access tab as an un-authenticated user 
 - Switch back to the logged in window and flip the switch in settings modal to disable Access tab viewing
 - Verify you (the author) can still see and modify the Access tab for any node while logged in
 - Log out/switch to incognito and click to edit that same node
 - Verify the Access tab is hidden


narendrasss:
I'm getting this error when opening an old tapestry without the default permissions:

```
<br />
<b>Notice</b>:  Undefined property: stdClass::$defaultPermissions in <b>/Users/narendrasyahrasyad/Documents/Career/ubc-psych/wp-content/plugins/tapestry/utilities/class.tapestry-node-permissions.php</b> on line <b>25</b><br />
{"nodes":[{"id":2036,"author":"1","type":"tapestry_node","size":"","title":"root","status":"publish","imageURL":"","lockedImageURL":"","mediaType":"text","mediaFormat":"","mediaDuration":0,"description":"qwe","behaviour":"embed","typeData":{"linkMetadata":null,"progress":[{"group":"viewed","value":0},{"group":"unviewed","value":1}],"mediaURL":"","mediaWidth":960,"mediaHeight":600,"subAccordionText":"More content:","textContent":"qwe"},"coordinates":{"x":2783.1943359375,"y":3220.636474609375},"permissions":{"public":["read"],"authenticated":["read"]},"hideTitle":false,"hideProgress":false,"hideMedia":false,"skippable":true,"quiz":[],"fullscreen":false,"conditions":[],"childOrdering":[]}],"groups":[],"links":[],"settings":{"tapestrySlug":"jarring-node-drag","title":"jarring-node-drag","status":"publish"},"rootId":2036}
```

templates/vue/src/components/SettingsModal.vue
```diff
@@ -74,10 +149,30 @@ export default {
...
+        public: ["read"],
+        authenticated: ["read"],
+      },
+      permissionsOrder: [
+        "public",
+        "authenticated",
+        ...Object.keys(wpData.roles).filter(
+          role => role !== "administrator" && role !== "author"
+        ),
+      ],
```
wynnset, line 162: Can you please pass these two as props into this component (from `Tapestry.vue`) similar to how `permissionsOrder` is passed into NodeModal? That way we only have to set it in one place.

templates/vue/src/components/SettingsModal.vue
```diff
@@ -98,23 +193,90 @@ export default {
         backgroundUrl = "",
         autoLayout = false,
         nodeDraggable = true,
+        defaultPermissions = {
+          public: ["read"],
+          authenticated: ["read"],
```
wynnset, line 198: Is this serving a purpose? If you are setting it here, you shouldn't set it when declaring it (or vise-versa).

templates/vue/src/components/SettingsModal.vue
```diff
@@ -98,23 +193,90 @@ export default {
...
+      const rowIndex = this.getPermissionRowIndex(rowName)
+      const lowerPriorityPermissions = this.permissionsOrder.slice(rowIndex + 1)
+      lowerPriorityPermissions.forEach(newRow => {
+        this.changeIndividualPermission(value, newRow, type)
+      })
+    },
+    getPermissionRowIndex(rowName) {
+      return this.permissionsOrder.findIndex(thisRow => thisRow === rowName)
+    },
+    isPermissionDisabled(rowName, type) {
```
wynnset, line 235: Since this function (and maybe some other ones) are now being used in two different components, can you please turn them into helper functions so that it's only defined in one place? Then you can utilize those helper functions in both components. Again, we want to avoid having to update things in multiple spots in the application if we need to make changes in the future - and also that way there's less code volume.

liam-armstrong: I've moved this function to Helpers, but long term it might be best to turn the entire permissions table into it's own component, seeing as we are repeating the same functionality.

templates/vue/src/components/SettingsModal.vue
```diff
@@ -98,23 +193,84 @@ export default {
...
+        Helpers.onlyContainsDigits(userId) &&
+        $("#user-" + userId + "-editcell").val() != ""
+      ) {
+        this.$set(this.node.permissions, `user-${userId}`, this.newPermissions)
+        this.permissionsOrder.push(`user-${userId}`)
+        this.userId = null
+      } else {
+        alert("Enter valid user id")
+      }
+    },
```
wynnset, line 273: I believe all these functions I have highlighted here can be moved to the Helper. If it's possible to create a component here and use it in both places, that would be even better so that we don't have to use a helper function.

wynnset:
Getting an error when trying to add a user-based permission in the Default Permissions section: 
image

narendrasss:
@wynnset Oops, I'm going to need the branch still active here because the TYDE version isn't up to date.

PR #479, opened 2020-05-04
Add Youtube support and auto-calculate video length for all videos

### LINKED ISSUE #477: Add YouTube support

liam-armstrong:
Closes #477 
#### Adds:
- Embedded YouTube player for video nodes (with type autodetection)
- Automatically populating `mediaDuration` field for YouTube and H5P videos
- Use H5P video settings to determine YouTube player options, and updating settings from YouTube player
____
#### Steps to test:
Basic Support:
 - Run NPM install to get YouTube component
 - Create a video node and fill in the URL section with any YouTube video link
 - After submission, verify the correct video duration appears on the node (under the title)
 - Click the media button and press play on the i-frame to begin the video

Progress:
 - Play a meaningful (> 2%) amount of the video
 - Pause the video and exit, verifying the progress bar updates 
 - Re-enter the video, and verify it begins from where you left it
 - Watch another portion, and close without pausing; verify the progress bar updates
 - Finish the video (or scroll to the end), and verify the end screen appears with the option to close, rewatch or take quiz if one is set up
 - Verify the node is marked as complete on the Tapestry

Settings:
 - Create an H5P video node
 - Open the video and alter the settings; mute, change the volume and playback speed
 - Open the YouTube video node and verify the appropriate settings have been changed. Note that playback speed changes may not work for all YouTube videos
 - Alter the settings in the YouTube player and either close or pause the video to save the changes
 - Open the H5P video node again, and verify the settings changes have saved and applied


narendrasss:
Getting a 403 Forbidden error when I paste a YouTube URL and press submit on the node modal. Is there some API key that I have to set up? If so, can you add it to the description?

liam-armstrong:
> 
> 
> Getting a 403 Forbidden error when I paste a YouTube URL and press submit on the node modal. Is there some API key that I have to set up? If so, can you add it to the description?

I've removed the API call that was requiring this, feel free to test it again for me!

narendrasss:
I'm able to submit the node now, but the video iframe fails to load:

Screen Shot 2020-05-14 at 13 07 34

Screen Shot 2020-05-14 at 13 07 41

Not sure if it's the specific video I chose, but it should work with any supplied video.

liam-armstrong:
> Screen Shot 2020-05-14 at 13 07 41
> 
> Not sure if it's the specific video I chose, but it should work with any supplied video.

The link I pulled from that console error ( https://www.youtube.com/watch?v=-_GuBr0hBby ) doesn't appear to be a valid YouTube URL. I get the same MIME error screen with it, but other videos are working on my machine. Could you try testing again with a different video?

wynnset:
This is also not working for me. I tried adding a video with this URL: https://www.youtube.com/watch?v=Bmv1nslpK8s

It seems that the URL is being used as the `src` attribute of a `video` element as if it's a video file. YouTube videos need to use the YouTube embed code as discussed. Here's a screenshot from my computer (Mac/Chrome):

image

wynnset:
YouTube is working correctly now, but H5P videos are no longer being sized properly and I get all these errors in my console when opening up an H5P video:

image

@liam-armstrong please try to do more testing and ensure no console errors are appearing before you submit the PR back. It creates a lot of back and forths that could be avoided if you just take a couple more minutes to test all the pieces of the code that you touched please -- Thank you.

wynnset:
Seems to be working now! Thanks

templates/vue/src/components/NodeModal.vue
```diff
@@ -560,6 +581,15 @@ export default {
...
       return ordered
     },
+    videoUrlYoutubeID() {
+      return this.videoUrlEntered
+        ? Helpers.getYoutubeID(this.node.typeData.mediaURL)
+        : ""
+    },
+    accessSubmit() {
+      // Locks access to submit button while youtube video loads to grab duration
+      return this.videoUrlYoutubeID === "" ? true : this.youtubeLoaded
```
wynnset, line 591: It would be more readable if you return `this.videoUrlYoutubeID === "" || this.youtubeLoaded` instead. Also, I think this may be what's causing what Narendra mentioned where the loading button is always present since you are only checking for YouTube.

templates/vue/src/components/NodeModal.vue
```diff
@@ -685,6 +720,13 @@ export default {
     submitNode() {
       this.formErrors = this.validateNode(this.nodeData)
       if (!this.formErrors.length) {
+        if (
+          this.node.mediaType === "video" &&
+          this.nodeType !== "h5p" &&
+          this.videoUrlYoutubeID === ""
```
wynnset, line 726: It would be more readable if you change this and other similar cases in this file to a computed value.

templates/vue/src/components/TapestryMedia.vue
```diff
@@ -147,7 +161,15 @@ export default {
...
   height: 100%;
+  padding: 0;
+
+  &-no-scroll {
+    overflow: hidden;
+  }
+}
 
+.media-wrapper-embed {
+  background: white;
```
wynnset, line 172: Please remove this and move the `&-embed` below under the above directive like it was before. I'm assuming there is no `media-wrapper-embed-embed` class, which is what is implied here.

templates/vue/src/components/lightbox/H5PIframe.vue
```diff
@@ -266,6 +264,7 @@ export default {
           h5pVideo.on("loaded", handleH5pAfterLoad)
         }
       }
+      this.$emit("is-loaded")
```
wynnset, line 267: Can you see if you can get rid of this event and use the `loaded` event instead that is already getting emitted at the right place? Unless they serve different functions - but it seems to me that this is emitting at the wrong time since technically it's not yet loaded until the `handleH5pAfterLoad` is called. (Please be careful with this and make sure the logic makes sense - I'm just observing the code without too much deep diving here)

templates/vue/src/main.js
```diff
@@ -5,7 +5,9 @@ import store from "./store"
 import router from "./router"
 import "bootstrap/dist/css/bootstrap.css"
 import "bootstrap-vue/dist/bootstrap-vue.css"
+import VueYouTubeEmbed from 'vue-youtube-embed'
```
wynnset, line 8: Do we have to import this here? Can we not do the import in a more specific component instead?

liam-armstrong: The [documentation](https://github.com/kaorun343/vue-youtube-embed) doesn't present another way to do this I'm afraid.

templates/vue/src/components/lightbox/H5PIframe.vue
liam-armstrong: This event already existed in the H5P iframe and seems to play a role in loading the video - I moved it to the bottom of the function when dealing with another bug, but as it stands the event isn't used in this PR. I refactored the H5P progress loading to just use a vanilla i-frame.

templates/vue/src/components/NodeModal.vue
```diff
@@ -398,8 +406,15 @@
       <b-button size="sm" variant="secondary" @click="$emit('close-modal')">
         Cancel
       </b-button>
-      <b-button size="sm" variant="primary" @click="submitNode()">
-        Submit
+      <b-button
+        id="submit-button"
+        size="sm"
+        variant="primary"
+        :class="accessSubmit ? '' : 'disabled'"
```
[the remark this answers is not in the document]
liam-armstrong: The `:disabled` attribute was bringing in some weird bootstrap CSS, so I kept this as a CSS solution and disabled clicker events that way

narendrasss:
Looks good so far, I made some minor stylistic changes and ran the linter again.

One thing though - the duration doesn't seem to be calculated for H5P videos created with Youtube links. There are no console errors.

narendrasss:
Awesome, works great now.

PR #480, opened 2020-05-05
TYDE: Fix H5P settings not being saved correctly

narendrasss:


templates/vue/src/components/tyde/TydeActivitySummary.vue
```diff
@@ -49,9 +49,6 @@ export default {
       }))
     },
   },
-  mounted() {
-    const entries = Object.entries(this.activity.entries)
-  },
```
narendrasss, line 54: This is different from the master version because it was causing a lint error that existed in TYDE only.

PR #481, opened 2020-05-05
Add ability to filter nodes based on certain attributes

### LINKED ISSUE #478: Add ability to filter nodes based on certain attributes

narendrasss:
Closes #478

**General Filtering**

- [x] Add filter component that computes filters based on current window URL
- [x] Should have two dropdowns: one for selecting which attribute to filter on, and the other the attribute value
- [x] On the D3 side, watch for changes in the window URL and update viewable nodes based on this filter

**Filtering by Author**

- [x] Save author information when node is created

**Testing**

- Create a new Tapestry as an admin
  - Make sure that either your default permissions allow add access for authenticated users or when you're adding a node that node has add access
- Login as another user
- Add some nodes
- Press the search button on the top left corner
- On the second combobox, select between the users
- The nodes not made by the user should have lower opacity (unless they're your nodes)

### DIFF
--- a/classes/class.tapestry-node.php
+++ b/classes/class.tapestry-node.php
@@ -251,7 +251,7 @@ class TapestryNode implements ITapestryNode
     {
         wp_update_post(array(
             'ID'            => $this->nodePostId,
-            'post_author'   => $this->author->id
+            'post_author'   => $this->author['id']
         ));
     }
 
--- a/classes/class.tapestry-node.php
+++ b/classes/class.tapestry-node.php
@@ -50,7 +50,7 @@ class TapestryNode implements ITapestryNode
         $this->nodePostId = 0;
         $this->nodeMetaId = (int) $nodeMetaId;
 
-        $this->author = wp_get_current_user()->ID;
+        $this->author = $this->_getAuthorInfo(wp_get_current_user()->ID);
         $this->conditions = [];
         $this->size = '';
         $this->title = '';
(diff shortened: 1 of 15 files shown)


liam-armstrong:
Upon loading an old tapestry as admin, I get an ever-increasing amount of `links is undefined` and nothing loads except the search icon. 
image
Error does not show up for tapestries made on this PR

liam-armstrong:
On a new tapestry, this is only highlighting nodes where I am the author, if that is the filter I have selected. Selecting the other user's filter keeps all nodes highlighted. So, while logged in as `liam`, filtering for nodes by `liam2` (of which there is only 1) results in:
image
and filtering for nodes created by `liam` when logged in as `liam`: 
image
Works as expected. 

Same issue appears on other account, where I can only properly filter for nodes created by `liam2`

narendrasss:
> On a new tapestry, this is only highlighting nodes where I am the author, if that is the filter I have selected. Selecting the other user's filter keeps all nodes highlighted. So, while logged in as `liam`, filtering for nodes by `liam2` (of which there is only 1) results in:
> image
> and filtering for nodes created by `liam` when logged in as `liam`:
> image
> Works as expected.
> 
> Same issue appears on other account, where I can only properly filter for nodes created by `liam2`

This is the expected behaviour, as per the requirements your own nodes will always be highlighted regardless of the filter that you chose.

wynnset:
Overall, this is working well but there are some minor issues that need ironing out:
- [ ] Filters are covering the site title (can we put this in the same place as the depth slider / settings icon? That would be ideal)
- [ ] After filtering, when you click to open a node, the filter gets reset. This is not ideal if an instructor is trying to go through the nodes added by a specific student
- [ ] When the filter gets open after the node was opened, sometimes the nodes are going out of the visible area (might be related to `originalDataset`)
- [ ] When the filter is opened, changing the depth slider doesn't actually change the depth

Please see this screen capture video to see these issues in action: https://drive.google.com/file/d/1D7akry3sw-cjsjd0DKCPZx_OmxU8VNRj/view?usp=sharing

narendrasss:
Closing this in favour of #503.

PR #486, opened 2020-05-08
Add new locking conditions based on date

### LINKED ISSUE #484: Unlock nodes based on date/time
Allow nodes to be locked/unlocked after a certain time

narendrasss:
Closes #484

* Adds two new conditions that allow authors to lock/unlock nodes based on current date
* Update authoring tool to show more descriptive conditions

**Testing**

* Create a new node with the new conditions
* Set a date with the date picker
* The node should be locked/unlocked according to the date that is set and the chosen condition

**Note**

This is implemented in a way that allows it to be completed quickly, that is I hard coded a `date` property into the condition object.

In the future, I think we should change the condition object itself to contain a `data` property instead of separate `nodeId` and `date` properties. This allows the condition to be flexible to whatever future condition we want the node to rely on. Since a lot of the code right now expects a `nodeId` at the top level of the condition object, this requires some refactoring in that regard.

For example, if we want to unlock a node based on the current tapestry progress, all we need is to add attributes to the `data` property rather than hard code these new values in.

**Todo**

- [x] Add support for time and timezone


liam-armstrong:
Not a bug, but we might want to adjust the wording of this message 
image

liam-armstrong:
Otherwise, feature is working as intended (Although I don't see any time-based locking)

wynnset:
Remove "To unlock this node:"
For time-based locking say: "This node will become available on Dec 23, 2019 at 9:00AM"
For node-completion-based locking say: "Complete [node title] first to view this node"

Can you also make this change: If there is at least 1 node-completion-based condition, make it so that if the user clicks on the locked node (or the condition text), we automatically navigate them to the node that they need to complete.

narendrasss:
@wynnset I thought about the text, and I think a better way to do it would be something like "This node will be unlocked:"
* After Dec 23, 2019 at 9:00AM
* When [node title] is completed

It reflects the change that I made in the authoring tool. What do you think?

templates/single-tapestry.php
```diff
@@ -118,6 +118,8 @@ function enqueue_vue_app_build()
         <script src="<?php echo plugin_dir_url(__FILE__) ?>libs/jscookie.js" type="application/javascript"></script>
         <script src="<?php echo plugin_dir_url(__FILE__) ?>libs/d3.v5.min.js" type="application/javascript"></script>
         <script src="<?php echo plugin_dir_url(__FILE__) ?>libs/dragselect.min.js"></script>
+        <script src="<?php echo plugin_dir_url(__FILE__) ?>libs/moment.min.js"></script>
+        <script src="<?php echo plugin_dir_url(__FILE__) ?>libs/moment-timezone-with-data-10-year-range.js"></script>
```
wynnset, line 122: If this is only until 2030, can you please change the filename to specify that it's 2020-2030 only?

narendrasss:
@wynnset What do you mean by switching the nodes?

PR #489, opened 2020-05-11
TYDE: Fix accordion not auto scrolling for some content types

narendrasss:


wynnset:
A couple of issues here:
1. H5P video is being cut off at the bottom in accordion for me and I cannot see the controls or the seek bar as a result.
2. The play screen is stuck for me when I open the video content that is already on the end screen and press "Replay Video".

You can see both of these issues in this screenshot:
image

wynnset:
Sorry the feedback above was for #490

narendrasss:
@wynnset I'll fix it on this branch instead since #490 deals only with mp4 videos.

narendrasss:
@wynnset I can't seem to replicate the H5P not fitting correctly. It wasn't correct on master, and I made #492 which addressed that. The only issue right now is that H5Ps made with YouTube videos automatically play when it is opened in an accordion. I'm not too sure how to fix that at the moment.

PR #490, opened 2020-05-11
TYDE: Fix videos getting clipped in some cases

narendrasss:


wynnset:
Please see #489 for feedback

PR #492, opened 2020-05-12
Fix play screen showing on rewatch and h5p not resizing correctly

narendrasss:
- Also fixes various linting errors in master


wynnset:
Looks good!

narendrasss:
@wynnset These changes were already applied in TYDE actually in #489. So this is good to go!

wynnset:
Oops.. I just encountered the play screen showing up in a regular video inside of an accordion. When I play the video, the play screen stays on.

PR #493, opened 2020-05-12
Do not return locked node content until node is actually unlocked

### LINKED ISSUE #485: Do not return content of locked nodes

narendrasss:
Closes #485

* Moves all locking functionality to backend
* Send node content only if the nodes are unlocked and accessible

**Testing**

* As a non-contributor, open a Tapestry with some locked nodes
* Verify that the locked nodes do not have their content through either the console or looking at the Vue extension and checking the Vuex store
  * Empty content in this case means the `quiz` property is an empty array and `typeData` only contains `progress` information
* Unlock the node
* Verify that the node unlocks without a need to refresh
* Open that node and verify the content looks correct


narendrasss:
Problems seen so far:

- [ ] When adding a child node, its accessible/unlocked status is not being set so it always appears locked until a refresh occurs.
- [ ] Conditions are not being updated. When a condition is fulfilled, the front-end doesn't reflect that it is (even on refresh).

liam-armstrong:
Tested as per the instructions and everything works as intended

templates/vue/src/components/node-modal/ConditionsForm.vue
```diff
@@ -1,6 +1,9 @@
 <template>
   <div>
-    <b-form-group class="mb-3">
+    <b-form-group
+      class="mb-3"
+      description="Users who are not logged in won't be able to unlock this node."
```
wynnset, line 5: Can you change this to say "Please note: Currently, locked nodes cannot be unlocked by users who are not logged in."

Also, please make it so it appears more like a warning and not a description. I suggest not showing it before the checkbox is checked, again so it doesn't appear like the description for that checkbox.

PR #499, opened 2020-05-14
Show favourites in spaceship cockpit

### LINKED ISSUE #498: Show favourites in spaceship cockpit

narendrasss:
Closes #498

**Testing**

* Create two modules with a few topics each
* Favourite several topics from each module
* Open the cockpit and click the favourites tab on the upper right corner
* The favourites from *both* modules should appear
* Open a module summary for one of the modules
* Only the favourites from that model should appear in the favourites tab


wynnset:
Thanks this is great progress so far.

I tested this and it seems that accordions are not showing properly (empty content):
image

Also, activities should not show the back button or anything. They should appear the way they show up in "previous answer".

wynnset:
When a video is set to open in full-screen, it's overflowing the favourite accordion trying to be in a full-screen size. It's bizarre. We need to make sure that the full-screen behaviour doesn't appear in the favourites or when the node is inside of an accordion in general.

narendrasss:
Labelling this as on hold because it's waiting for improvements to #501

PR #482, opened 2020-05-07
482 - Move D3 functionality into Vue

narendrasss:
Refactor out all functionality out of `tapestry.js` into Vue. The goal is to have more readable code and improve overall performance of Tapestries.

List of things that were previously in `tapestry.js` that **was not** moved to this PR:

* Autolayout -- I believe we reached a consensus to deprecate this feature
* `hardCodedDimensions` setting for improved performance -- I don't think this is a bottleneck in this PR so I've left them out

PR #512, opened 2020-05-19
509 - Update Cypress tests so we can run them individually or in portions

### LINKED ISSUE #509: Make Cypress tests more modular

narendrasss:
Closes #509

* Made tests independent so that they can be run individually
* Add several tapestries to the `fixtures` folder for testing use
  * A tapestry with only one root node
  * A tapestry with two nodes linked to one another
  * A complex tapestry with most current features (this was made a while ago so it's missing some more recent features)

I know we've had issues with Cypress not running on different machines, so I think it's best if both @wynnset and @liam-armstrong try running it before we merge this in.

Before running the suite, make sure to update your `cypress.json` config file to include the following line:

```
"baseUrl": "http://localhost:8888"
```

And verify that the rest of your config file looks like the `cypress-sample.json` file included in the repo:

```
{
  "env": {
    "ENV": "TYDE",
    "ADMIN_USERNAME": "admin_name",
    "ADMIN_PASSWORD": "admin_password",
    "SUBSCRIBER_USERNAME": "subscriber_name",
    "SUBSCRIBER_PASSWORD": "subscriber_password"
  },
  "baseUrl": "http://localhost:8888"
}
```


liam-armstrong:
image

 I get this error trying to run either test suite, on both Chrome 81 and Electron 78.

narendrasss:
> image
> 
> I get this error trying to run either test suite, on both Chrome 81 and Electron 78.
> E: on Cypress 3.8.3, if that helps

Oh sorry, please add this to your `cypress.json`:

```
"baseUrl": "http://localhost:8888"
```

I'll add it to the OP.

liam-armstrong:
image
After adding the baseUrl, I only make it to the first step. It looks like we're posting to an `undefined` variable as the link.

narendrasss:
@liam-armstrong I took a quick look and it's some config issue. Working on a fix right now

narendrasss:
Waiting until #581 has been thoroughly tested.

wynnset:
#581 is ready for merge. Please merge that then work on this.

liam-armstrong:
The command to run the tests is:
`docker exec e2e cypress run --headless --browser chrome`

narendrasss:
The following tests are failing because of a bug where there are 2 media buttons on each node:

- `Should be able to add a quiz to a video and have that quiz appear at the end of the video`
- `Should be able to add an external link node and have it show a summary of the external page`

wynnset:
> The following tests are failing because of a bug where there are 2 media buttons on each node:
> 
> * `Should be able to add a quiz to a video and have that quiz appear at the end of the video`
> * `Should be able to add an external link node and have it show a summary of the external page`
> 
> @wynnset Should we merge this in despite the failing tests? I was thinking the bug can be handled in a separate PR.

I would be OK with merging this if that's the only issue. We just need to make sure we address those tests that are failing. Is this the only thing holding this PR back from being merged @narendrasss ?

narendrasss:
@wynnset Yes I believe that's the only thing. On second thought though I might need to do some work to restructure this again after #482. Should I rebase this to merge into 482 and fix the tests there?

classes/class.tapestry.php
```diff
@@ -342,17 +342,15 @@ private function _recursivelySetAccessible($node, $visited, $nodeList)
...
+            foreach ($neighbourIds as $nodeId) {
+                foreach ($nodeList as $otherNode) {
+                    if ($otherNode->id === $nodeId) {
+                        array_push($neighbours, $otherNode);
                     }
-                },
-                $neighbourIds
-            );
+                }
+            }
```
wynnset, line 353: This is not related to this PR and should not be in here.

templates/vue/.gitignore
```diff
@@ -4,6 +4,8 @@ npm-debug.log
 yarn-error.log
 config.js
 cypress.json
+cypress/screenshots/
+cypress/videos/
```
wynnset, line 8: These should be removed since cypress is no longer here, right?

narendrasss: Cypress still generates these folders despite running inside the container, so I think it should be there still.

PR #519, opened 2020-05-21
WIP - Add ability to import/export/duplicate a Tapestry

### LINKED ISSUE #518: Add ability to import/export/duplicate a Tapestry

narendrasss:
Closes #518

### DIFF
--- a/endpoints.php
+++ b/endpoints.php
@@ -44,19 +44,26 @@ $REST_API_ENDPOINTS = [
             'callback'  => 'deleteTapestry'
         ]
     ],
+    'GET_TAPESTRY' => (object) [
+        'ROUTE'     => '/tapestries/(?P<tapestryPostId>[\d]+)',
+        'ARGUMENTS' => [
+            'methods'   => $REST_API_GET_METHOD,
+            'callback'  => 'getTapestry'
+        ]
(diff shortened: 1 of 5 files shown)


narendrasss:
Oh I forgot to do the Drag-Drop feature. I'll work on that, but I'll leave this PR up for testing because the core functionality is implemented already.

liam-armstrong:
The basic exporting/importing, and duplicating processes both work and feel very smooth. However, once I do either, and open the new Tapestry, the nodes go far away from the top left, and I can't change the root node. I can, however, still click the media buttons to view the content.

Tapestry JS throws this `node is undefined` error on startup:
image

and when I click on a mediabutton, these errors pop up:
image

From running a text diff on the Vuex stored nodes, it seems the copied nodes are missing the following elements: `index, x, y, vy, vx, nodeType, unlocked, depth, accessible`. Is that intended?

narendrasss:
Hey @liam-armstrong, thanks for this. It looks like the locking conditions of a node doesn't update with the new node IDs when it gets duplicated.

I'll work on it ASAP.

liam-armstrong:
Tested after the changes and both functions work very well

PR #526, opened 2020-05-25
TYDE: Add Youtube support and auto-calculate video length for all videos

liam-armstrong:
Same steps to test as #479


narendrasss:
I also found a small padding problem with the end screen:

Screen Shot 2020-06-01 at 11 19 52

No worries if you don't have time to fix this today, but please assign it to me if that's the case, I can take a quick look.

wynnset:
@liam-armstrong can you please update this PR now that the master PR is completed?

liam-armstrong:
The branch is up to date with it's master counterpart and is ready for testing!

wynnset:
Found a bug: If a node is set as an H5P video, then you edit it and try to change it to "Video", the options underneath do not change to show video-related fields, rather they keep showing H5P fields. Once this is fixed, we also need to push a change to fix this bug on master.

PR #534, opened 2020-05-30
533 - If user is able to add/edit, they should automatically be able to view that node too

### LINKED ISSUE #533: If user is able to add/edit, they should automatically be able to view that node too

narendrasss:
Closes #533

**Testing**

* Uncheck the read permission if it is currently selected
* Check the add or edit permission
    * The read permission should be checked and the checkbox disabled
* Uncheck the add or edit permission
    * The read permission should be unchecked
* Check the read permission
* Check then uncheck the add or edit permission
    * The read permission should be unaffected


liam-armstrong:
With the addition of the editor role, the component no longer loads as the `isPermissionsDisabled` function is called with the existing node table and the new role, which is not included in this value. This can be fixed by adding a blank array in the table for any encountered permissions, like so: 
image
But, I'm not sure if that's how we want to address this? Adding a new role to WordPress shouldn't break the permissions until we open the table component for each node. 

Otherwise the table works great, and the read permission is cascading properly.

narendrasss:
@liam-armstrong Thanks for pointing it out. After looking back at the PR again I noticed a couple of new issues too.

liam-armstrong:
Looks good, and properly handles the Editor field now. I did notice some unexpected (?) behavior around cascading though. Not huge issues, but: 
When I check add/edit for Authenticated, the checked permission and the read permission cascade down (as expected):
image

But, when I uncheck that add/edit, only Authenticated loses the read permission; every other named role maintains it.
image

I also can't uncheck read permissions if add or edit is selected. That feels a little anti-pattern-ish but I understand if that's intended.

wynnset:
Bug found: Checking the "Edit" or "Add" checkbox of a role row makes the "Read" checkbox cascade down incorrectly. E.g. if I select the "Edit" checkbox for "Editor", the "Read" checkboxes for all the rows below it also get incorrectly checked off:
image
Note that this does not happen if I just check the "Read" checkbox for "Editor":
image

wynnset:
Found a bug: When I press the "Edit" button, it opens the lightbox in addition to opening the node modal.

narendrasss:
> Found a bug: When I press the "Edit" button, it opens the lightbox in addition to opening the node modal.

This appears to be a problem on `master`. When I reverted it to commit `b50cea8`, before the merge commits, everything works fine. This is a pretty breaking bug so I'll work on it ASAP, but do you mind moving forward with the code review for this PR in the meantime?

narendrasss:
@wynnset The master bugs are now fixed, feel free to do the code review again.

PR #535, opened 2020-06-01
Containerize Application With Docker

### LINKED ISSUE #530: Containerize Application With Docker

liam-armstrong:
Closes #530 
Adds:
 - Docker-compose server configuration for full wordpress stack
 - Dockerfile for building vue application

Steps to test:
 1. Download and install [Docker](https://docs.docker.com/get-docker/). The installation comes with the command line tools required.
 2. Create your `wp-config.php` in the Docker folder. The only changes that need to be made from the `wp-config-sample.php` currently there is generation of the secret keys. There are instructions to do so in the file. If you want, you can also just copy the keys from your current wordpress folder.
 3. Navigate to the directory and start the stack with `docker-compose up`
 4. Load `localhost:8000` in your browser and check if wordpress is loading
 5. Create a wordpress account and activate the plugins (Should be the same plugins installed on your local wordpress)
 6. Change your hyperlink settings to post-title
 7. Create a new Tapestry and view it in the browser. The app should hook in, and you should see the loading page.


narendrasss:
It's looking good so far! I noticed a couple problems though:

* Can you add a note to the PR description saying *when* the `docker-compose` completes? The console never says that it's finished or anything which made me think the build was still going.
* I wasn't able to upload a video that I was previously able to without the docker build.
  * Specifically, I was getting this error message: `The uploaded file exceeds the upload_max_filesize directive in php.ini.`
  * If it helps, I remember in MAMP I made a config change to circumvent this. Is there any way we can change the `php.ini` file that docker generates?

PR #532, opened 2020-05-26
531 - Improve NodeModal.vue and Tapestry.vue code structures

narendrasss:
Closes #531

**Changes**

* Completely move node submission logic into `NodeModal`
* Restructure submission logic to be more concise
* Restructure `NodeModal` component so that each section of the modal belongs in its own component
  * This also removes a lot of state from `NodeModal` that was really only being used in one particular part of the form
* Update `NodeModal` API to be shorter and more readable
  * It now accepts a `node-id` and `modal-type` as props and emits `submit` and `cancel` events
* Add loading states to the modal on start and on submission
* [Fix node values not reverting when a user cancels an edit ](https://app.asana.com/0/1176098622515689/1148603786212633)

PR #537, opened 2020-06-01
536 - Show a sidebar when selecting a node

narendrasss:
Closes #536 

**Changes**

- Adds a sidebar on the side of a Tapestry that shows information on the currently selected node
- Sidebar contains the node title, description, copyright and references information
- Sidebar is responsive and should look good on all screen sizes including mobile
- Reposition node modal so that sidebar is always visible when open

**Testing**

Load a Tapestry, the sidebar should be visible on the right side and span the height of the screen.

- Click on the information icon and the sidebar should slide open with the information icon highlighted green
- Click on the arrow in the middle and the sidebar should close
- Click on the view button should open the lightbox for that node
- Click on the edit button should open the node modal with that node loaded in
  - The node modal should be positioned to the left of the sidebar and not cover it **unless your screen is less than 1280 pixels wide**

Create a node with a very long copyright and reference list (long enough that the sidebar has a scrollbar).

- Select the node, then open the sidebar
- Scroll down. When you reach the copyright title, the green highlight should move to the copyright logo
- Click on the information icon. The sidebar should scroll back up to the node title and description and the information icon should now be highlighted.

Create a node without public edit permissions.

- Open the Tapestry in incognito mode (or log out)
- Click on the node without edit permissions and open the sidebar
- The edit button on the top right corner should not be visible

Shrink your browser so it's <500px wide.

- The sidebar should cover the entire window and the icons should appear at the bottom

PR #539, opened 2020-06-01
538 - Allow author to add copyright and references for each node

### LINKED ISSUE #538: Allow author to add copyright and references for each node
For the moment only have basic textbox inputs.

narendrasss:
Closes #538 

At the moment this PR only includes textboxes for authors to copy and paste their license/reference information. I expect this will be changed soon as per the [discussion](https://app.asana.com/0/1126491658233864/1132584490660298) on Asana.

Nevertheless this version is ready for testing:

* Open up the node, there should be a "more information" tab in the modal with a copyright and reference text fields
* Write down anything in the copyright text area and submit
* Open the node modal again, the copyright field should be populated

**Note**

I made this PR to merge into #532 to make use of the node modal improvements I did on that PR. Let me know if this should be done into master instead.


liam-armstrong:
Fields save properly, good to merge

classes/class.tapestry-node.php
```diff
@@ -390,7 +400,9 @@ private function _formNode()
             'fullscreen' => $this->fullscreen,
             'conditions' => $this->conditions,
             'childOrdering' => $this->childOrdering,
-            'fitWindow' => $this->fitWindow,
+            'fitWindow'     => $this->fitWindow,
+            'license'       => $this->license,
+            'references'    => $this->references
```
wynnset, line 405: These need PHP linting

wynnset:
I'd like to hold off on merging this until we have the text formatting feature and we can use that for these 2 fields as well.

wynnset:
@narendrasss this task is no longer blocked as the text formatting PR is now merged in. Let's update it with the formatted field. I've asked Simi to provide designs on Asana for adding licenses. Let me know if you need any further clarifications. Thanks!

liam-armstrong:
Text field for the references looks good. But, the license field doesn't seem to be working as intended - the `license` field inside the node definition is still an empty string after setting it and submitting from the Modal. 
A few other comments on the UX:
 - After populating the license field and re-selecting it, the drop down options seem to be filtered with a search. This is a little weird because I expect to see all the license options every time I select it. 
 - There's a plus icon on the `Custom license` option, but pressing it, or selecting it for submission doesn't give me the option to enter a custom license. Not sure what the expected behavior is there.

narendrasss:
@liam-armstrong I think I accidentally addressed these changes in #537 instead, can you try testing that one to see if you're encountering similar problems?

liam-armstrong:
The issues are fixed in #537 , you're right. Maybe you want to move the code into this PR? or merge this PR into 536.

narendrasss:
#537 is actually going into this branch, so I'll merge the sidebar into this. Thanks for testing!

wynnset:
@narendrasss could you please bring the changes over to this PR actually? I think it's better if we merge this into master separately from #537

narendrasss:
I remember doing exactly that but looks like I never pushed! Should be good now.

PR #541, opened 2020-06-01
540 - Change external link to open in new tab by default

### LINKED ISSUE #540: Change external link to open in new tab by default

narendrasss:
Closes #540 

Note that all tapestries created prior to this PR will not have the new tab selected by default (e.g. when editing). The warning message however will show regardless.

**Testing**

* Create a new node
* Change the node type to url embed
  * The "Open in a new window" radio button should be selected by default
* Press the "Embed in Tapestry" option
  * A warning message should appear telling the user that links may not work correctly.


liam-armstrong:
Radio button is selected by default, and warning pops up if you select embed. However, the actual site doesn't pop up in a new window when I click the media button. I'm greeted with this screen: 
image Which has a link to the external site (I used https://liam-armstrong.com, hence my name on the modal)

I see you haven't made any changes to the actual new-window behavior in this PR though. Is this a separate issue? If so, the actual changes in this PR are good to go.

narendrasss:
Yes, when we mean by "new window" it means to open the preview of whatever that link will be.

liam-armstrong:
Ok, in that case this PR is ready to go.

PR #552, opened 2020-06-04
551 - Add support for "Column" H5P content

### LINKED ISSUE #551: Add support for "Column" H5P content
More specifically, scrolling for H5P iframes are disabled by default and so non-video content doesn't work.

narendrasss:
Closes #551

This PR simply changes the iframe scrolling behaviour so that scrolling is disabled only if the H5P is an Interactive Video.

Tested with the following content types:

* Interactive Video (Youtube and non-Youtube)
* Column
* True/False Question
* Timeline
* Agamotto

**Testing**

* Try various H5P content types and ensure that they show up and look presentable.


liam-armstrong:
Scroll changes look good. Every H5P content type I tested (Column, Image) that was too large for the lightbox had a scrollbar, and the videos fit perfectly.

PR #555, opened 2020-06-05
Various improvements for TYDE (as part of release)

wynnset:
- Change "Loading Tapestry" to just "Loading"
- Update help menu to include ask@tyde.ca email
- Comment out the settings button since there are none
- Reduce size of buttons a bit in the spaceship
- Remove "back to top" button in accordions
- Adjust placement of spaceship cockpit background
- Ensure vue router loads AFTER tapestry.js is loaded (was causing console errors)
- Version bump
- Tapestry.js improvements:
  - Set tapestry depth to 0
  - Remove obsolete h5p settings from tapestry.js
  - Change `jQuery` to `$` to ensure newer jQuery is used in WP
  - Load tapestry after jquery / document ready
  - Other improvements to make tapestry.js more resilient

PR #556, opened 2020-06-06
[No ticket] Fix H5P videos not playing in accordions

narendrasss:


wynnset:
I believe this was already fixed as part of #555

wynnset:
To be clear, this does not fix the issue with the videos in accordions ending early.

PR #558, opened 2020-06-08
557 - Show error and prevent submission when file upload fails

### LINKED ISSUE #557: Show error and prevent submission when file upload fails

narendrasss:
Closes #557


liam-armstrong:
Tested and successfully shows error, and doesn't update node as expected

PR #563, opened 2020-06-08
Fix tooltip sometimes covering parts of the UI

### LINKED ISSUE #562: Fix tooltip sometimes covering other parts of the UI

narendrasss:
Closes #562


templates/tapestry.css
```diff
@@ -382,6 +381,10 @@ input[type="checkbox"][disabled] {
     bottom: 8px;
 }
 
+.tooltip-wrapper {
```
wynnset, line 384: Sorry can you actually move this up in the code since it's the parent element of some of the above elements? (I believe?)

PR #565, opened 2020-06-09
Update npm packages

### LINKED ISSUE #564: some npm packages are deprecated

seanRong:
This fixes #564

### DIFF
--- a/templates/vue/config-sample.js
+++ b/templates/vue/config-sample.js
@@ -1,1 +1,0 @@
-export const LINK_PREVIEW_API_KEY = ""
--- a/templates/vue/package.json
+++ b/templates/vue/package.json
@@ -25,7 +25,8 @@
     "vue": "^2.6.10",
     "vue-router": "^3.1.3",
     "vue-slicksort": "^1.1.3",
-    "vuex": "^3.1.1"
+    "vuex": "^3.1.1",
+    "watchpack-chokidar2": "^2.0.0"
   },
   "browserslist": [
     "> 1%",
--- a/templates/vue/package.json
+++ b/templates/vue/package.json
@@ -19,6 +19,7 @@
     "babel-polyfill": "^6.26.0",
     "bootstrap": "^4.3.1",
     "bootstrap-vue": "^2.13.1",
+    "jquery": "^1.9.1",
     "moment": "^2.25.3",
     "moment-timezone": "^0.5.28",
     "vue": "^2.6.10",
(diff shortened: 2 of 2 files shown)


templates/vue/package.json
```diff
@@ -19,12 +19,14 @@
     "babel-polyfill": "^6.26.0",
     "bootstrap": "^4.3.1",
     "bootstrap-vue": "^2.13.1",
+    "jquery": "^1.9.1",
```
narendrasss, line 22: Any reason why this was installed? We don't use jQuery in the vue build.

templates/vue/package.json
```diff
@@ -19,12 +19,14 @@
...
     "moment": "^2.25.3",
     "moment-timezone": "^0.5.28",
     "vue": "^2.6.10",
     "vue-router": "^3.1.3",
     "vue-slicksort": "^1.1.3",
-    "vuex": "^3.1.1"
+    "vuex": "^3.1.1",
+    "watchpack-chokidar2": "^2.0.0"
```
narendrasss, line 29: What is this package for?

templates/vue/config-sample.js
```diff
@@ -1 +0,0 @@
-export const LINK_PREVIEW_API_KEY = ""
```
narendrasss, line 1: Please keep this file. We need it for new users so they know what the `config.js` file should look like. It's also unrelated to the deprecated problem (to my understanding). We want to keep the PRs as contained to their original issue as possible.

templates/vue/package.json
```diff
@@ -41,7 +43,7 @@
     "cross-env": "^5.0.5",
     "css-loader": "^0.28.7",
     "cypress": "^3.8.0",
-    "eslint": "^6.5.1",
+    "eslint": "^5.16.0",
```
narendrasss, line 46: This looks like we're downgrading to me. Could you try updating this to `7.1.0` and see if that works? We have a pending PR at the moment upgrading the eslint version to that version.

templates/vue/package.json
seanRong: I believe this came in when I attempted to update fsevents, which I couldn't/ didn't need to update anyway.

templates/vue/package.json
seanRong: Will remove

narendrasss:
Closing as not needed as per our discussion.

PR #570, opened 2020-06-11
566 - Submit button should be disabled while any file is uploading

seanRong:
Display a spinner in settings and node modals and disable submit button when a file is in the process of being uploaded via computed property.


liam-armstrong:
Tested and the functionality is working well!

templates/vue/src/components/FileUpload.vue
```diff
@@ -101,6 +101,7 @@ export default {
           },
           onUploadProgress: progressEvent => {
             this.isUploading = true
+            this.$emit("isUploading", this.isUploading)
```
narendrasss, line 104: This is good! You got the hang of it fast.

templates/vue/src/components/FileUpload.vue
```diff
@@ -113,6 +114,7 @@ export default {
         })
         .then(response => {
           this.$emit("input", response.data.data.url)
+          this.$emit("isUploading", false)
```
narendrasss, line 117: I think you should wrap this in a `.finally(...)` instead, since in the event of an error the `then` block wouldn't be called.

templates/vue/src/components/SettingsModal.vue
```diff
@@ -112,10 +120,14 @@ export default {
...
       showAccess: true,
       defaultPermissions,
+      fileUploading: false,
     }
   },
   computed: {
     ...mapGetters(["settings", "tapestryJson"]),
+    accessSave() {
+      return !this.fileUploading
+    },
```
narendrasss, line 130: I don't think this extra computed property is necessary, you can just reference `fileUploading` directly in the template.

templates/vue/src/components/FileUpload.vue
seanRong: We don't catch any errors at all in this part either - is that a problem?

seanRong: Also, would it be sufficient to set isUploading to false before the reset hook during the finally?

templates/vue/src/components/SettingsModal.vue
seanRong: I might have to read vue docs for more understanding on this, although I agree its redundant

templates/vue/src/components/FileUpload.vue
narendrasss: > We don't catch any errors at all in this part either - is that a problem?

The error part was fixed in another PR that's awaiting review at the moment. 

> Also, would it be sufficient to set isUploading to false before the reset hook during the finally?

Yes I think so.

liam-armstrong:
Tested and working!

wynnset:
@seanRong could you please fix the merge conflicts?

wynnset:
@seanRong this still has some merge conflicts sorry

seanRong:
@liam-armstrong the internal implementation for video and H5P seem to have changed, and the infinite loading is happening again. I tried using what you had last time to patch it but it hasn't worked :/

templates/vue/src/components/NodeModal.vue
```diff
@@ -212,11 +221,20 @@ export default {
         ? true
         : wpData.wpCanEditTapestry !== ""
     },
+    accessSubmit() {
+      if (!this.ready || this.fileUploading) {
+        return false
+      }
+      return this.node.mediaType === "video" || this.node.mediaType === "h5p"
+    },
```
liam-armstrong, line 229: When I open the modal, the submit button is inaccessible by default; this computed property is returning false.

PR #572, opened 2020-06-12
TYDE: 531 - Improve NodeModal.vue and Tapestry.vue code structures

narendrasss:


liam-armstrong:
Root node submission fails for a new Tapestry with this error:
image
Modal's loading spinner continues forever while this occurs

Also, small thing to note was the the Ordering tab was available even though the node had no children, and the tab was empty

wynnset:
Please fix merge conflict

PR #575, opened 2020-06-13
TYDE: Fix Import Statment Causing Docker Compliation Failure

liam-armstrong:
Docker containers are Linux based, which uses a case-sensitive file system. Both OSX and Windows are case insensitive, which is why this statement was loading fine on WAMP/MAMP.

Steps to test:
- Run `docker-compose up` and verify the `app` container compiles without issue


liam-armstrong:
Marking this as ready for review since @codygagnon tested the same changed in PR #576

PR #581, opened 2020-06-15
579 - Containerize Modular Cypress Tests

### LINKED ISSUE #579: Containerize Cypress

liam-armstrong:
Closes #579
This PR should allow us to run out Cypress tests from within a Docker container, and hopefully reduce the inconsistencies we're facing across our local environments

To run the tests, start the docker containers as usual with:
`docker-compose up -d`
This will start the containers in detached mode, so once they start you will be returned to the regular console.
While the containers are up, start the tests with:
`docker exec e2e cypress run --headless --browser chrome`
Finally, you can stop the containers with:
`docker-compose down`


liam-armstrong:
Tested and it runs well! A little slow, but I don't believe that can be avoided. @codygagnon and @seanRong, give this a quick test and reply here if everything's in order!

seanRong:
My runs with cypress on docker:

For author spec, most of the times tests are failing due to timing out after being unable to find `#node-modal-container` (kind of strange imo, maybe the element does not load in time, or perhaps a GET request authentication issue?). This might also be because of my laptop's speed or the fact I can't run docker if I try to allocate >1GB of RAM for it. I attached some photos of the failed runs

The user spec tests are actually usually passing, but the furthest I got was the 4th test in the user spec before it hung indefinitely.
Each test usually takes upwards of half a minute to process. 
[authoring.spec.js.zip](https://github.com/wynnset/tapestry-wp/files/4928405/authoring.spec.js.zip)

codygagnon:
0/20 tests in authoring.spec.js pass on my machine. All tests output: "AssertionError: Timed out retrying: Expected to find element: `#node-modal-container`, but never found it." Duration of authoring.spec.js was 4 minutes, 34 seconds

2 tests in user.spec.js pass (Should be able to view nodes (14147ms) and Should be able to click on media icon to open lightbox (7687ms)). It got hung up after the first two tests.

PR #591, opened 2020-06-18
TYDE: 586 - Refactor tyde and master branches to minimize conflicts

narendrasss:
Changes:

* Move all TYDE related content to the bottom of their respective sections as much as possible
* Move TYDE related styles into a single stylesheet that is imported by Tyde.vue
* Lint files using common ruleset


wynnset:
@narendrasss can you please fix the merge conflicts when you get a chance?

templates/tapestry.js
```diff
@@ -2513,7 +2510,7 @@ function tapestryTool(config){
                     node.typeData = content.typeData
                 }
 
-                if (node.mediaType !== "accordion" && tapestry.dataset.nodes[index].typeData.progress) {
```
wynnset, line 2516: We should keep this condition as it has caused some issues in the past because it was not set. We can change it to `&& node.typeData.progress` I guess. We should make sure this is on master as well

templates/vue/src/components/AudioRecorder.vue
```diff
@@ -259,9 +258,5 @@ export default {
   i {
     margin: auto;
   }
-
-  .teen-status {
-    display: block;
-  }
```
wynnset, line 265: Just double-checking... Should this be deleted? It's not used?

templates/vue/src/components/ModalButton.vue
```diff
@@ -81,15 +81,17 @@ export default {
   border-radius: 50%;
   width: 40px;
   height: 40px;
-  font-size: 18px;
```
wynnset, line 84: I think this is supposed to stay as `px` because it is a button with an icon. Icon-only buttons should not generally use `em` since they are not really text.

templates/vue/src/components/lightbox/AccordionMedia.vue
```diff
@@ -469,44 +403,27 @@ button[disabled] {
...
-  margin: 0;
-  padding: 8px 36px;
-  color: #026c93;
-  font-weight: bold;
-  font-size: 1.2em;
-}
-
-.sub-accordion-text {
-  margin-bottom: 0;
+.button-scroll-top {
```
wynnset, line 406: The scroll to top button has been removed and should remain removed.

templates/vue/src/components/AudioRecorder.vue
narendrasss: It's moved to `Tyde.scss`.

templates/vue/src/components/ModalButton.vue
narendrasss: I was keeping this consistent with master that's why it was changed. I personally think it's ok to keep it as `em`. `em` is not (and I don't think should be) limited to text only. I used it extensively in the sidebar PR so everything stays consistent and responsive.

PR #596, opened 2020-06-22
595 - Tooltips at the Top of a Tapestry Get Hidden

### LINKED ISSUE #595: Tooltips at the top of a tapestry get hidden

liam-armstrong:
Closes #595 
Generates tooltip location based on node coordiates

Steps to test: 
- Make a locked node
- Drag it to the top of the tapestry page
- Hover over it and make sure tooltip renders under node rather than on top
- Drag away and check that it resets


seanRong:
Doesn't work on my computer. Even if the tooltip would display out of window view, it still is displayed at the top. 
Maybe use browser window as a reference instead of tapestry dimensions?

liam-armstrong:
> Doesn't work on my computer. Even if the tooltip would display out of window view, it still is displayed at the top.
> Maybe use browser window as a reference instead of tapestry dimensions?

@seanRong 
Which WordPress theme are you using? And what are your browser dimensions

seanRong:
I'm using a standard WP theme, called "Twenty Nineteen".

view width = 1920
view height = 937

narendrasss:
Works great on my machine!

liam-armstrong:
@seanRong & @codygagnon - I'm assigning this to you since it's not working on either of your machines. If the node is too close to the top of the Tapestry, it should render underneath the node rather than above. The calculation for whether or not the node is too close to the top boils down to: `d.y - tapestry.getTapestryDimensions().startY < getRadius(d) * 2.5`.

If one of you could take a shot at finding out why it's not working on your machines and fixing it, that would be great. Let me know if you have any questions!

seanRong:
Tested it again, got it to work as intended 

But here's the thing: it only considers tapestry dimensions. Shouldn't we consider browser view instead? Otherwise this still happens:

pr595

narendrasss:
Hey @liam-armstrong, I think the implementation should be updated according to @seanRong's recommendation above.

liam-armstrong:
It's a pretty specific edge case, but I've added support for browser scroll. Please test again @seanRong and @narendrasss , and code review / mark this PR as reviewed so we can get it merged.

narendrasss:
The browser scroll change doesn't work for me still:

Screen Shot 2020-07-15 at 13 55 45

templates/tapestry.js
```diff
@@ -1349,9 +1351,20 @@ function tapestryTool(config){
...
+                const onBottom = (d.y - tapestry.getTapestryDimensions().startY < tooltipHeight 
+                    || $(window).scrollTop() > tooltipHeight) && d.x > 0 && d.y > 0
+                return onBottom ? getRadius(d) + 27.5 + 5 : -(getRadius(d) * 3 + 27.5 + 20)
+            })
             .append("xhtml:div")
             .attr("class", "tapestry-tooltip")
+            .style("align-items", d => {
+                const tooltipHeight = getRadius(d) * 2.5
+                const onBottom = (d.y - tapestry.getTapestryDimensions().startY < tooltipHeight 
+                    || $(window).scrollTop() > tooltipHeight) && d.x > 0 && d.y > 0
```
seanRong, line 1365: I think here, we should use the scrollTop of the tooltip element itself, not the scrollTop of window (which isn't really relevant)

liam-armstrong: Since we're using the y position of the node in the window, the $(window).scrollTop() is the value we need to compare against the position to determine where the tooltip should render. My testing for the tooltip's scrollTop value was always 0, because you don't scroll within that element

codygagnon:
Doesn't work on my machine.

image

narendrasss:
@seanRong I think you have to approve this PR before I can go ahead and merge it in.

PR #597, opened 2020-06-23
583 - Switch all textareas to allow formatting options and links

seanRong:
This upgrades all existing text areas to accept formatted text, lists, and edit history, with options of:
 ```

        bold,
        italic,
        underline,
        strike,
        code,
        paragraph,
        bullet list,
        numbered list,
        quote,
        horizontal break,
        undo,
        redo,
     
```

All previous saving and display mechanisms are maintained. 

tiptap is added as a dependency

### CODE REVIEW

**Comment by narendrasss** [comment]
Hey @seanRong for future PRs can you link the related issue to the PR? This makes sure that the related issue gets closed when the PR gets merged. I've done it for this one. You can do this either through [closing words](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) in the PR description or under "linked issues" on the bottom right.

**Comment by seanRong** [comment]
I think before styling the icons, we could decide which of the functionalities we actually need?

**Comment by wynnset** [comment]
Most importantly bold, italic, strikethrough, underline, bulleted list, and ordered list, plus link:
image

Some other things if we have space (these should be hidden if the screen is not wide enough to show them):
H1, H2, H3, quote, code, undo, redo

H1, H2, H3 should really be h2, h3, h4, but we will label them as H1, H2, H3. This is because the node title should be the only H1.

No need for a specific button for paragraph. We don't need anything more than that for the time being.

Thanks!

**Comment by seanRong** [comment]
Let me know any thoughts you all have on the aesthetics too please!

File: templates/vue/src/components/node-modal/content-form/RichTextForm.vue [hunk 1]
```diff
@@ -0,0 +1,273 @@
...
+        <button class="menubar__button" @click="commands.horizontal_rule">
+          <icon name="hr" />
+        </button>
+        <button class="menubar__button" @click="commands.undo">
+          <icon name="undo" />
+        </button>
+        <button class="menubar__button" @click="commands.redo">
+          <icon name="redo" />
+        </button>
+      </div>
```
narendrasss, line 98: Could this be converted to a `v-for` on a `computed` property?

seanRong: Some of them but not all because there's slightly different structure for h1 h2 h3 and hr undo redo

templates/vue/src/components/node-modal/content-form/Icon.vue
```diff
@@ -0,0 +1,94 @@
...
+        "ol",
+        "quote",
+        "hr",
+        "undo",
+        "redo",
+      ]
+      for (let i in icons) {
+        lookup.set(iconsStrings[i], icons[i])
+      }
+      let usedIcon = lookup.get(this.name)
```
narendrasss, line 77: I notice the icons and the strings are all the same, so we can just do:

```suggestion
      let usedIcon = this.name
```

templates/vue/src/components/node-modal/content-form/RichTextForm.vue
narendrasss: At the least I think we can have all the similarly structured ones on a `v-for` and have everything else come either afterwards or before.

templates/vue/src/components/node-modal/content-form/Icon.vue
seanRong: in the icon array, those aren't strings, those are references to strings with the full url as provided by webpack (with the ./dist..... and stuff), which is why the map is used to translate between the two.

narendrasss: Ah good point, nevertheless I think we can shorten this to something like:

```
const lookup = {
  bold,
  italic,
  underline,
  ...etc
}
const usedIcon = lookup[this.name]
```

seanRong: Does it really work without the Map?

narendrasss: Yeah, javascript objects are functionally identical to Maps if the keys are strings or numbers

seanRong: Still confused

`const lookup = {
  bold,
  italic,
  underline,
  ...etc
}
const usedIcon = lookup[this.name]
`

Will this not become 

`const lookup = {
  wpaddress/src/dist/bold7098709871234.svg,
  wpaddress/src/dist/italic7098709871234.svg,
  wpaddress/src/dist/underline7098709871234.svg,
  ...etc
}
const usedIcon = lookup[this.name]`

after compilation?

narendrasss: No it's going to compile to:

```
const lookup = {
  bold: "wpaddress/src/dist/bold7098709871234.svg",
  italic: ".....",
  etc
}
```

It's called "object shorthand notation": https://alligator.io/js/object-property-shorthand-es6/

seanRong:
Back to testing. 

Does H1/H2/H3 or the horizontal rule icon look weird to anybody else?

templates/vue/src/components/node-modal/content-form/RichTextForm.vue
```diff
@@ -0,0 +1,288 @@
...
+          name: "ol",
+          text: "",
+        },
+        {
+          vclass: "{ 'is-active': isActive.blockquote() }",
+          click: "commands.blockquote",
+          name: "quote",
+          text: "",
+        },
+      ],
```
liam-armstrong, line 151: All these vclass elements are being placed, as text, right into the element's class tag. I'm not sure this is what you intended, because it looks like you want this JavaScript to evaluate somewhere. What's this supposed to do?

seanRong: Solved via reverting template to not use v-for

templates/vue/src/components/node-modal/content-form/RichTextForm.vue
seanRong: Was implemented, then un-implemented due to being non-functional

templates/vue/src/components/node-modal/content-form/Icon.vue
seanRong: Added!

wynnset:
Sorry this still needs work. Most importantly, I don't see a way to insert a hyperlink, which was one of the stated requirements.

Also:
- The H1 that was being used before in text content needs to be reset to its original size now that we have these additional heading sizes since the new H1 (coded as H2) is now bigger than the title for the text node. I believe you just need to delete the SCSS for `article h1` that we had custom-created for the text content type.
- What is this icon? 
image
- Can we please put a gap before the H1 and after H3 to create some perceived groupings?
- Can we please put a very light background on the toolbar (e.g. `#eee`)

PR #599, opened 2020-06-24
594 - Allow creation of tapestry using a shortcode / button / widget

### LINKED ISSUE #594: Allow creation of tapestry using a shortcode / button / widget
Currently, the user experience to add a new tapestry is quite ugly. They have to go to the Wordpress backend, press Tapestry > Add New and then enter the title, then publish and view the tapestry.

Instead of this, let's add a simple way so that users can click on a button on the frontend (in the form of a Wordpress shortcode) to Add a new Tapestry. This button should ask them for a title and then immediately add the Tapestry and take them to their new Tapestry.

We also want to make this button available as a widget.

codygagnon:
Closes #594 

Adds the ability to create a button on any post via shortcode` new_tapestry_button`.

Clicking the button will prompt for a new tapestry name; if tapestry is created successfully, it will redirect to the new page of the tapestry; otherwise, will alert with an error.

This also changes the add tapestry endpoint object to have a new "permalink" property under the settings object.

HOW TO TEST:
Make a new post with any title, and in the body write `[new_tapestry_button]`. Navigate to the post, and a button should appear to add a new tapestry.


seanRong:
Is there a way to infer the necessary URL prefixes for the endpoint and the created tapestry?

wynnset:
> Is there a way to infer the necessary URL prefixes for the endpoint and the created tapestry?

Yes: `<?php echo get_rest_url(null, 'tapestry-tool/v1'); ?>` 
We are doing that in `single-tapestry.php` to get the API endpoint and using it in `TapestryAPI.js` as well.

The endpoint should return the ID of the newly created tapestry post and you should use that to retrieve the permalink for it using the normal Wordpress way: https://developer.wordpress.org/reference/functions/get_permalink/

seanRong:
Updated description to describe what has changed

liam-armstrong:
Not sure if this is ready for testing or not, but I tried to load it up and wasn't able to create the button with the shortcode. Anything I'm missing here?

image

seanRong:
Ah shoot it should actually be 

`[new_tapestry_button]`, and yes it should be ready for testing

codygagnon:
I get a compile error:

./src/components/node-modal/AppearanceForm.vue
Module build failed: Error: ENOENT: no such file or directory, open '/app/src/components/node-modal/AppearanceForm.vue'

I don't know why, because I can see the file at this path and it is spelled and capitalized correctly.

seanRong:
I think its fine to just comment it out @cody because that doesn't really affect the shortcode

codygagnon:
I wrote the shortcode in the body of a new post, but there is no new tapestry button.

image

seanRong:
As discussed it is ok if it doesn't work on tapestries.

narendrasss:
@seanRong I moved the permalink settings initialization to `class.tapestry.php` because all of the settings were moved out of `tapestry.php` a few commits back.

PR #611, opened 2020-06-29
29-06-2020 - Merge master into tyde 

narendrasss:
Includes the following additions from master:

* Change linter output to include what errors were found
* Fixes node selection interfering with the depth slider
* Change external links to open in new tab by default
* Synchronize read permissions with add/edit
* Add support for non-video H5P content
* Show error message if file upload fails


liam-armstrong:
image

- Link's aren't rendering (node with thumbnail is a module in screenshot above ^) 
- Nodes starting as locked bug is appearing here, which I believe is expected as the bug isn't fixed yet
- Trying to launch a module fails, returning just a white screen with the following info in the console: 
image

narendrasss:
@liam-armstrong Thanks for testing, should be fixed now. Can you take a look?

PR #618, opened 2020-06-30
612 - Add progress support for non-logged-in users

### LINKED ISSUE #612: Progress Not Saved If User Is Not Logged In

narendrasss:
Closes #612 

Changes:

* Save user progress to `localStorage` if they're not logged in
* If a node's tooltip is not time-based, add a warning that the user must login to unlock the node
* Add a warning to activities telling users to login if they wish to have their answers saved


seanRong:
What I had on my setup (WAMP):

- Nothing to do with progress was in my localStorage
- Voice recorder activity would spin endlessly after submitting voice recording (auth issue?)
- After finishing a H5P quiz, would not let me progress further

narendrasss:
@seanRong When you say you can't "progress further", do you mean you can't continue to the next question or does the node not get marked as completed?

seanRong:
Does not mark as complete.

also text only should mark complete after viewing right? It doesn't work either.

templates/vue/src/store/actions.js
```diff
@@ -71,6 +80,13 @@ export async function updateUserProgress() {
...
 export async function completeNode({ commit, dispatch, getters }, nodeId) {
+  if (!wpData.wpUserId) {
+    const progressObj = JSON.parse(localStorage.getItem(LOCAL_PROGRESS_ID))
+    const nodeProgress = progressObj[nodeId] || {}
+    nodeProgress.completed = true
+    localStorage.setItem(LOCAL_PROGRESS_ID, JSON.stringify(progressObj))
+  }
+
   await client.completeNode(nodeId)
   commit("updateNode", {
```
liam-armstrong, line 91: Small suggestion: Shouldn't we only make this API call if the user is logged in? Seeing as we know it's not doing anything.

liam-armstrong:
Other than the single potential change above and the merge conflicts, this functions beautifully and is ready to get merged once those are addressed

templates/vue/src/store/actions.js
narendrasss: Good point, I notice there are a few other API calls that might be redundant as well. Thanks for pointing it out!

narendrasss: Good point, I notice there are a few other API calls that might be redundant as well. Thanks for pointing it out!

wynnset:
@narendrasss this looks awesome for the most part. Just wanted to chat about some potential consequences and edge cases of this soon. For example, if a user is logged out and completes a bit of the tapestry, then logs in, it would be nice if we can transfer their progress onto the server (maybe ask them first). This is not super necessary but would be a nice addition. I also have not tested this, but saw you removed some stuff about saving H5P settings and still need to make sure I test that and a few other things (this is more of a note for myself haha). Thanks.

seanRong:
Progress is being tracked for images, videos. 

H5P is always stuck at no progress (no blue ring at all). In fact, it doesn't even work when not logged in. Issue just on my end? I made a basic quiz with 1 question, with two MC and 1 correct answer. After clicking finish I thought it would give me the progress ring but it doesn't happen.

seanRong:
As discussed some types of nodes should not be supported for progress tracking if not logged in.

narendrasss:
@liam-armstrong Can you review this again so we can get it merged in?

PR #621, opened 2020-07-03
600 - Bug fix: New nodes are locked by default

codygagnon:
- Added unlocked and accessible properties to default node
- Modified submit function to evaluate node conditions on submission

### DIFF
--- a/templates/vue/src/components/NodeModal.vue
+++ b/templates/vue/src/components/NodeModal.vue
@@ -234,7 +234,13 @@ export default {
   },
   methods: {
     ...mapMutations(["updateOrdering", "updateSelectedNode", "updateRootNode"]),
-    ...mapActions(["addNode", "addLink", "updateNode", "updateNodePermissions"]),
+    ...mapActions([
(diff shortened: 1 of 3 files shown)


templates/vue/src/components/NodeModal.vue
```diff
@@ -284,6 +284,11 @@ export default {
           }
         }
 
+        if (this.node.conditions.length > 0) {
```
wynnset, line 287: Conditions existing does not mean that a node will be locked. The conditions would first have to be evaluated to determine if the node will be locked or not. Since this evaluation is done in the backend, you might have to send the request to the backend to do the evaluation and then receive the result of that to update the accessible and unlocked properties here.

wynnset:
image

As an example to my review, here I set a condition where a date has to be passed. The date is already past, but because the condition is not being evaluated, the node is getting locked incorrectly. This could exist for any of the conditions and would need to be evaluated in a central place (backend).

wynnset:
Excellent!

PR #624, opened 2020-07-03
623 - Extend Vue Router

### LINKED ISSUE #623: Extend Capabilities of Vue Router

liam-armstrong:
Closes #623 

This PR updates Tapestry to be predominantly URL-based. Specifically, it adds the following routes:

**Nodes**

- `/nodes/:id` - Sets `id` to be the currently selected node
- `/nodes/:id/add` - Opens the add node modal with id `id`
- `/nodes/:id/edit` - Opens the edit node modal with id `id`
- `/nodes/:id/<add|tab>/:tab` - Opens the add/edit modal with the tab open
- `/nodes/:id/view` - Opens the node's lightbox
- `/nodes/:id/view/:row_id` - Opens the `row_id` row in the lightbox (if the node is multi-content)
- `/nodes/:id/view/:row_id/row/:sub_row_id` - Opens the sub accordion row with id `sub_row_id`

**Settings**

- `/nodes/:id/settings/:tab` - Open the settings modal with the corresponding tab

When any node ID doesn't exist or the route is invalid in any way, this PR makes a best guess as to where it should redirect:

* Nested lightbox routes will redirect to the lightbox route, and
* Other invalid routes will redirect to the `/nodes/:id` route where `id` is the `rootId` of the dataset


templates/vue/src/components/NodeModal.vue
```diff
@@ -206,35 +218,47 @@ export default {
...
+      },
+    },
+  },
+  watch: {
+    $route(to, from) {
+      if (this.currentTab !== to.params.tab && to !== from) {
+        this.currentTab = to.params.tab
+      }
+    },
+  },
```
narendrasss, line 243: I think rather than doing this we should read the active tab directly from the current route. We can do this through the `active` prop in `b-tab` and by listening to the `@click` event also emitted by `b-tab`. This makes it so that the single source of truth is the current route, and we don't have to manage state inside `NodeModal`.

This also means that we have to link directly to the tab that we want to go to, e.g. pressing the add button would link to `/add/1234/content` instead of just `/add/1234`.

The docs for the tab component could be a good starting point for this: https://bootstrap-vue.org/docs/components/tabs

I hope that makes sense! If not, DM me and we can talk about it more.

liam-armstrong: I think that you make a really good point here, so I removed the parts where we keep track of the `currentTab` and deal with storing that state, and instead refer directly to the route. Take a look at the changes to the NodeModal file and let me know if that's what you had in mind

liam-armstrong:
@narendrasss I addressed your concerns about the subaccordions in the last few commits. Seeing as it was your code originally, and my solution is (necessarily) a little messy, I was hoping you could take a look at it as well. Otherwise though, this PR is back to being Ready for Testing

templates/vue/src/components/TapestryAccordion.vue
```diff
@@ -6,45 +6,53 @@ export default {
...
     },
-  },
-  created() {
-    if (this.defaultIndex >= 0) {
-      const row = this.rows[this.defaultIndex]
-      if (row) {
-        this.activeRow = row
-      }
-    }
+    activeRowIndex() {
```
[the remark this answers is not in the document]
wynnset: I think it would be OK to see the route changes there as well, as long as we also see the route changes for everything else in the spaceship cockpit so that the URL can take the user to a specific module / subtopic / accordion row in the cockpit. It's OK with me as long as it behaves in a way that the user doesn't get surprised and that it's consistent, so I'll leave it up to you to make that decision.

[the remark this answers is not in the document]
liam-armstrong: I've added a prop to the component to allow for the route changes to remain constant on TYDE as well. When we port the feature, it should behave as expected which I believe is OK.

codygagnon:
1. Sometimes when I refresh with the node modal open and then close the node modal, the tapestry does not appear. This has happened several times, but it does not happen every time, so I have not been able to get a video of it yet.
1. Should there be one URL for the Content tab? Currently, the Content tab could be /nodes/add/851 or /nodes/add/851/content

liam-armstrong:
>     1. Sometimes when I refresh with the node modal open and then close the node modal, the tapestry does not appear. This has happened several times, but it does not happen every time, so I have not been able to get a video of it yet.

You're right it does! Fixed.

>     2. Should there be one URL for the Content tab? Currently, the Content tab could be /nodes/add/851 or /nodes/add/851/content

Yes, both URLs go to the content tab. I was thinking each tab should have a specific URL, even the default content tab. But, if a user was trying to figure out the link to the node modal based on the other patterns, they would guess it as blank. Easy to change, but I think it's fine as is.

seanRong:
If you enter a url for edit with a node id that doesn't exist, it causes infinite loading.

If you enter a url for view with a node id that doesn't exist, it causes the screen to zoom out massively. 

I think in both these cases, we should default to rendering the tapestry as if there was no edit or view suffix.

wynnset:
@narendrasss would you mind fixing the merge conflicts here and then marking this as ready for review, please?

narendrasss:
@wynnset Should I still add those new features and rebase it on top of 482 as we discussed?

seanRong:
When I tested the standard behaviour yesterday it worked really well.

I tried some random edge cases today for the fallbacks, the nested accordion stuff works great, for the root node it's fine if it's `/view` but when I did `/add` it actually nuked my entire tapestry...

narendrasss:
@seanRong Thanks! I added a guard that prevents navigating to `/nodes/:nodeId/<tab>` if the `nodeId` doesn't exist in the Tapestry.

wynnset:
@narendrasss can we do a screenshare and rebase this on the new master using your script, please?

narendrasss:
@wynnset This particular PR bounced around between me and Liam so my script would've been a bit complicated I think. I ended up doing a manual merge through GitHub, but I haven't tested it yet.

wynnset (changes requested):
Could we also add a couple of things using the URL parameters after a question mark `?`:
1. Depth parameter (unless it's the default depth)
2. Author filter parameter (unless there is no filter)

e.g. `#/nodes/123?depth=4&author=2` or `#/nodes/123/edit?depth=3` or `#/nodes/123/view?author=2`

This page might help: [https://reactgo.com/vue-router-query-params/](https://reactgo.com/vue-router-query-params/)

I know we are also changing the author filter in another PR, so if you want, we can address it in that PR instead. But the depth needs to be implemented here if possible.

templates/vue/src/config/routes.js
```diff
@@ -0,0 +1,100 @@
...
+  lightbox,
+  modal,
+  settings,
+  subAccordion,
+  redirects,
+}
+
+export const names = Object.fromEntries(
+  Object.entries(routes).map(([name, route]) => [name.toUpperCase(), route.name])
+)
```
wynnset, line 98: 👌🏼

narendrasss:
>  If an invalid URL is entered, we should handle it gracefully (delete the invalid part or something)

I believe this is already the case, unless there was a specific case that you tested that it didn't do so.

wynnset:
> > If an invalid URL is entered, we should handle it gracefully (delete the invalid part or something)
> 
> I believe this is already the case, unless there was a specific case that you tested that it didn't do so.

Yes, I entered an invalid node ID and the nodes extended out of the svg when the page loaded: 
image

wynnset (approved):

wynnset:
Everything looks good now, minus the issue mentioned above.

jaskwong:
1. If user without editing permission goes to /nodes/:id/edit or /nodes/:id/add, it opens the edit/add node modal, (but user cannot actually submit an edit, leads to infinite buffering)

2. Navigating first to edit and then deleting the "edit/content" at the end of the url and replacing it with "view" will redirect to just the node selected (unexpected behaviour)
(ex. replacing the "edit/content" in http://localhost/tapestry/tapestry-1/#/nodes/108/edit/content with "view" to http://localhost/tapestry/tapestry-1/#/nodes/108/view will just redirect to http://localhost/tapestry/tapestry-1/#/nodes/108)

narendrasss:
Thanks @jaskwong! I totally missed the permissions case.

seanRong:
I've noticed lately that a "depth=n" suffix now exists on the URL.
When I delete a node, this suffix is lost and I get sent back to a depth level that's equal to the default depth (root node also gets reset), which could cause user to lose track of where in the tapestry they were looking. Funny enough if you had copied that URL and enter the URL into navbar again you'll go to the location/zoom level where the deleted node was.

Relevant console err: ```Error: Redirected when going from "/nodes/17/edit/content?depth=7" to "/" via a navigation guard```

wynnset (approved):

wynnset:
Bug: If I'm not logged in, I keep getting redirected to the root node at default depth level, no matter what I click on. If I manually enter the URL, I also get redirected to the same place.

narendrasss:
Thanks, good catch. Should be fixed now (along with a few console error fixes).

wynnset (approved):

wynnset:
Hallelujah 🎉

PR #630, opened 2020-07-07
628 - Enable hiding nodes from Admin and implement Author filter in backend

### LINKED ISSUE #628: Implement Filter in Backend

liam-armstrong:
Closes #628 

Changes: 
 - Add new `superuserOverridesPermission` flag to settings. This is true by default.
 - Add `userId` parameter to `getTapestry` endpoint. If specified, this second user's authored nodes are returned with the current user's, overriding read permissions.
 - Refactored `tapestry-user-roles.php` to no longer be static, and to take a userId argument. Changed all calls to it's functions in the backend to work as objects calls.
- Added `getAllContributors` endpoint to `tapestry.php`. This returns a list of unique authors of nodes within the tapestry.

Steps to test:
1. Login as your tapestry author/editor.
2. Create a node with public add/edit permissions and another with non-public read access.
3. Log out, and log in as another user which is not a super user. 
Superusers are either: Author of the Tapestry, Administator role on WP or Editor role on WP
4. As this non-superuser, you should only be able to see the node with all public permissions.
5. Create a new node as the non-superuser. Turn it's public read permissions off, so only that user can view it.
6. Log back into the admin account and verify the superuser can see all 3 nodes in the tapestry
7. Go to settings and flip the `superuserOverridesPermission` flag to false
8. After reload, verify the superuser can no longer see the 2nd account's node. Check the backend and ensure the visible nodes are the only ones within the dataset.
9. Activate the filter, and select the 2nd user's node as filtered
10. Ensure the 2nd users nodes appear along with the 1st.
11. Deactivate the filter and ensure the nodes go back to the original state. Check the backend and ensure the visible nodes are the only ones within the dataset.

### DIFF
--- a/templates/vue/src/components/SettingsModal.vue
+++ b/templates/vue/src/components/SettingsModal.vue
@@ -99,6 +99,14 @@ const defaultPermissions = Object.fromEntries(
   ].map(rowName => [rowName, ["read"]])
 )
 
+const defaultSettings = {
+  backgroundUrl: "",
+  autoLayout: false,
+  nodeDraggable: true,
+  showAccess: true,
+  superuserOverridePermissions: true,
+}
+
 export default {
   name: "settings-modal",
   components: {
@@ -142,6 +150,7 @@ export default {
     openModal() {
       this.$bvModal.show("settings-modal")
       this.getSettings()
+      this.synchronizeSettings()
     },
     closeModal() {
       this.$bvModal.hide("settings-modal")
@@ -192,6 +201,14 @@ export default {
       URL.revokeObjectURL(fileUrl)
       document.body.removeChild(a)
     },
+    synchronizeSettings() {
+      const tapestrySettings = this.settings
+      for (const setting in defaultSettings) {
+        if (!tapestrySettings.hasOwnProperty(setting)) {
+          this.updateSettings()
+        }
+      }
+    }
   },
 }
 </script>
--- a/templates/vue/src/components/SettingsModal.vue
+++ b/templates/vue/src/components/SettingsModal.vue
@@ -35,6 +35,15 @@
               {{ autoLayout ? "Enabled" : "Disabled" }}
             </b-form-checkbox>
           </b-form-group>
+          <b-form-group
+            label="Superusers Override Permissions"
+            description="When this is enabled, all WordPress superusers (Administator roles, Editor roles or the
+             Author of this Tapestry) will be able to see and edit all nodes, regardless of the node's set permissions."
+          >
+            <b-form-checkbox v-model="superuserOverridePermissions" switch>
+              {{ superuserOverridePermissions ? "Enabled" : "Disabled" }}
+            </b-form-checkbox>
+          </b-form-group>
         </b-tab>
         <b-tab title="Advanced">
           <b-button block variant="light" @click="exportTapestry">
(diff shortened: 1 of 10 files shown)


wynnset:
Yes, this is working as expected for the most part, thanks! Just a couple of things:

- [ ] I noticed that when I first loaded an existing tapestry and went into the setting, the "Superusers Override Permissions" toggle was on, yet I could not see any private nodes added by my other users. Once I turned it off, then saved, and then turned it on again and saved, the node showed up.
- [ ] The author filter does not show me the users who have contributed to the tapestry. As mentioned in the Asana task description, they should always show up for superusers and selecting them should add their nodes and highlight it in the current tapestry.

templates/vue/src/components/TapestryFilter.vue
```diff
@@ -79,25 +85,42 @@ export default {
     comboboxValueOptions() {
       switch (this.activeFilterOption) {
         case filterOptions.AUTHOR: {
-          const authors = new Map(
-            this.nodes.map(node => [node.author.id, node.author])
-          )
-          return [...authors.values()]
+          // console.log(this.nodes.map(node => [node.author.id, node.author]))
```
codygagnon, line 88: Should this console log be removed?

templates/vue/src/store/actions.js
```diff
@@ -131,3 +131,21 @@ export async function updateUserFavourites({ commit }, favourites) {
...
+      delete n.fy
+    } else {
+      n.fx = n.coordinates.x
+      n.fy = n.coordinates.y
+    }
+  })
+  thisTapestryTool.setDataset(tapestry)
+  thisTapestryTool.setOriginalDataset(tapestry)
+  thisTapestryTool.reinitialize()
+  // commit("init", tapestry)
```
codygagnon, line 150: Should this comment be removed?

templates/vue/src/services/TapestryAPI.js
```diff
@@ -201,4 +204,10 @@ export default class {
     })
     return response
   }
+
+  async getAllContributors() {
+    const url = `${apiUrl}/tapestries/contributors?post_id=${this.postId}`
```
wynnset, line 209: Shouldn't this be ``${apiUrl}/tapestries/${this.postId}/contributors`` to be consistent with the rest of the endpoints?

templates/vue/src/components/SettingsModal.vue
```diff
@@ -35,6 +35,15 @@
               {{ autoLayout ? "Enabled" : "Disabled" }}
             </b-form-checkbox>
           </b-form-group>
+          <b-form-group
+            label="Superusers Override Permissions"
```
wynnset, line 39: Can you change this to say "Show all nodes to me by default"

templates/vue/src/components/SettingsModal.vue
```diff
@@ -35,6 +35,15 @@
...
+            description="When this is enabled, all WordPress superusers (Administator roles, Editor roles or the
```
wynnset, line 40: Can you change this to the following:
`If enabled, editors of this tapestry would be able to see and edit all nodes even for nodes with the "read" permission off. If disabled, superusers will be able to use the filter to view such nodes, but they won't appear in the tapestry by default. Note: Editors of this tapestry include users with the Administator or Edit roles, and the author of this Tapestry.`

PR #631, opened 2020-07-08
629 - Newly added nodes should not overlap each other

codygagnon:
Randomized new node coordinates such that they are ~~approximately within the bounds of the existing tapestry~~ within 3 x the selected node radius.


wynnset:
Hi @codygagnon this is not an ideal solution. Please bring it up in our next dev meeting so we can chat about it and devise a better solution.

narendrasss:
Thanks @codygagnon, this is better. Might be a good idea to add a minimum distance as well otherwise I'm getting this overlap:

Screen Shot 2020-07-09 at 13 46 33

Having said that, I'm still not sure if I like the random behaviour. I get that it might be time consuming though, so I'm not sure if it's worth spending 1-2 full days to design an algorithm to place the nodes in its most ideal location. Thoughts? @wynnset @seanRong @liam-armstrong

codygagnon:
Although this may not be a perfect solution, I believe that it is a good _interim_ solution until a perfect solution can be implemented. This code solves the original problem of newly added nodes completely overlapping each other.

seanRong:
In general the overlapping has been reduced a lot, and even when it does overlap you can tell that you actually successfully made a new node since it isn't right ontop of an old one. The 3*radius boundary helps because the new ones aren't appearing halfway across the tapestry. 

Seconding what @narendrasss mentioned about making an algorithm to put a new node in an optimal location, though I think we should get this merged as it's an improvement over what master has currently.

narendrasss:
@codygagnon I updated the code slightly so that the maximum distance is 2x the radius instead of 3x. I made this change because I saw some nodes being created outside of my visible window.

Otherwise, it's good to go so I'm merging it in!

narendrasss:
@liam-armstrong Can you code review this quickly so we can merge it in?

PR #645, opened 2020-07-16
653 - Integrate Cypress to run automatically on Github

narendrasss:
(This is a test PR)


liam-armstrong:
Some of the tests aren't passing, but the infrastructure for running the tests is up and going, and won't impact local dev environments so this is good to merge into #509 AFAIK.

narendrasss:
Thanks for your hard work on this @liam-armstrong! I think we can get around the link preview issue by using [secrets](https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets) and attaching the API key there.

I think @wynnset is the only one who can set this up however so we can talk about it more in tomorrow's standup.

wynnset:
We should deal with the link preview issue as per @narendrasss above and also install gf using the gravityforms cli (which requires wp-cli as well): https://wordpress.org/plugins/gravityformscli/

narendrasss:
@liam-armstrong Can you look into integrating the gravity forms cli when you get a chance?

wynnset:
`LINK_PREVIEW_API_KEY` and `GRAVITY_FORMS_API_KEY` secrets have been added. I had to choose different secret names because only uppercase characters and underscores are allowed so CamelCase did not work, so those will need to be updated.

narendrasss:
Going to keep this up for the time being but might be replaced with #705.

narendrasss:
Gonna close this in favor of #707.

PR #651, opened 2020-07-22
650- WordPress content titles become distorted in Tapestry

### LINKED ISSUE #650: WordPress content titles become distorted in Tapestry
Bita:

A student brought this to my attention and their experience can be seen in the image I have attached. Their Wordpress content looks fine on the Wordpress side, but on the Tapestry it changed to a different name. I noticed this was the case for other students too when looking through the Wordpress content list.

seanRong:
Closes #650

Renders post titles with v-html directive in WpPostMedia.vue, so special characters such as apostrophes are not rendered as literal unicode.

Some backend changes to how title and content is returned, and parsing is done there. Changed how apostrophes are handled so it matches commonly typed apostrophe.

### DIFF
--- a/templates/vue/src/components/Combobox.vue
+++ b/templates/vue/src/components/Combobox.vue
@@ -16,7 +16,7 @@
         @mousedown.prevent="handleClick(option)"
       >
         <div class="combobox-item">
-          <slot :option="option">{{ option.toString() }}</slot>
+          <slot :option="option"></slot>
         </div>
       </button>
     </div>
@@ -119,11 +119,11 @@ export default {
   },
   watch: {
     text(newText) {
-      this.inputValue = newText
+      this.inputValue = this.decodeHtmlEntity(newText)
     },
   },
   created() {
-    this.inputValue = this.text
+    this.inputValue = this.decodeHtmlEntity(this.text)
   },
   methods: {
     handleBlur() {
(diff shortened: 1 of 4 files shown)


codygagnon:
This still needs to be fixed in the dropdown menu in the node modal.

image

seanRong:
Thanks @codygagnon , I fixed it

liam-armstrong:
@seanRong I'm marking this as needs work following our conversations this morning

liam-armstrong:
I'm still having the same issue as Cody above, the dropdown text looks fine, but when the post is selected it renders wrong inside the input

image

templates/vue/src/components/Combobox.vue
```diff
@@ -151,6 +152,11 @@ export default {
     getValue(option) {
       return typeof option === "string" ? option : option[this.itemValue]
     },
+    decodeHtmlEntity(str) {
+      return str.replace(/&#(\d+);/g, function(match, dec) {
+        return String.fromCharCode(dec)
+      })
+    },
```
liam-armstrong, line 159: Seeing as you're calling this method regularly, and on change to the input, this might be better done as a computed property.

liam-armstrong:
Along with the code change above, the search feature is also not working to match special characters with post names. For example, if I search for a normal string, it matches the beginning characters fine:

image

But as soon as I add a special character, it returns all post names, which I believe is the default setting for a non-match:

image

I would guess this is because you're just re-encoding the string for display and not changing the way the data is being stored and then matched in the `visibleOptions` computed property.

templates/vue/src/components/Combobox.vue
seanRong: Ended up not needing this method, so it was removed.

PR #660, opened 2020-07-28
654 - Be told if my system, browser, or resolution does not meet the requirements

### LINKED ISSUE #654: Be told if my system, browser, or resolution does not meet the requirements
Show warning if user is on IE

codygagnon:
Closes #654 

If the user is accessing a tapestry via Internet Explorer, a warning message should appear. Should work for Internet Explorer 10 and below (in which case the user agent will include "MSIE") and Internet Explorer 11 (in which case the user agent will include "Trident/7.0").


templates/single-tapestry.php
```diff
@@ -9,6 +9,12 @@
  * @return object null
  */
 
+$browser = $_SERVER['HTTP_USER_AGENT'];
+
+if(is_int(strpos($browser, "MSIE")) || is_int(strpos($browser, "Trident/7.0"))) {
+    echo "<html><div>To use the Tapestry Tool, please use a browser other than Internet Explorer.</div></html>";
```
narendrasss, line 15: I think inserting this `html` tag makes it invalid HTML. Does it work if you remove the html tag?

```suggestion
    echo "<div>To use the Tapestry Tool, please use a browser other than Internet Explorer.</div>";
```

wynnset:
Let's expand the check to make sure the browser can run ES6. Some older browsers (non-IE) also cannot run ES6.

codygagnon:
> Let's expand the check to make sure the browser can run ES6. Some older browsers (non-IE) also cannot run ES6.

@wynnset Should we use feature detection? What features of ES6 should be detected?

Feature detection is recommended here: https://stackoverflow.com/questions/29046635/javascript-es6-cross-browser-detection and here: https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection

templates/single-tapestry.php
```diff
@@ -9,6 +9,26 @@
...
+    var str = 'class ಠ_ಠ extends Array {constructor(j = \"a\", ...c) {const q = (({u: e}) => {return { [`s${c}`]: Symbol(j) };})({});super(j, q, ...c);}}' + 
+    'new Promise((f) => {const a = function* (){return \"\u{20BB7}\".match(/./u)[0].length === 2 || true;};for (let vre of a()) {' +
+    'const [uw, as, he, re] = [new Set(), new WeakSet(), new Map(), new WeakMap()];break;}f(new Proxy({}, {get: (han, h) => h in han ? han[h] ' + 
+    ': \"42\".repeat(0o10)}));}).then(bi => new ಠ_ಠ(bi.rd));';
+
+    try {
+        eval(str);
+    } catch(e) {
+        var div = document.createElement(\"div\")
+        var message = document.createTextNode(\"To use the Tapestry Tool, please use a browser that supports ES6.\")
```
wynnset, line 24: This message is not very user friendly. Average Joe has no idea what ES6 is so we should say "Please update your browser" and list a list of links to common browser websites or instructions for updating their browsers. The messaging should be created for a very non-savvy user.

wynnset (changes requested):

templates/single-tapestry.php
wynnset: This whole JS is better to be included as a separate JS file rather than being inserted inline.

wynnset:
Couple more things:
- It would be great to include the JS in `tapestry.php` instead the same way we are including other JS.
- Please hide the tapestry div if this test fails
- It might be nicer to put the JS fail message inside a div in `single-tapestry.php` with `display: none;` and then show it if the test fails using the script.

narendrasss:
@wynnset I added a list of older browsers and moved it to a standalone script. I decided not to move the html to `single-tapestry` because I remember Cody had issues where the document `body` wasn't even loaded (in which case no other html would be loaded as well).

@seanRong Can you test if this works? I can't test it because the older browsers don't work on latest MacOS :(

wynnset (approved):

wynnset (approved):

wynnset (approved):

PR #666, opened 2020-07-29
664 - Keep node authors on import

### LINKED ISSUE #664: Keep node authors on import

narendrasss:
Closes #664

### DIFF
--- a/classes/class.tapestry-node.php
+++ b/classes/class.tapestry-node.php
@@ -38,6 +38,8 @@ class TapestryNode implements ITapestryNode
     private $fullscreen;
     private $childOrdering;
     private $fitWindow;
+    private $license;
+    private $references;
 
     /**
      * Constructor.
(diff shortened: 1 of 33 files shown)


narendrasss:
@liam-armstrong Thanks for the in-depth review!

1. That is quite strange. This is definitely a bug so even though this may not be an issue at the moment it should still be something we look at. Thanks for pointing out that edge case.
2. I agree with this. I was thinking we should improve the export functionality to include the locked node and to also include all possible nodes even those that aren't loaded. This feature would have saved us a lot of trouble earlier this week when Steven only managed to recover his own nodes.
3. I agree with this as well! I think this is part of an Asana task that we have already though.

wynnset:
@narendrasss are you going to address @liam-armstrong 's comments above? Otherwise, can you make sure we take note of them in the form of an Asana task so we can address it later? I think those are all very valid points.

narendrasss:
@wynnset I think it's better to address these issues as separate PRs because it would make this one quite bloated. I'll make the related asana tasks.

PR #667, opened 2020-07-30
655 - WordPress posts only search within the first 10 posts 

### LINKED ISSUE #655: WordPress posts only search within the first 10 posts
Hi,
If I try to create a Wordpress node, the only available posts are the 10 most recent ones published on the "Tapestry @ UBC" site. If I try to search for it, it will only search from that specific list. The post would be "Inferential Statistics and Errors". 

Because of the error, I decided to do a couple of different nodes that no longer requires that specific post. However, I created a new Wordpress post and it appears as though the same issue will occur once my new post is pushed out of the 10 most recent posts list.

Aidin: Remove the limit altogether for search - show max 100 and allow user to scroll.

https://stackoverflow.com/questions/35728943/wordpress-rest-api-v2-return-all-posts

seanRong:
Instead of just fetching only 10 posts, we now fetch all posts from our database. 
The query may take a long time - about 13 seconds for 330 posts, so they will be cached after the first time the call is made. Thus the user may peruse options from last time (which probably will be more or less the same) as we fetch again in the background. There is a different empty message now for when posts are being fetched.

I tried to optimize it as much as possible, but we're getting bottlenecked by the WP api speed.
~~For reference, on my machine with postman, calling the first page of all posts with a 100 limit needs 11 secs to complete, while setting the limit to 10 only reduces that time to 7 secs.~~
After removing unnecessary info from the api call we are down to a bit over 6 secs for 100 posts at a time on my machine, Aidin's has it at under 2 seconds which is good. Mine just might be slow. 

There is an extension called fakerPress which helps generate fake posts (for testing).

Closes #655

### DIFF
--- a/templates/vue/src/components/node-modal/content-form/WpPostForm.vue
+++ b/templates/vue/src/components/node-modal/content-form/WpPostForm.vue
@@ -41,7 +41,6 @@ export default {
     WordpressApi.getPosts().then(posts => {
       this.wpPosts = posts
     })
-
   },
 }
 </script>
--- a/templates/vue/src/components/node-modal/content-form/WpPostForm.vue
+++ b/templates/vue/src/components/node-modal/content-form/WpPostForm.vue
@@ -4,7 +4,7 @@
       v-model="node.typeData.mediaURL"
       item-text="title"
       item-value="id"
-      empty-message="There are no Wordpress posts yet. Please add one in your WP dashboard."
+      :empty-message="emptyMessage"
       :options="wpPosts"
     >
       <template v-slot="slotProps">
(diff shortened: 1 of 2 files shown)


seanRong:
This should be a lot faster now, also please check if it does display a different message when there are no posts to be found.

templates/vue/src/components/node-modal/content-form/WpPostForm.vue
```diff
@@ -34,12 +34,21 @@ export default {
...
+    WordpressApi.getPosts()
+      .then(posts => {
+        this.wpPosts = posts
+      })
+      .then(() => {
+        if (this.wpPosts.length < 1) {
+          this.emptyMessage =
+            "There are no Wordpress posts yet. Please add one in your WP dashboard."
+        }
+      })
```
liam-armstrong, line 51: This is a great way to handle this! Good thinking.

PR #673, opened 2020-08-07
662 - [TYDE] Bug fix: Copilots cannot access their nodes if it depends on a teen node

### LINKED ISSUE #662: Copilots cannot access their nodes if it depends on a teen node
If a copilot-only node is locked and it depends on a teen-only node, the copilot node never unlocks even if the teen node is completed.

Steps to reproduce:
Create a teen-only node and a copilot-only node
Under access in the copilot-only node, add a "node completion" lock condition that depends on the teen-only node
Login as a teen and complete the teen-only node
Login back as a copilot and try to access the copilot-only node

Expected behaviour:
Copilot can access the node

Actual behaviour:
The node is still locked

narendrasss:
Closes #662

This PR fixes an issue where a copilot node will never be accessible if its locking condition is dependent on the completion of a teen-only node (a regular node).

The bug occurred because the old code only checks if the user has completed the node specified by the locking condition. Copilots will never complete a teen node because they are not able to update the teen's progress.

The fix works by checking if the current user is a copilot, and if they are, check if their associated **teen** has completed the node instead.

### DIFF
--- a/classes/class.tapestry-node.php
+++ b/classes/class.tapestry-node.php
@@ -246,10 +246,16 @@ class TapestryNode implements ITapestryNode
         foreach ($conditions as $condition) {
             switch ($condition->type) {
                 case ConditionTypes::NODE_COMPLETED:
-                    if ($userId && $userProgress->isCompleted($condition->nodeId, $userId)) {
-                        $condition->fulfilled = true;
+                    $user = get_userdata($userId);
+                    if ($user) {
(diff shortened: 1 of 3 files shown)


wynnset:
We're gonna need to also fix this on the [release-tyde-1.0](https://github.com/wynnset/tapestry-wp/tree/release-tyde-1.0) branch since that's the live version

PR #678, opened 2020-08-10
674 - accept/reject: Add private nodes to a tapestry that only I have access to

### LINKED ISSUE #674: Add private nodes to a tapestry that only I have access to
All users would be able to add private nodes to any nodes that are only visible to them. There shouldn't be any restrictions for this (as in anyone should be able to add private nodes without being given explicit permission or anything)

Users should be able to add nodes even if the author hasn't given them add/submit permissions.

Instead of the "Submit" button on the NodeModal, we would add 2 buttons:
    "Save as Private Draft"
    "Publish"

liam-armstrong:
Closes #674 
Adds: 
 - Creation of private nodes by any authenticated user
 - Filter for private nodes from everyone, including administrators, regardless of Tapestry settings
 - Custom styling for private nodes

There was some ambiguity in the requirements for this task, so any suggestions regarding UX or visuals is appreciated!


seanRong:
`Uncaught (in promise) TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
    at parseDataset (mutations.js?ba46:39)
    at Store.init (mutations.js?ba46:6)
    at wrappedMutationHandler (vuex.esm.js?358c:785)
    at commitIterator (vuex.esm.js?358c:407)
    at Array.forEach (<anonymous>)
    at eval (vuex.esm.js?358c:406)
    at Store._withCommit (vuex.esm.js?358c:565)
    at Store.commit (vuex.esm.js?358c:405)
    at Store.boundCommit [as commit] (vuex.esm.js?358c:350)
    at VueComponent.mappedMutation (vuex.esm.js?358c:936)`

And the tapestry does not load.

This is the error given when admin logs in after an authenticated user has created their private node. It is also the case if authenticated logs in after admin makes a private node. This happens regardless of incognito/cache status

seanRong:
OK that took a bit longer than expected, but it loads properly now.

wynnset:
> Something I noticed is that front-end code relating to Wordpress is a bit spread out across the app, e.g. checking if the user is logged in, if they can edit the Tapestry, etc. It might be good to extract this out into a "wordpress layer". Something to think about.

I do like this idea. @narendrasss can you create a task on Asana in the backlog and detail what your thoughts are in there? Put it at the top of the backlog for now and I'll prioritize it.

narendrasss:
> > Something I noticed is that front-end code relating to Wordpress is a bit spread out across the app, e.g. checking if the user is logged in, if they can edit the Tapestry, etc. It might be good to extract this out into a "wordpress layer". Something to think about.
> 
> I do like this idea. @narendrasss can you create a task on Asana in the backlog and detail what your thoughts are in there? Put it at the top of the backlog for now and I'll prioritize it.

Done! https://app.asana.com/0/1176098622515689/1195710906078745

wynnset:
Bug: I cannot save/publish an existing non-draft (already published) node if it's attached to more than a single node:
image

seanRong:
The bug has been fixed.

I think making a published note back into a draft was in the spec somewhere. There's a nontrivial amount of code relating to it. (Although, I did not see a guard for something like published -- draft -- published which wouldn't be possible without converting a published back into a draft)

wynnset:
Hi @seanRong , the "Save as Draft" button has now completely disappeared, even when I'm adding a new node:
image

seanRong:
Merge conflicts might take a while to figure out

The automerge is removing a lot of important code

wynnset:
Couple of issues:

- [ ] Node progress bars are appearing way off where they need to be: 
image

- [ ] Leaf nodes that are published still have the "Save as Draft" as an option. This seems to have been fixed for non-leaf nodes (i.e. nodes that are connected to more than a single node)

seanRong:
@wynnset  I've addressed the weird progress bars, but I could not reproduce leaf nodes having safe as draft option. It works fine on my end both for leaf and non-leaf nodes.

wynnset:
@seanRong those issues I reported seem to be fixed now. Thanks

yungalyx:
@seanRong did a pretty extensive test.. found no relevent bugs... (creating a new private node resulted in duplications once but i think that was addressed by a seperate PR). everything LGTM!

wynnset:
I'm getting this error which should technically never happen: 
image

Note: I was logged in as a user that did not have "add" permission for any nodes.

wynnset:
Bug: Permission tab not appearing and I'm seeing a bunch of errors in the console. This is not just for draft nodes, but for all nodes.
image

wynnset:
By the way, I found a few different issues and fixed them. Most importantly, I realized I couldn't add draft nodes to other draft nodes, which is a limitation that shouldn't have been there. As a result of fixing this, I discovered several other issues and fixed those as well. You can see a list of what I fixed in the previous commit message.

seanRong:
This could be good news, or bad news, but I don't have that permissions table issue at all.

- even after hard reload + cache clean
- even on incog, even as non-admin user
- even on old tapestries with old private nodes
- published nodes all fine

I made sure I pulled and all. Did this thing happen consistently?

wynnset:
I actually realized I'm experiencing this on master as well... So I guess it's not related to this PR.

rlnsy:
Added myself to try to reproduce the error, but realized it was on master as well. As it turns out I'm having other errors on master - will follow up in slack.

PR #679, opened 2020-08-11
[No ticket] Bug fix: Buggy traversal algorithm causing some nodes to not appear

narendrasss:
**Changes**

* Change the `filterNodeIdsByPermission` implementation to use a BFS beginning from the root node.
  * Previously this was a recursive DFS beginning from the target node, causing some strange behaviour where nodes that should be returned are not returning.
* Remove the `_pathIsAllowed` helper as it is now obsolete with the new implementation.


wynnset:
This is causing the node progress bar to not appear for some reason. Screen recording: https://share.getcloudapp.com/6que52Kq

wynnset:
One of Steven's TA's, Fiza Arshad, reported that the Week 5 Grading tapestry was not loading for her. I was also not able to load that tapestry. After looking into it, I realized it was because of this change. I reverted the file back to before we pushed these changes and it started working.

We should look into seeing why that tapestry was not loading with these changes and fix it, or otherwise close this PR

narendrasss:
@wynnset I just pushed a fix for the infinite loop problem. It seems that this Tapestry may have nodes with the same node ID, causing some nodes to be traversed more than once. I changed the algorithm to ensure that each node is only visited one time at most.

wynnset:
Also, I've pushed this code live and going to keep it there for now to see if we get any negative reports, but I tested it with Week 5 grading tapestry and some other tapestries and it seemed to work well so far!

narendrasss:
@wynnset It might be useful, but because it's now iterative I wasn't sure where to put it. For the time being I've removed the include from the file.

wynnset:
This tapestry failed to load with this code change so I, again, had to switch back. https://ubc.tapestry-tool.com/spatial-xp/tapestry/geob-103/#/

narendrasss:
@wynnset I copied that Tapestry over locally and it loads fine, could it be something else that was in production?

narendrasss:
As discussed, let's wait to verify this until the next release.

wynnset:
#771 fixed this

PR #681, opened 2020-08-12
652 - enhance duplication and import/export

seanRong:


classes/class.tapestry-node.php
```diff
@@ -448,7 +448,20 @@ private function _getAuthorInfo($id)
...
                 'id' => $id,
                 'name' => $user->display_name,
+                'email' => $user->user_email,
+                'original_author_name' => "",
+                'original_author_email' => "",
             ];
         }
     }
+
+    public function update_original_author($node) {
```
seanRong, line 458: This tries to update the original author field under the author field in a tapestry node, but this change doesn't seem to be saved. The next export still has original author being empty.

PR #682, opened 2020-08-12
652 - enhance duplication and import/export

### LINKED ISSUE #652: Enhance import/export functionality

seanRong:
Closes #652

Gives a loading screen for both export and duplicate.
Export now includes wordpress site url.
Strips permissions for users and user groups that don't exist on the new site if the import is from a different site.
Maintains permission level if the user group still exists.
Shows a summary screen after a different site import to see what has changed.


wynnset:
Please fix conflicts @seanRong

endpoints.php
```diff
@@ -285,6 +285,13 @@
             'callback' => 'getTapestryContributors',
         ],
     ],
+    'GET_ALL_ROLES' => (object) [
+        'ROUTE' => '/all_roles',
```
narendrasss, line 289: Small nitpick, but I think a better endpoint for this would be `/roles`. I think it better conforms to the standard of REST being organized by resources.

classes/class.tapestry.php
```diff
@@ -134,7 +134,9 @@ public function getNodeIds()
     public function addNode($node)
     {
         $tapestryNode = new TapestryNode($this->postId);
+        // $tapestryNode->update_original_author($node);
```
narendrasss, line 137: I think it would be cleaner to remove code that is commented out, unless it's commented out as part of a WIP feature (in which case, I think there should be a comment stating so).

templates/vue/src/components/RootNodeButton.vue
```diff
@@ -75,15 +89,52 @@ export default {
...
+          .catch(err => {
+            this.error = err
+          })
       }
       reader.readAsText(file)
     },
+    async prepareImport(data) {
+      let wp_roles = await client.getAllRoles()
+      for (let node of data.nodes) {
+        const originalPerms = node["permissions"]
```
narendrasss, line 114: Let's try to avoid accessing object properties using strings wherever possible (unless it's required, like `site-url`). I feel like this should be in the linter, but if not, I'll add a rule for it.

templates/vue/src/components/SettingsModal.vue
```diff
@@ -64,8 +64,17 @@
...
         <b-tab title="Advanced">
-          <b-button block variant="light" @click="exportTapestry">
-            Export Tapestry
+          <b-button
+            id="export-button"
+            block
+            variant="light"
+            :class="isExporting ? 'disabled' : ''"
+            @click="exportTapestry"
+          >
```
narendrasss, line 73: I think we should also set the `disabled` prop to true when it's exporting, otherwise the button would still be clickable while it's loading (if not by the cursor, then programmatically).

narendrasss:
Just tested this - it works well! I think it might be good for the confirmation modal not to show up at all when there are no changes, and to add a text that says "Import successful". What do you think? @wynnset

templates/vue/src/components/SettingsModal.vue
seanRong: Was a little confused about this one, it's currently disabled when exporting

narendrasss: I might be misunderstanding, but right now it looks to me like the class is set to disabled but the button isn't actually disabled, i.e. the `disabled` attribute of the button is not set to `true`.

jaskwong:
Doesn't compile (duplicated updateDataset() function in mutations.js)

seanRong:
Fixed, must've not dumped cache when testing!

shirleywbi:
Are the following expected behaviour?
- Invalid tapestry import json leads to infinite load
- Added users do not appear under permissions in json

Currently having a bit of trouble testing this since I'm having trouble generating a json from a different site.

shirleywbi (commented):
Not sure if Export Tapestry is just executing very quickly but I can't seem to see the loader when it's exporting. Ran the following with a setTimeout to slow it down to see the spinner. The overlay looks a bit off-centered.
image
Also, expected to see a modal confirmation of export and duplicate. I got an alert confirmation for duplicate ("Your new Tapestry is ready! Click on the link below to view it."). No feedback from export button except for the fact that a file had been downloaded.
image

shirleywbi (commented):
Infinite load when importing file with error then clicking import button and not going through with it: 
1. Click on Import a Tapestry
2. Use invalid json
3. Click on Import a Tapestry again
4. Closing the file explorer leads to infinite loading

narendrasss (approved):
Looks good! Just ran the linter and updated the error message to be a bit more descriptive.

wynnset:
When I press "Export Tapestry", it downloads and I get a message saying "Your new Tapestry is ready! Click on the link below to view it." -- how was this not caught by the testing?!

wynnset:
@seanRong just wondering why you have a commented out line in your PR. Otherwise looks good.

PR #683, opened 2020-08-12
680 - accept/reject: Allow users to submit nodes and resubmit rejected nodes

### LINKED ISSUE #680: Allow users to submit nodes and resubmit rejected nodes
Instead of the submit buttons, user would see:
    "Save as Private Draft"
    "Submit to Main Tapestry"

If the node has already been rejected, instead of "Submit", it should say "Resubmit"

liam-armstrong:
Closes #680 
Adds:
 - Node visibility controls based on status attribute:
Nodes with `publish` can be seen by all users
Nodes with `draft` and `reject` can only be seen by their author
Nodes with `submitted` can only be seen by their author and Tapestry administrators
- Editors see a review form when editing `submitted` nodes, and are given the option to publish or reject a node, with a comment
- Comments are stored as an array of objects related to each node. Each comment stores the rich text comment, a time stamp in ISO 8601 format, and comment author's name and email
- Nodes on the frontend have different border colors based on status. The colors currently here are just placeholders, but demonstrate what can be done.


seanRong:
After testing, I think everything here works ok with regards to publish/draft and reject/accept. I might have to test again after fixing the issue in 674 though.

seanRong:
Verified that the non-reviewer version of the modal is the default, so under laggy circumstances the reviewer is the one who will briefly see the submitter modal, which shouldn't be an issue.

narendrasss (approved):
Code LGTM, once tested should be good to go.

seanRong:
Submission form issues should be fixed here

jaskwong:
- After submitted nodes are approved and published, the author of the node can still add/edit/delete the node (and save it as a private draft again)
- The administrator cannot change the permissions for the author (but these 2 are like the same problem)

seanRong:
Merge confs will be addressed once 674 fully done

jaskwong:
Tested, works great!

jaskwong:
Creating a node and adding 2 links to the node (like a triangle) and then submitting it, if the admin rejects it, the user can no longer resubmit the node nor save as private draft

seanRong:
@jaskwong This shouldn't even be possible, irrc. How did you manage to add more than 1 link to a draft node?

jaskwong:
@seanRong it's a pretty obscure case but if the user has add privileges to another node, then I was able to add a link from that node to the new node

seanRong:
NOTE: There are some issues even after resolving conflicts

narendrasss:
Assigning this to myself to add integration tests.

narendrasss:
@seanRong @wynnset I notice that when a node is accepted, the author of the node is set to the _reviewer_ rather than the original node author. Is this intended? Would it not be more appropriate to leave the author as the original submitter?

wynnset:
> @seanRong @wynnset I notice that when a node is accepted, the author of the node is set to the _reviewer_ rather than the original node author. Is this intended? Would it not be more appropriate to leave the author as the original submitter?

Yes, the author should not be changed to the reviewer. It should remain as the original author.

seanRong:
@wynnset Looking back in the commit history there's one that explicitly changes the author from the submitter to the admin. Do you remember why this was requested?

wynnset:
@seanRong just looking at the comment history here and found [this](https://github.com/wynnset/tapestry-wp/pull/683#issuecomment-702421676) by @jaskwong:
> * After submitted nodes are approved and published, the author of the node can still add/edit/delete the node (and save it as a private draft again)
> * The administrator cannot change the permissions for the author (but these 2 are like the same problem)

This makes sense in that the person who submitted the node should not be able to edit a submitted node. However, we do want to have them remain as the author for attribution / copyright purposes. We should keep the author the same, but not allow them to make changes anymore after it's submitted. We can do this by either:
1. Marking the accepted nodes with a different status (e.g. "accepted") rather than just "publish", OR
2. Create a new property to denote a particular node was a submitted node originally.
We can then ensure we do not allow the author of the node to add/edit any nodes with such a distinction.

Does that make sense? Open to other ideas.

seanRong:
PR 712 actually has the accepted status versus just the normal published, so it could be differentiated during that PR. Would it be fine to leave it for now?

narendrasss:
@seanRong gonna assign this back to you to discuss Aidin's comment above. I added some tests for the `ReviewForm` and fixed a minor edge case. There's more tests to add here but I think they belong under e2e/Cypress.

wynnset:
As discussed, please add the feature I mentioned above into this PR. Also, as discussed, the scope of this PR can be expanded to include the actual accept/reject part of the workflow by the reviewer and PR #712 can be changed to be only for the commenting part. Thanks @seanRong

jaskwong:
Tested, works well!

wynnset:
If I submit a node, it should not allow me to edit it anymore (it currently does). This task requires the approval process to happen within the sidebar in a new "Review" section (under the copyright section), with the review comments in there. I currently see cannot see the approval process in the sidebar. Instead I have to press "Edit" to be able to approve/reject something. Additionally, I cannot see the comment thread either. Also, the "Accept" button is supposed to read "Accept and Add".

seanRong:
> This task requires the approval process to happen within the sidebar in a new "Review" section (under the copyright section), with the review comments in there. I currently see cannot see the approval process in the sidebar. Instead I have to press "Edit" to be able to approve/reject something. Additionally, I cannot see the comment thread either.

These changes were made in 712, and weren't part of the original scope for 680. (Asana reflects that)

wynnset:
> These changes were made in 712, and weren't part of the original scope for 680. (Asana reflects that)

Ok we can move the "Accept / Reject" buttons to the sidebar in #712, then, and also do the commenting part in there (as planned).

templates/vue/src/components/NodeModal.vue
```diff
@@ -257,16 +293,19 @@ export default {
...
           Helpers.hasPermission(this.parent, this.type) &&
           (!this.parent || this.parent.status !== "draft")
         )
-      } else if (this.node.status === "draft" && this.type === "edit") {
+      } else if (this.node.status !== "draft" && this.type === "edit") {
         return this.getNeighbours(this.nodeId).some(neighbourId => {
           let neighbour = this.getNode(neighbourId)
           return (
-            neighbour.status !== "draft" && Helpers.hasPermission(neighbour, "add")
+            neighbour.status === "publish" && Helpers.hasPermission(neighbour, "add")
```
wynnset, line 308: If a node is published, why would it matter what the status of its neighbours are?

templates/vue/src/components/NodeModal.vue
```diff
@@ -287,6 +326,9 @@ export default {
       }
       return this.hasDraftPermission(ID)
     },
+    tapestryEditor() {
```
wynnset, line 329: All methods and calculated values should start with a verb for clarity. This should read `canEditTapestry`.

templates/vue/src/components/NodeModal.vue
```diff
@@ -516,7 +555,7 @@ export default {
             type: "",
           }
           await this.addLink(newLink)
-          if (this.node.status !== "draft") {
+          if (this.node.status !== "draft" && this.node.status !== "submitted") {
```
wynnset, line 558: It's not easy to understand why these conditions are here. It would be great if you could create some calculated values that are named in a way to describe what you are checking for exactly or otherwise add a comment to say why this condition is being added.

templates/vue/src/components/node-modal/DeleteNodeButton.vue
```diff
@@ -40,11 +37,14 @@ export default {
       if (this.isDisabled) {
         return this.isRoot
           ? "Root node can only be deleted if there are no other nodes in the tapestry."
-          : "Only nodes with a single connection can be deleted."
+          : "Only nodes with a single connection can be deleted or made into drafts."
```
wynnset, line 40: Nodes cannot be made into drafts.

templates/vue/src/utils/Helpers.js
```diff
@@ -143,8 +143,13 @@ export default class Helpers {
...
     }
 
-    // Check 2: User is the author of the node
-    if (node.author && wpData.currentUser.ID == parseInt(node.author.id)) {
+    // Check 2: User is the author of the node, if the node is not an accepted or submitted node
+    if (
+      node.author &&
+      wpData.currentUser.ID == parseInt(node.author.id) &&
+      node.status !== "accept" &&
+      node.status !== "submitted"
```
wynnset, line 151: These conditions are very confusing. Why are accepted and submitted nodes not permitted, but the other ones are? Can you explain that in a comment to clarify or otherwise think how you can put this code in a way that is easier to understand?

wynnset (changes requested):
Sorry @seanRong, this still needs a bit more work to become functional and maintainable. I've made some comments in the code but also:
- Accepted nodes should have a progress bar same as published nodes.
- Accepted nodes should appear for everyone. They are the same as published nodes. Right now, they only appear to the author of the node and admins.

I find adding of all the conditions around statuses has made the code a lot harder to read and understand. Maybe it's better if we create a different field to designate a node as accepted rather than using the status field - that way we won't have to check everywhere in the code for both. We might also want to do the same with other fields so that we only have draft and published as statuses. We could have a new field called `reviewStatus` and have it set to either submitted, rejected, accepted, or null (in case it was never reviewed).

templates/vue/src/components/NodeModal.vue
seanRong: Done

templates/vue/src/components/NodeModal.vue
```diff
@@ -106,41 +106,74 @@
...
-        :disabled="!canSubmit"
-        @click="handleSubmit"
+        <delete-node-button
+          v-if="type === 'edit'"
+          :node-id="Number(nodeId)"
+          @submit="loading = true"
+          @message="setDisabledMessage"
+        ></delete-node-button>
+        <span style="flex-grow:1;"></span>
+        <b-button size="sm" variant="light" @click="close">
```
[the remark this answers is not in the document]
seanRong: Added

templates/vue/src/components/node-modal/DeleteNodeButton.vue
seanRong: Probably missed this in merge, fixed.

classes/class.tapestry-node.php
```diff
@@ -38,6 +38,7 @@ class TapestryNode implements ITapestryNode
     private $fullscreen;
     private $childOrdering;
     private $fitWindow;
+    private $comments;
```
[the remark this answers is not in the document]
seanRong: Changed

templates/vue/src/components/NodeModal.vue
```diff
@@ -106,41 +106,74 @@
...
+                : "Submit"
+            }}
+            to Administrators for Review
+          </div>
+        </b-button>
+      </div>
+      <b-form-invalid-feedback
+        v-if="!(tapestryEditor && node.status === 'submitted')"
+        :state="canMakeDraft"
       >
```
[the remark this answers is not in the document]
seanRong: Gave me a headache until I remembered <template> existed haha

templates/vue/src/components/NodeModal.vue
seanRong: Yeah this check doesn't make much sense - removed.

seanRong:
> Accepted nodes should have a progress bar same as published nodes.
Accepted nodes should appear for everyone. They are the same as published nodes. Right now, they only appear to the author of the node and admins.

These  two points have now been addressed.

templates/vue/src/components/NodeModal.vue
seanRong: Clarified

templates/vue/src/utils/Helpers.js
seanRong: Clarified with minor change

templates/vue/src/components/NodeModal.vue
```diff
@@ -257,18 +291,14 @@ export default {
...
           Helpers.hasPermission(this.parent, this.type) &&
           (!this.parent || this.parent.status !== "draft")
         )
-      } else if (this.node.status === "draft" && this.type === "edit") {
-        return this.getNeighbours(this.nodeId).some(neighbourId => {
-          let neighbour = this.getNode(neighbourId)
-          return (
-            neighbour.status !== "draft" && Helpers.hasPermission(neighbour, "add")
-          )
-        })
```
wynnset, line 271: I had said [before](https://github.com/wynnset/tapestry-wp/pull/683#discussion_r516274856) (I think it might have been missed in between my other feedback): The concept of `this.parent` does not exist within edit context. How do you decide which node is the parent if the node is connected to more than a single other node? That is why the `else if` statement was added. I think you'd need to revert so that we check ALL the neighbouring nodes and see if ANY of them are not published or accepted.

templates/vue/src/components/NodeModal.vue
```diff
@@ -287,6 +317,9 @@ export default {
       }
       return this.hasDraftPermission(ID)
     },
+    canEditTapestry() {
+      return wpData.wpCanEditTapestry == 1
```
wynnset, line 321: Please use the new `wp` service to be consistent.

templates/vue/src/components/NodeModal.vue
```diff
@@ -486,22 +521,18 @@ export default {
...
-
-        if (this.node.mediaType == "url-embed" && this.node.behaviour != "embed") {
-          await this.setLinkData()
-        }
-
-        if (this.shouldReloadDuration()) {
-          this.loadDuration = true
-        } else {
-          return this.submitNode()
-        }
```
wynnset, line 503: Just checking.. was this removed on purpose? I don't see it in any other part so wanted to make sure it's not a mistake.

templates/vue/src/components/NodeModal.vue
```diff
@@ -886,4 +918,8 @@ table {
 button:disabled {
   cursor: not-allowed;
 }
+
+.buttons-container > * {
+  margin: 0.25rem !important;
+}
```
wynnset, line 924: I suggest you use [bootstrap margin classes](https://getbootstrap.com/docs/4.3/utilities/spacing/) instead of adding a custom style.

templates/vue/src/components/tapestry-node/ProgressBar.vue
```diff
@@ -34,6 +33,11 @@ export default {
       required: false,
       default: false,
     },
+    status: {
+      type: String,
+      required: false,
+      default: "publish",
+    },
```
wynnset, line 40: I don't think this is needed anymore.

templates/vue/src/components/TapestryNode.vue
```diff
@@ -26,12 +26,27 @@
...
+        :y="node.coordinates.y"
+        :radius="radius"
         :data-qa="`node-progress-${node.id}`"
-        :radius="node.status === 'draft' ? radius + 15 : radius"
         :progress="progress"
         :locked="!node.accessible"
-        :draft="node.status === 'draft'"
+        :status="node.status"
       ></progress-bar>
+      <dash-array
```
wynnset, line 37: Could you please name this something more meaningful like `status-bar` maybe? Also, submitted, accepted, and rejected nodes should not be dashed - they should be solid. Only simple draft nodes should have a dashed line. Please also make it one or two pixels thicker. It's a bit hard to see when the nodes are small.

templates/vue/src/components/node-modal/ReviewForm.vue
```diff
@@ -0,0 +1,87 @@
...
+    return {
+      comment: "",
+    }
+  },
+  mounted() {
+    moment.tz.setDefault("America/Vancouver")
+  },
+  methods: {
+    handleReject() {
+      this.node.status = "draft"
```
wynnset, line 49: This line could be safely removed, I believe.

templates/vue/src/utils/Helpers.js
```diff
@@ -143,9 +143,16 @@ export default class Helpers {
...
 
-    // Check 2: User is the author of the node
-    if (node.author && wpData.currentUser.ID == parseInt(node.author.id)) {
-      return true
+    // Check 2: User is the author of the node, if the node is a draft
+    if (
+      node.author &&
+      wpData.currentUser.ID == parseInt(node.author.id) &&
+      node.status === "draft"
+    ) {
```
wynnset, line 151: You are erroneously removing the ability for node authors to edit their own node here (if their node is not draft). Published nodes should remain editable by their authors (unless they were submitted and accepted).

templates/vue/src/utils/constants.js
```diff
@@ -30,3 +30,11 @@ export const licenses = {
...
 
 export const DEFAULT_DEPTH = 3
+
+export const nodeStatuses = {
+  PUBLISH: "publish",
+  REJECT: "reject",
+  ACCEPT: "accept",
+  SUBMITTED: "submitted",
+  DRAFT: "draft",
+}
```
wynnset, line 40: a) this is never used anywhere
b) if you are planning to use this, please split it into `nodeStatuses` and `reviewStatuses`

wynnset (changes requested):
The updates look really good. Just a few small fixes and we are golden!

templates/vue/src/components/tapestry-node/ProgressBar.vue
seanRong: Yeah, this prop is redundant.

templates/vue/src/utils/constants.js
seanRong: Deleted (might be a good idea in the future though?)

templates/vue/src/components/node-modal/ReviewForm.vue
seanRong: Yep.

templates/vue/src/components/NodeModal.vue
seanRong: OK this one, I don't really want to change, as I would need to put a margin class on every single button under that div. Thoughts?

Also I don't think 0.25rem is one of the bootstrap defaults

templates/vue/src/components/NodeModal.vue
seanRong: This was removed because it is almost an exact duplicate of the error checking under handleSubmit(). It probably does not turn up if you cmd+F for it, because there's minute differences that set node status to draft in there.

templates/vue/src/components/NodeModal.vue
wynnset: Why was this marked as resolved? It has not been resolved...

seanRong: Sorry, I didn't actually push!

narendrasss:
@seanRong I pushed a super small commit that hopefully fixes the cy tests

wynnset (approved):

templates/vue/src/components/NodeModal.vue
```diff
@@ -470,44 +510,37 @@ export default {
...
         if (this.node.mediaType === "url-embed" && this.node.behaviour !== "embed") {
-          await this.setLinkData()
+          ;(async () => await this.setLinkData())()
         }
 
         if (this.shouldReloadDuration()) {
           this.loadDuration = true
         } else {
-          return this.submitNode()
+          return (async () => await this.submitNode())()
```
narendrasss, line 526: @wynnset Any reason why this was changed to not use async await? I think it's cleaner the way it was before.

narendrasss (changes requested):

templates/vue/src/components/NodeModal.vue
wynnset: I changed this because the `loading` data property was immediately changing to false and so the node modal was not disabling the buttons while that was happening and I could click cancel or submit again while the node was submitting. Do you have a suggestion of how to avoid that in a different way?

narendrasss (approved):

PR #685, opened 2020-08-14
684 - Bug fix: Tapestry nodes displaying oddly after a refresh when a node is open

### LINKED ISSUE #684: Bug fix: Refreshing with H5P media window open causes strange behaviour

seanRong:
Closes #684 

If you refresh with H5P window open, it will cause tapestry to appear strangely afterwards.

This forces the tapestry to rerender after clicking out of the H5P window.


seanRong:
I've noticed that this seems to happen if you refresh with _any_ modal open at all, can somebody else confirm?
I'll need to change the code a little if this is the case.

wynnset:
Unfortunately I've pushed this change as a patch and it's still having the same problem.

seanRong:
@wynnset , works on my machine. Did you reload on a non H5P modal?

wynnset:
Sorry, I meant to leave this on #686 not this one. This one is working now. I can merge it into master, but did you say you experienced it on non-H5P nodes too @seanRong ?

seanRong:
For me locally yes, does it occur for you as well?

seanRong:
For some reason, this fixed it for all other node types as well. Maybe because H5PMedia component is created and destroyed for all nodeModals?

Anyways, its ready to go.

narendrasss (approved):
LGTM

PR #692, opened 2020-08-18
691 - Add tapestry setting to load a map of earth instead of tapestry

### LINKED ISSUE #691: Add tapestry setting to load a map of earth instead of tapestry

wynnset:
## Changes
* Adds a "Geography Map" option that, when checked, shows a map of earth instead of the tapestry
* Map displays markers for each node and clicking on each marker will show a popup to view the node 
* When this options is checked and the user is logged in:
  * Show a list of nodes on the side for authors to add, edit, and place nodes on the map
  * Show a "Geography" tab on the node modal where the user can enter coordinates for that node
## Screenshots
image
image
image
## Issue Linkage
Closes #691 
## PR Dependency
Depends on: N/A
## Automated Testing
...


wynnset:
@yungalyx testing

wynnset:
@yungalyx testing 2

wynnset:
@yungalyx getting this error which is preventing the page from loading: 
```
ERROR in ./node_modules/leaflet/dist/leaflet.css 3:0
Module parse failed: Unexpected token (3:0)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| /* required styles */
|
> .leaflet-pane,
| .leaflet-tile,
| .leaflet-marker-icon,
 @ ./src/components/TapestryApp.vue?vue&type=script&lang=js& (./node_modules/babel-loader/lib!./node_modules/vue-loader/lib??vue-loader-options!./src/components/TapestryApp.vue?vue&type=script&lang=js&) 77:0-34
 @ ./src/components/TapestryApp.vue?vue&type=script&lang=js&
 @ ./src/components/TapestryApp.vue
 @ ./node_modules/babel-loader/lib!./node_modules/vue-loader/lib??vue-loader-options!./src/App.vue?vue&type=script&lang=js&
 @ ./src/App.vue?vue&type=script&lang=js&
 @ ./src/App.vue
 @ ./src/main.js
 @ multi @babel/polyfill ./src/main.js
```

yungalyx:
@wynnset could you try running 'npm install && npm run dev'?

wynnset:
@yungalyx that's what I did.

templates/vue/src/components/TapestryMap.vue
```diff
@@ -0,0 +1,89 @@
+<template>
+
+  <div style="height: 800px; width: 100%">
+    <div style="height: 150px overflow: auto;">
+      <p>First marker is placed at {{ withPopup.lat }}, {{ withPopup.lng }}</p>
+      <p> The bounds are: {{ mapBounds }} </p>
+      <p> Settings are: {{this.settings}} </p>
```
wynnset, line 7: You never use `.this` in the template. The template is bound to `this` already so the variables you use in it are assumed to be properties of `this` already. So this should be just `settings`.

templates/vue/src/components/TapestryMap.vue
```diff
@@ -0,0 +1,89 @@
...
+
+export default {
+  name: "Tapestry Map",
+  components: {
+    LMap,
+    LTileLayer,
+    LMarker,
+    LPopup,
+    LTooltip
+  },
```
wynnset, line 42: If you are not using `LMarker`, `LPopup`, and `LTooltip`, do not include them here (or comment them out for your draft PR). That way, linter will not complain

templates/vue/src/components/TapestryMap.vue
```diff
@@ -0,0 +1,89 @@
...
+        :url="url"
+        :attribution="attribution"
+
+      />
+    </l-map>
+  </div>
+</template>
+
+<script>
+import { latLng, latLngBounds } from "leaflet";
```
wynnset, line 30: This would usually get caught by the linter, but in our code, we drop the semicolon at the end of each line.

templates/vue/src/components/TapestryMap.vue
```diff
@@ -0,0 +1,89 @@
...
+      this.mapBounds = x;
+    }
+  },
+  methods: {
+    zoomUpdate(zoom) {
+      this.currentZoom = zoom;
+    },
+    centerUpdate(center) {
+      this.currentCenter = center;
+    },
```
wynnset, line 80: Please name your methods with the verb at the beginning, then the subject. So these would be `updateZoom` and `updateCenter`.

templates/vue/src/components/TapestryMap.vue
```diff
@@ -0,0 +1,89 @@
...
+    <l-map
+      :zoom="zoom"
+      :center="center"
+      :options="mapOptions"
+      :max-bounds="mapBounds"
+      :bounds='setBounds'
+      style="height: 80%"
+      @update:center="centerUpdate"
+      @update:zoom="zoomUpdate"
+
```
wynnset, line 18: This would also get caught by the linter usually, but there is no need for this empty line here.

templates/vue/src/components/TapestryMap.vue
```diff
@@ -0,0 +1,89 @@
...
+      withPopup: latLng(47.41322, -1.219482),
+      mapBounds: latLngBounds([
+        [90, 180],
+        [-90, -180]
+      ]),
+      withTooltip: latLng(47.41422, -1.250482),
+      showParagraph: false,
+      mapOptions: {
+        zoomSnap: 0.5
+      },
```
wynnset, line 59: You have a few data elements that are not used. Make sure you review and remove them before finalizing your PR. If you think we might want to add/utilize some of them in the future, you can comment them out but if you do comment something out, leave a comment above it to say why it's commented out and what it could be used for.

wynnset (commented):
@yungalyx I left you some comments here to give you a few tips about coding standards and Vue in general. I know you are still working on this, but thought this might help.

wynnset:
@narendrasss might be able to help you out with the css include issue I reported up there - I think it might have something to do with webpack.

yungalyx:
@wynnset fixed linting bug, renamed variables (hopefully they're a bit better 😄), and improved the UX of adding map bounds in settings with form validation. Check it out!

yungalyx:
@wynnset can you try running this branch when you have time? css issues should be fixed, thanks!

templates/vue/src/components/SettingsModal.vue
```diff
@@ -192,11 +247,19 @@ export default {
...
   },
   computed: {
     ...mapGetters(["tapestryJson"]),
     ...mapState(["settings", "rootId"]),
+    latbound() {
+      return (this.mapBounds.neLat || 90) > (this.mapBounds.swLat || -90)
+    },
+    lngbound() {
+      return (this.mapBounds.neLng || 180) > (this.mapBounds.swLng || -180)
+    },
```
wynnset, line 262: `latbound` and `lngbound` have 2 problems in their naming:
1. `lat` and `bound` are 2 different words and you should be using camelCase, as previously mentioned.
2. `latbound` is not explaining exactly what this variable does. You should name it something that is easy to understand like `latitudeRangeInvalid` or something like that. This will help other devs understand better what that variable is without having to look at its implementation.

templates/vue/src/components/TapestryApp.vue
```diff
@@ -50,25 +50,31 @@
...
 import TapestryNode from "@/components/TapestryNode"
+import TapestryMap from "@/components/TapestryMap"
 import TapestryLink from "@/components/TapestryLink"
 import TapestryDepthSlider from "@/components/TapestryDepthSlider"
 import SettingsModalButton from "@/components/SettingsModalButton"
 import RootNodeButton from "@/components/RootNodeButton"
 import LockedTooltip from "@/components/LockedTooltip"
 import TapestryFilter from "@/components/TapestryFilter"
 import Helpers from "@/utils/Helpers"
+import "leaflet/dist/leaflet.css"
```
wynnset, line 71: I don't know if it's a good idea to load the leaflet css file here. What if we don't need it? That means it will load all the time even if we don't need it? Maybe a solutions like what's suggested [here](https://stackoverflow.com/questions/52767025/how-to-load-css-file-dynamically) would work. @narendrasss what do you think?

wynnset (commented):
Good start. The map looks fine now and is loading for me. A few things: 
- I can't seem to be able to pan the map. It just bounces back every time I drag to move the map. 
- I think we should disable scroll to zoom.
- I think the coordinates that are being accepted are for north west and south east, even though you have labelled them as north east / south west. When I enter coordinates for north east and south west, it gives me an error and I have to change them to north west and south east.
Once those are fixed, the next step is to implement actually putting the nodes on the map as discussed. Thanks!

yungalyx:
@wynnset i moved the import css import statement from tapestryApp to tapestryMap.. would that maybe improve import so we only import if `<tapestry-map/>` is rendered? 

<code>In TapestryApp.vue:</code> 
`<div v-if="settings.renderMap" id="map-container"> 
      <tapestry-map />
    </div>`

wynnset:
Hi @yungalyx I don't think you addressed the items I mentioned in [my previous comment above](https://github.com/wynnset/tapestry-wp/pull/692#pullrequestreview-510152862) yet.

yungalyx:
the bounds won't let you pan beyond the coordinates! i added a red box to visually indicate that at the moment.
Scroll has been disabled and bounds will automatically dictate initial zoom level. 
map bound
Screenshot 2020-10-20 114557

wynnset:
@yungalyx 
- [ ] The user should still be able to pinch to zoom in/out
- [ ] The zoom icons (+/-) have disappeared since you moved the css import
- [x] Scroll to zoom is no longer working which is great
- [ ] I'm not sure what the purposes of the red line is... We should fix that issue instead. Is there no setting in leaflet to change that? It seems like a very weird behaviour. Maybe you are approaching it the wrong way? We could have a call to try and resolve this together if you want.
- [ ] I still cannot map nodes

yungalyx:
Still a few things to incorporate from my point of view:

- [ ] map nodes shouldn't show based on permission settings

- [ ] changes to node map coordinates are not rendered into map unless we refresh the page (i populate the marker array on create ) i.e, no 'live' updates

jaskwong:
- when dragging to move the map, this activates drag select 
- it seems that only text nodes can be mapped (for video, h5p, external link, etc. it doesn't show the option to specify coordinates after checking the "show this node on map" box)
- at the ends of the map (ex. at the very top), the map can still be dragged but this just shows the grey background behind it, maybe it would be better at the map bounds to not let it be dragged any further (this is just my opinion, but currently the map can basically be dragged out of view)
- based on the asana task, "we will hide the tapestry of nodes/links...simply list the nodes on the side of the map" so this is probably something to incorporate

I know this is still a draft pr so I apologize in advance if any of this is not currently applicable

templates/vue/src/components/TapestryMap.vue
```diff
@@ -0,0 +1,117 @@
...
+    resetMarkers() {
+      this.markers = []
+      for (const [key, value] of Object.entries(this.nodes)) {
+        if (this.hasMapCoordinates(value.mapCoordinates)) {
+          this.markers.push({
+            id: key,
+            pos: latLng(value.mapCoordinates.lat, value.mapCoordinates.lng),
+          })
+        }
+      }
```
narendrasss, line 113: With regards to your message on slack, I would maybe try this:

```suggestion
      const markers = []
      for (const [key, value] of Object.entries(this.nodes)) {
        if (this.hasMapCoordinates(value.mapCoordinates)) {
          markers.push({
            id: key,
            pos: latLng(value.mapCoordinates.lat, value.mapCoordinates.lng),
          })
        }
      }
      this.markers = markers
```

More than that, I would suggest changing markers from a data property to a computed property instead. Using a computed property will tell Vue to keep `markers` in sync with `this.nodes`, which means Vue knows to recalculate when `this.nodes` changes. This removes the duplication of code in `resetMarkers` and `created`, and removes the need for the event bus (which I thought was removed?)

yungalyx:
@jaskwong yeah i'm getting an error with the "show this node on map" error too. I'm not sure why because sometimes when I refresh the page, it'll work again..

jaskwong:
@yungalyx do you want me to take a look at it?

yungalyx:
@jaskwong sure! that would be super helpful, thank you!

jaskwong:
@yungalyx I think it's cause for new nodes, node.mapCoordinate isn't actually defined and in CoordinateForm.vue when created() tries to access the lat/lng coordinates it produces an error and same with the b-form-input which stops it from displaying, are you setting node.mapCoordinates for nodes, other than in the initial loading of the tapestry?

A tip for debugging: try opening the console in your browser inspect tool, errors are usually logged there so you can see what they say!

templates/vue/src/components/NodeModal.vue
```diff
@@ -257,6 +268,9 @@ export default {
         ? true
         : wpData.wpCanEditTapestry !== ""
     },
+    showMap() {
+      return this.settings.renderMap
+    },
```
wynnset, line 273: I think it would be better if you put `settings.renderMap` directly in your `v-if`. No reason to make the reader go searching for where `showMap` comes from.

templates/vue/src/components/SettingsModal.vue
```diff
@@ -109,6 +109,45 @@
               </b-col>
             </b-row>
           </b-form-group>
+          <b-form-group
+            label="World map"
+            description="Replace Tapestry with a new authoring experience over a map of the Earth"
```
wynnset, line 114: This should read "Replace Tapestry with a map of Earth with placeholders for each node."

templates/vue/src/components/SettingsModal.vue
```diff
@@ -109,6 +109,45 @@
...
+              {{ renderMap ? "Enabled" : "Disabled" }}
+            </b-form-checkbox>
+          </b-form-group>
+          <div v-if="renderMap">
+            <p v-if="!latitudeRangeValid" class="error-message">
+              North coordinate must be greater than south coordinate
+            </p>
+            <p v-if="!longitudeRangeValid" class="error-message">
+              East coordinate must be greater than west coordinate
+            </p>
```
wynnset, line 126: As mentioned before (but not yet done) please use Bootstrap-Vue provided functionality for showing errors. See https://bootstrap-vue.org/docs/components/form-input#contextual-states

templates/vue/src/components/SettingsModal.vue
```diff
@@ -109,6 +109,45 @@
...
+            <b-row>
+              <b-col sm="3">
+                <span>Northeast bound:</span>
+              </b-col>
+              <b-col sm="3">
+                <b-form-input v-model="mapBounds.neLat" placeholder="90" />
+              </b-col>
+              <b-col sm="3">
+                <b-form-input v-model="mapBounds.neLng" placeholder="180" />
+              </b-col>
```
wynnset, line 136: Please label which one is latitude and which one is longitude.

templates/vue/src/components/SettingsModal.vue
```diff
@@ -109,6 +109,45 @@
...
+            <b-row>
+              <b-col sm="3">
+                <span>Southwest bound:</span>
+              </b-col>
+              <b-col sm="3">
+                <b-form-input v-model="mapBounds.swLat" placeholder="-90" />
+              </b-col>
+              <b-col sm="3">
+                <b-form-input v-model="mapBounds.swLng" placeholder="-180" />
+              </b-col>
```
wynnset, line 147: Please label which one is latitude and which one is longitude.

templates/vue/src/components/SettingsModal.vue
```diff
@@ -213,12 +255,21 @@ export default {
...
   computed: {
     ...mapGetters(["tapestryJson"]),
     ...mapState(["settings", "rootId"]),
+    // the + before this.mapBounds changes type from string to int for correct comparison
+    latitudeRangeValid() {
+      return (+this.mapBounds.neLat || 90) > (+this.mapBounds.swLat || -90)
+    },
+    longitudeRangeValid() {
+      return (+this.mapBounds.neLng || 180) > (+this.mapBounds.swLng || -180)
+    },
```
wynnset, line 272: You probably also want to check that latitude and longitude are valid values / ranges. I entered `1000` and it accepted it!

templates/vue/src/components/TapestryMap.vue
```diff
@@ -0,0 +1,126 @@
...
+        style="height: 80%"
+        @update:center="updateCenter"
+        @update:zoom="updateZoom"
+      >
+        <l-tile-layer :url="url" :attribution="attribution" />
+        <l-marker
+          v-for="marker in markerlocations"
+          :key="marker.id"
+          :lat-lng="marker.pos"
+          :icon="marker.access ? icon : inaccessibleIcon"
```
wynnset, line 17: Can you also add a separate icon for draft nodes? It would be nice if it follows the look and feel of draft nodes (i.e. dashed outline)

templates/vue/src/components/TapestryMap.vue
```diff
@@ -0,0 +1,126 @@
...
+        [
+          +this.settings.mapBounds.neLat || 90,
+          +this.settings.mapBounds.neLng || 180,
+        ],
+      ])
+      return x
+    },
+    markerlocations() {
+      const markers = []
+      for (const [key, value] of Object.entries(this.nodes)) {
```
wynnset, line 94: I suggest naming your variables more clearly like `const [nodeId, node]`.

templates/vue/src/components/TapestryMap.vue
```diff
@@ -0,0 +1,126 @@
...
+        </l-marker>
+      </l-map>
+    </div>
+  </div>
+</template>
+
+<script>
+import "leaflet/dist/leaflet.css"
+import mapMarker from "@/assets/map-marker.png"
+import invalidMarker from "@/assets/map-marker-inaccessible.png"
```
wynnset, line 36: Could you change this icon and put a lock in it instead of the dot?

templates/vue/src/components/TapestryMap.vue
```diff
@@ -0,0 +1,126 @@
...
+        <l-tile-layer :url="url" :attribution="attribution" />
+        <l-marker
+          v-for="marker in markerlocations"
+          :key="marker.id"
+          :lat-lng="marker.pos"
+          :icon="marker.access ? icon : inaccessibleIcon"
+        >
+          <l-popup>
+            <div v-if="marker.access" @click="openNode(marker.id)">
+              {{ marker.title }}
```
wynnset, line 21: Can you add a style for this and set `cursor: pointer;` so it appears more like a link? Maybe also make it bold. (This does not apply for the inaccessible nodes)

templates/vue/src/components/node-modal/CoordinateForm.vue
```diff
@@ -0,0 +1,77 @@
...
+      <b-form-group>
+        <b-form-checkbox v-model="isOnMap">
+          Show this node on map
+        </b-form-checkbox>
+      </b-form-group>
+      <b-form-group v-if="isOnMap">
+        <b-form-input
+          v-model="node.mapCoordinates.lat"
+          :number="true"
+          placeholder="enter latitude coordinate here"
```
wynnset, line 13: Don't forget to always capitalize. You could also shorten it a bit like "Enter latitude".

templates/vue/src/components/node-modal/CoordinateForm.vue
```diff
@@ -0,0 +1,77 @@
...
+        />
+
+        <b-form-input
+          v-model="node.mapCoordinates.lng"
+          :number="true"
+          placeholder="enter longitude coordinate here"
```
wynnset, line 19: Don't forget to always capitalize. You could also shorten it a bit like "Enter longitude".

templates/vue/src/components/node-modal/CoordinateForm.vue
```diff
@@ -0,0 +1,77 @@
...
+  },
+  watch: {
+    /*
+    nodeLat(){
+      this.node.mapCoordinates.lat = this.nodeLat
+    },
+    nodeLng(){
+      this.node.mapCoordinates.lng = this.nodeLng
+    },
+    */
```
wynnset, line 52: Might want to remove this if unused

wynnset:
Hi @yungalyx we are almost here! I left you some code comments, but also noticed you have not addressed some comments from @jaskwong above, so I'm assigning back to you. Should be easy fixes. Please check off or mark as resolved when you address each point so we can keep track. Thanks!

templates/vue/src/components/NodeModal.vue
```diff
@@ -97,6 +98,14 @@
             </slick-list>
           </div>
         </b-tab>
+        <b-tab
+          v-if="showMap"
+          title="Coordinate"
```
wynnset, line 103: Please change the tab to be called "Map". Please also change the variable names to be plural i.e. `coordinates` and the component to `coordinates-form`.

yungalyx:
@jaskwong Sean's 809 PR fixes the dragSelect issue on my map! I will start the doing the nodes on the side of map tomorrow, thanks again for the help

yungalyx:
@wynnset everything should be fixed, i've included a list to add nodes alongside the map. Only no icons for draft nodes yet because I couldn't find any good ones that were free. All the nice looking ones need a license :0

wynnset:
Thanks @yungalyx. The nodes should not appear if the user cannot edit the settings for the tapestry. I'm working on the placeholder icons for you and will update soon.

templates/vue/src/components/TapestryMap.vue
```diff
@@ -0,0 +1,324 @@
...
+      draftIcon: icon({
+        iconUrl: draftMarker,
+        iconSize: [32, 33],
+        iconAnchor: [16, 32],
+        popupAnchor: [0, -20],
+      }),
+      worldBounds: latLngBounds([
+        [-90, -180],
+        [90, 180],
+      ]),
```
narendrasss, line 141: Do these attributes change? Typically `data` attributes are reserved for those that change throughout the component lifecycle. Otherwise you would put them under `computed`. Part of the reason is performance, but it also shows people who are reading the code that it won't change.

templates/vue/src/components/TapestryMap.vue
```diff
@@ -0,0 +1,324 @@
...
+
+    > div {
+      height: 80%;
+    }
+  }
+}
+</style>
+
+<style lang="scss">
+#map-container .nodes-list > .card .card-body table tbody > tr {
```
narendrasss, line 313: This selector looks very specific - not super clear to me what element(s) this selector is actually affecting. If you're overriding bootstrap values, could a [scoped slot](https://bootstrap-vue.org/docs/components/table#custom-data-rendering) work better for this use case?

templates/vue/src/components/TapestryMap.vue
```diff
@@ -0,0 +1,324 @@
...
+    },
+    selectedId() {
+      return this.$route.params.nodeId
+    },
+    isLoggedIn() {
+      return isLoggedIn
+    },
+    canEdit() {
+      return wpData.wpCanEditTapestry === "1"
+    },
```
narendrasss, line 157: Can you change this to use the [wp module](https://github.com/wynnset/tapestry-wp/blob/master/templates/vue/src/services/wp.js) instead for consistency?

templates/vue/src/components/TapestryMap.vue
```diff
@@ -0,0 +1,324 @@
...
+import "leaflet/dist/leaflet.css"
+import mapMarker from "@/assets/map-marker.png"
+import invalidMarker from "@/assets/map-marker-inaccessible.png"
+import draftMarker from "@/assets/map-marker-draft.png"
+import { latLng, latLngBounds, icon } from "leaflet"
+import { LMap, LTileLayer, LMarker, LPopup } from "vue2-leaflet"
+import { mapState } from "vuex"
+import { names } from "@/config/routes"
+import Helpers from "@/utils/Helpers"
+import { isLoggedIn } from "@/utils/wp"
```
narendrasss, line 95: This import may not be necessary after the change to the `wp` module.

narendrasss (changes requested):
A few nitpicks, but otherwise looks good!

templates/vue/src/components/TapestryMap.vue
wynnset: There's only one table element in there. Not sure if the scoped slot can do what is intended as we are using the `stacked` table. This table doesn't provide much in terms of built-in modifiers, so it's necessary to make these minor modifications here to match the intended styling.

wynnset:
@yungalyx could you please add automated tests to this?

rlnsy:
@yungalyx @wynnset This works for me! Anything in particular I should test? I succesfully created a new video Node and placed it on the map. Also, looking at the merge conflicts they appear to just be indentation related?

yungalyx:
@rlnsy awesome! a lot of manual testing was done so I'm glad it works well. Only concern at the moment is how much automated tests need to be written before merge. thoughts? @wynnset

narendrasss:
@yungalyx ideally everything that you're manually testing should be automated. I also like to add in edge cases in the form of integration tests whenever that's applicable

templates/vue/debug.log
```diff
@@ -0,0 +1,4 @@
+[1111/155015.365:ERROR:directory_reader_win.cc(43)] FindFirstFile: The system cannot find the path specified. (0x3)
+[1112/111843.045:ERROR:directory_reader_win.cc(43)] FindFirstFile: The system cannot find the path specified. (0x3)
+[1113/123327.742:ERROR:directory_reader_win.cc(43)] FindFirstFile: The system cannot find the path specified. (0x3)
+[1113/140254.460:ERROR:directory_reader_win.cc(43)] FindFirstFile: The system cannot find the path specified. (0x3)
```
wynnset, line 4: @yungalyx what is this file and why is it committed? I generally avoid using `git commit -a` because of this.

wynnset:
> @yungalyx ideally everything that you're manually testing should be automated. I also like to add in edge cases in the form of integration tests whenever that's applicable

I agree with @narendrasss - here are a few more things that I think could be added (feel free to add more if you think of some):
- [ ] Ensure changing map coordinate settings change the coordinates
- [ ] Ensure adding a node to map works
- [ ] Ensure deleting a node from map works
- [ ] Ensure editing the coordinates for a node updates its position on the map
- [ ] Ensure clicking on a pop-up link opens the node

PR #695, opened 2020-08-19
694 - Change neighbours calculation

### LINKED ISSUE #694: Change _recursivelySetAccessible neighbours calculation
This was changed on live to fix some problems with tapestries loading.

wynnset:
This was changed on live to fix some problems with tapestries loading.

Closes #694

### DIFF
--- a/classes/class.tapestry.php
+++ b/classes/class.tapestry.php
@@ -343,16 +343,15 @@ class Tapestry implements ITapestry
         if ($node->accessible) {
             $neighbourIds = $this->_getNeighbours($node);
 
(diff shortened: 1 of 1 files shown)


wynnset:
@narendrasss we made this change on live at some point and I think it was because of some issues with tapestry not loading (not too sure). Do you remember? I was just comparing live with master and this was one of the differences.

narendrasss:
@wynnset This change is basically the same change as #659. I realize that there were a few cases outside of TYDE where a Tapestry could crash so I added this guard in.

PR #702, opened 2020-08-21
701 - Move cypress and app out of docker containers

narendrasss:


narendrasss:
Closing this as this will be integrated as part of #705.

PR #705, opened 2020-08-24
653 - Improve Docker setup and add CI [WIP]

### LINKED ISSUE #653: Integrate Cypress to run automatically on Github

narendrasss:
Closes #653 

## What?

This PR completely overhauls the current docker setup so that:

- **Debugging is easier** by making Cypress no longer a docker container but rather part of the app
- **E2E tests run as part of Github Actions** (and runs in a simpler, more deterministic manner)
- **Onboarding is faster** through automated Wordpress setup

> This PR is meant as a replacement for #646 

## Rationale

- Cypress needed to be extracted out of docker for easier test debugging and coverage reports
- The previous github actions PR wasn't deterministic; WP installation doesn't always seem to work
- I remember encountering a lot of setup issues when Cody and Sean was on-boarding. This PR also automates setup for the local dev environment so that the new hires (hopefully) won't run into much problems

## Notes

- The directory structure was changed to make the new docker setup work. All plugin related files are now moved to a folder called `/plugins` and the root folder now contains scripts and github workflows (@wynnset Let me know if this is something you don't like -- I can maybe figure out an alternative way to keep the old structure).
- I'm running into some permissions issues so that H5P and GF cannot be installed. A workaround to this would be to commit the plugin folders themselves (not ideal).
- Not all of the tests are passing. There seems to be some bugs on master causing some of the tests to fail (a bunch of console errors when the Tapestry is empty):

Screen Shot 2020-08-24 at 14 12 35

(Naturally any tests related to H5P and GF would fail as well)

## Testing

- In the root directory, run `sh ./bin/start.sh`. This will spin up the docker containers and start the Vue app server.
  - This will create two directories: `wp-app` and `wp-data`. These directories contain the files for the wordpress installation and the wordpress db respectively.
- Give it some time for the containers to be ready - this may take up to a minute. To check if it's ready, visit `localhost:8000` on your browser. If the containers are ready, you should be greeted with a wordpress installation page. **Don't proceed with the installation**.
- Instead, go back to the root directory, open a new terminal window and run `sh ./bin/install.sh`. This will install all necessary plugins, configure the theme and other WP settings necessary for Tapestry to work.
- Once that's done, visit `localhost:8000` in your browser again. If everything works you should see a webpage with the title "Tapestry".
- Go to `/wp-admin` and login with the admin user. The username is `admin` and the password is `secret`.
- Double check that everything works. Go to plugins and check that both H5P and Tapestry is installed and activated.
- Make a new Tapestry and ensure you don't run into any issues there.
- To shut down the server, go to the original terminal window where you ran `start.sh`. Here, press `Ctrl + C` to quit the server. Then run `docker-compose down`.


narendrasss:
Split this up into #706 and #707 while keeping folder structure the same to minimize conflicts.

PR #711, opened 2020-08-29
708 - accept/reject: Node status filter

### LINKED ISSUE #708: Submit/accept/reject - Update search bar to filter by node status
Currently the search bar filters only by the node author. We want to update this so that the author is able to filter by the status of nodes (i.e. submitted/approved/rejected)

seanRong:
Gives the ability to filter all nodes by their status to any admin or reviewer through the search bar, shows available options according to existing nodes (ex. "submitted" will not show as an option where there are no submitted nodes). 

Closes #708

Depends on 680


templates/vue/src/components/TapestryFilter.vue
```diff
@@ -110,6 +121,10 @@ export default {
       const tapestryApi = new TapestryApi(wpPostId)
       this.allContributors = await tapestryApi.getAllContributors()
     }
+    this.allStatuses = new Set()
+    for (let node of Object.values(this.nodes)) {
+      this.allStatuses.add(node.status)
+    }
```
narendrasss, line 127: I'm thinking it might be better to hard code the statuses here since it's only going to be "submitted", "rejected", "draft" or something. Maybe move the status to a constants file?

narendrasss (changes requested):

templates/vue/src/components/TapestryFilter.vue
seanRong: OK hear me out, users will be confused if say "submitted" is an option, but no nodes are "submitted", so it shows all nodes faded out. Unless you want that option to be greyed out if none match - in which case you still have to do this set thing.

narendrasss: I think that's fine, since if all nodes are faded out it also explicitly shows that there's no nodes submitted. What do you think @wynnset?

seanRong: I think it's a bit harder to tell when the only valid ones are grandchildren

wynnset: Let's:
- Store the statuses in a constants file
- Still loop through the nodes to find how many nodes have each status 
- Display that number in a bracket next to each status (e.g. "Submitted (0)")
- If the number is equal to 0, let's disable the option so the user cannot select it and also visually grey it out or something

shirleywbi (commented):
Questions:
- Given an existing node, what is the expected behaviour when you click save as private draft? Should it completely remove the node from public view or should the old version of the node remain?

Issues:
- Adding a node initially shows different buttons "Submitting to Administrator for Review" then visibly converts to "Publish".
- Creating a published node, changing it to "save as private draft", then publishing does not publish the node.
- Upon clicking publish, a modal appears with "Submitting to Administrator for Review" for a short period of time and modal disappears before you have enough time to do anything. If you don't click "Submitting to Administrator for Review" in time, there are no visible changes to the node. If you do click it, it shows "Approve Node" and "Reject Node".
- Rejected nodes are visible to public.
- Search Bar not visible for public user but not sure if part of scope of 708.
- Rejected in filter shows up as NaN
708-Filter
- Publish button overflowing to next row
708-Styling
- Filter bar counts are only updated upon refresh
- Console.logs still in code.

wynnset:
> * Given an existing node, what is the expected behaviour when you click save as private draft? Should it completely remove the node from public view or should the old version of the node remain?

The "Save as private draft" button should only be displayed for new nodes (not yet saved ever) and nodes that are already in draft status. Otherwise (e.g. if the node is not draft) the button should not be displayed at all. In other words, once a node is published, it should remain published and cannot be converted back to draft.

seanRong:
Fixes:
Console logs removed
Used vuex to listen for new nodes and update counts
NAN fixed for reject
It's intentional that the public cannot use the filter, it's only for admins

The rest of the issues, I believe should be applied to one of the previous PRs instead of this one in particular as this one focuses on the filter itself.

P.s. Not able to replicate the rejected visible to public issue.

jaskwong:
Admins cannot see nodes that they previously rejected (but since the filter is only for admins, the reject filter will just always come up empty) - I'm assuming the intended behavior was to show nodes they previously rejected

seanRong:
Would like some input to above.
Here's the thing: if there's going to be a lot of submitted nodes, and the admin rejects a large majority of them, they will all still show up on their screen. It could get really crowded. 

In the same vein having a lot of submissions at once will also crowd the screen. IIRC in the spec for the final accept/reject task there's nothing about limiting number of shown submitted nodes either.

wynnset:
How about this: By default, let's hide any rejected nodes; however, if the user filters the nodes to show rejected nodes, we should then show the rejected nodes.

seanRong:
Implementation completed as above. 
Is fairly efficient as well.

shirleywbi:
Finding it a bit odd that while I am in the cell, I can't read the instructions below b/c of the dropdown.
image

Also noticed its dependent on 680 so will wait for that for further testing.

seanRong:
@shirleywbi I think this occurs even for the version on master, autofill gets in the way all the time.

@wynnset  separate ticket?

seanRong:
I addressed the dropdown issue in #796.

jaskwong:
Tested, works well!

yungalyx:
tested in effort to recreate issue from 796 (disable autofill).. 708 works great!

wynnset:
Waiting for #683 to be done first before this

wynnset:
#683 has been merged in. We are good to start again on this PR. Given the large number of merge conflicts, I suggest creating a new branch from master, adding the changes needed manually, and then force pushing to this PR (or closing this PR and opening a new one).

seanRong:
Agreed, but do you also think there's an argument for putting it on top of #847 ?

wynnset:
Sure. That's a good idea.

narendrasss:
@wynnset @seanRong Is there anything this PR covers that #847 doesn't? I was thinking to make both the same PR.

wynnset:
This is a dead PR as its changes are being implemented in #847. Will wait to close this PR until #847 is merged.

wynnset:
Actually going to close this PR now since it will still be available for reference after closing @seanRong @narendrasss

PR #722, opened 2020-09-10
[No ticket] Fix hot reload not working in 482

narendrasss:
After #482 was merged in, hot reload was broken because I hadn't correctly migrated some parts of the old webpack config over to webpack 4. This PR fixes these issues.


wynnset (approved):

seanRong:
Works, happy it's back. We even have named modules now yay

PR #732, opened 2020-09-13
760 - Bug fix: YouTube video in accordion appears outside of accordion

seanRong:
Note: I didn't have as large of an overflow issue without this fix, but I do believe the overflow has been reduced, this might need to be tested on the test site as well.

### DIFF
--- a/templates/vue/src/components/lightbox/AccordionMedia.vue
+++ b/templates/vue/src/components/lightbox/AccordionMedia.vue
@@ -183,7 +183,7 @@ button[disabled] {
 
 .media-container {
   height: 100%;
-  overflow: scroll;
+  overflow: hidden;
   scrollbar-color: auto black;
   scrollbar-width: none;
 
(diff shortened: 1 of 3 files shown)


templates/vue/src/components/lightbox/AccordionMedia.vue
```diff
@@ -184,7 +184,7 @@ button[disabled] {
 
 .media-container {
   height: 100%;
-  overflow: scroll;
+  overflow: hidden;
```
wynnset, line 187: This is not working for me. When my accordion is longer than the page, it doesn't let me scroll down to see the rest of the accordion.

wynnset (changes requested):
See comment below. Additionally, my YouTube video is appearing too narrow: 
image

yungalyx:
cannot test because I'm stuck in loop during video content/ upload youtube link to node. ):

seanRong:
Tried it and it's working for me, what link did you use Alex?

Additionally, vue-youtube-embed module is giving red console errors like
```Error in callback for watcher "playerHeight": "TypeError: this.player.setSize is not a function"
```

But I'm not sure if we really want to update that module?

shirleywbi (approved):
Looks good.
image

wynnset (approved):

PR #745, opened 2020-09-16
738 - Public users should not be able to add or edit

### LINKED ISSUE #737: Unable to add a connection as a user

### LINKED ISSUE #738: Public user can be given permission to add or edit, but cannot actually add

narendrasss:
Closes #738 

This PR removes support for node creation by public users. Specifically:
- The add button will now _always_ be hidden for public users, regardless of the permission set in the node
- The "add" permission checkbox for public users is now disabled, so authors won't be able to set this permission for future nodes
- For current nodes, the add permission is stripped when the node is loaded in

### DIFF
--- a/endpoints.php
+++ b/endpoints.php
@@ -620,8 +620,13 @@ function deleteTapestryLink($request)
         if ($postId && !TapestryHelpers::isValidTapestry($postId)) {
             throw new TapestryError('INVALID_POST_ID');
         }
+        if (!TapestryHelpers::userIsAllowed('ADD', $linkIndex->source, $postId)) {
+            throw new TapestryError('ADD_NODE_PERMISSION_DENIED');
+        }
+        if (!TapestryHelpers::userIsAllowed('ADD', $linkIndex->target, $postId)) {
+            throw new TapestryError('ADD_NODE_PERMISSION_DENIED');
+        }
         $tapestry = new Tapestry($postId);
-
         return $tapestry->removeLink($linkIndex);
     } catch (TapestryError $e) {
         return new WP_Error($e->getCode(), $e->getMessage(), $e->getStatus());
--- a/endpoints.php
+++ b/endpoints.php
@@ -615,15 +615,17 @@ function addTapestryLink($request)
 function deleteTapestryLink($request)
 {
     $postId = $request['tapestryPostId'];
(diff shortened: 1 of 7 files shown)


shirleywbi:
Tested!
737: Connections are stable.
738: Can create connections between child and grandchild nodes and view other nodes.
Removal of support for node creation by public users: Looks good.

shirleywbi:
Does anything need to be changed server-side in terms of permissions?

wynnset:
Sorry to completely change the scope of this PR, but I feel like it's incomplete without everything else and we might as well address the root issue here. Please take a look at the Asana issue for this to understand and implement the full requirements.

yungalyx:
Public can delete links between nodes! 😱 This destroys access to other nodes. This affects the tapestry from admin side and doesn't fix on page reload either. Caught a big bug here!

shirleywbi (commented):
Issues:
* If there's an existing node where public has permissions to read and (add or edit), it unchecks read in the Permissions Table when updated.
* If there's an existing tapestry where public has permissions to read, add, and edit, the table keeps add and edit checked and does not allow you to uncheck.

All other behaviour looks good.

shirleywbi (changes requested):
New commit fixed existing individual nodes but looks like there's still an issue with the main settings permissions.

If I have a Tapestry with existing defaults set to allow Read/Add/Edit permissions for Public, the Tapestry still shows that the default is Read/Add/Edit. This is also the case if I create a new node.  
image

If I try to change the Tapestry Settings by deselecting Add/Edit, I can't reselect them.
ezgif com-gif-maker

shirleywbi (approved):
Looks good to me!

wynnset:
When "Authenticated" is checked for either the "Add" or "Edit" column, the checkboxes below should become disabled. It doesn't make sense to allow **_ALL_** authenticated users to have a certain permission, but then **_at the same time_** not allow a particular role to have that same permission. 

image

jaskwong:
@wynnset I think that's a separate issue ("Clicking on "Add" or "Edit" for "Public" or "Authenticated" rows doesn't disable permissions under it" on asana), this PR is just for preventing public users from adding/editing and also preventing unauthorized users from deleting links

wynnset (approved):

PR #747, opened 2020-09-16
746 - Bug fix: Adding child node makes it appear like a grandchild node

### LINKED ISSUE #746: Adding child node makes it appear like a grandchild node

narendrasss:
Closes #746 

Changes the default depth to 3 rather than `maxDepth + 1` so that nodes made off an empty Tapestry wouldn't appear as grandchild nodes. I believe this also better reflects the old functionality in `tapestry.js`.

### DIFF
--- a/templates/vue/src/components/TapestryDepthSlider.vue
+++ b/templates/vue/src/components/TapestryDepthSlider.vue
@@ -97,7 +97,7 @@ export default {
   methods: {
     ...mapMutations(["updateNode"]),
     setDefaultDepth() {
-      this.currentDepth = this.settings.defaultDepth || this.maxDepth + 1
+      this.currentDepth = this.settings.defaultDepth || 3
     },
     updateNodeTypes() {
       const depth = parseInt(this.currentDepth)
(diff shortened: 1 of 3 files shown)


seanRong:
Looks good to me, definitely prefer this implementation

yungalyx:
nice fix!

templates/vue/src/components/TapestryDepthSlider.vue
```diff
@@ -97,7 +97,7 @@ export default {
   methods: {
     ...mapMutations(["updateNode"]),
     setDefaultDepth() {
-      this.currentDepth = this.settings.defaultDepth || this.maxDepth + 1
+      this.currentDepth = this.settings.defaultDepth || 3
```
wynnset, line 100: `3` should be stored as a constant in the app or at minimum as a data variable. It's not a good idea to hard-code values in the code.

wynnset (changes requested):
Looks good just a small nit

narendrasss:
@wynnset I updated the implementation to fill the default depth with the constant if it's undefined in settings. This should also fix the other issue I encountered in #624.

wynnset (approved):

PR #749, opened 2020-09-17
736 - Bug fix: control+A selection doesn't work in node modal

seanRong:
- Adds functions to add and remove the dragSelect event listener programmatically
- Modularizes DragSelect to be a utility class
- Uses static functions of DragSelectModular to add or remove event listener when relevant modals open.

### DIFF
--- a/tapestry.php
+++ b/tapestry.php
@@ -4,12 +4,12 @@
  * Plugin Name: Tapestry
  * Plugin URI: https://www.tapestry-tool.com
  * Description: Custom post type - Tapestry
- * Version: 2.30.0-beta
+ * Version: 2.32.0-beta
  * Author: Tapestry Team, University of British Coloumbia.
  */
 
 // Used to force-refresh assets
-$TAPESTRY_VERSION_NUMBER = '2.30.0-beta';
+$TAPESTRY_VERSION_NUMBER = '2.32.0-beta';
 
 // Set this to false if you want to use the Vue build instead of npm dev
 $TAPESTRY_USE_DEV_MODE = true;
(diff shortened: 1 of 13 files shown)


shirleywbi:
Tested. Current behaviour allows select all with Ctrl+A with the media modal open. Should that be the case?
Otherwise, looks good.

seanRong:
Good question, is there a case where text inputs are needed when just the lightbox is open, not the node form itself?
I think gravity forms, but weren't we phasing that out?

shirleywbi:
We'll be phasing Gravity forms out when we create our own built-in replacements so we'll likely still have text inputs.

wynnset:
I think it's best we don't allow node selection if lightbox is open. That would totally be unexpected behaviour from the user's end. I can't see any case in which a user wants to select nodes when they are viewing or editing nodes. It should also be disabled if any other modal (e.g. settings modal) is open. Sorry to change the scope of this PR, but I think this is the right way of doing it.

shirleywbi (approved):
Tested. Looks good to me.

wynnset (changes requested):
Again, I wonder if we could use Vuex here instead of an event bus so we don't unnecessarily introduce a new mechanism. For example, I imagine we could maintain a Vuex state that indicates whether any modal is open or not. If one is open, then we can disable drag-select altogether? In the future we might add other things that we also might want to be affected by this state.

shirleywbi (approved):
Tested. LGTM!

narendrasss (approved):
Code LGTM with the modular changes we discussed!

wynnset (approved):

PR #750, opened 2020-09-17
739 - cannot select new nodes

seanRong:
Utilizes bus to add all new nodes to selectables array, also refactors TapestryApp.vue to support this implementation.


shirleywbi:
Tested. No issues found.

wynnset:
@seanRong / @narendrasss is it possible to use Vuex instead of an event bus for this? I prefer not to add an event bus to the app if we can use Vuex.

shirleywbi (approved):
LGTM. Working great. Also covers [736](https://github.com/wynnset/tapestry-wp/pull/749/files).

narendrasss (approved):
Code LGTM!

PR #758, opened 2020-09-20
753 - Fix and Improve WP Init Script

rlnsy:
Addresses: #753.
When merged, this PR should

- [x] ensure that the docker Wordpress Initialization script is able to find the wp installation and perform the install within a reasonable amount of time.
- [x] add a way to programmatically put API key in for config.js for LinkPreview

**Testing Instructions**:

0. Make sure Docker daemon is running.

1. In root project, run `make uninstall`. This will remove all data related to the WordPress setup, including pages and  users!

2. Run `make init`. This should load up the wordpress containers and then run the installation script. Keep an eye on the output until you see a message that WP was successfully installed. You should now be prompted to enter your LinkPreview key.

3. If it worked properly, run `make start-app` to start the tapestry application. Now, open up the application in a browser and ensure that external links are working properly (create an external link node with a working thumbnail such as a Wikipedia entry) , otherwise run `make stop` to bring down all containers and try increasing the container delay as described in the file`bin/config-sample.sh` (and run `make-init` again).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ start-app:
 
 init:
 	make start-wp
-	sh ./bin/await-containers.sh
+	bash ./bin/await-containers.sh
 	make install
 	make env
 	@echo "All done!"
--- a/Makefile
+++ b/Makefile
@@ -14,10 +14,14 @@ init:
 	# Change 30 to a higher number if it fails to install
 	@sleep 30
 	make install
+	make env
 
 install:
 	sh ./bin/install.sh
 	cd templates/vue && npm i
+
+env:
+	bash ./bin/setup-env.sh
 
 clear-db:
 	docker volume rm tapestry-wp_db_data
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ clear-db:
 uninstall:
 	make stop
 	make clear-db
+	cd templates/vue && rm -r node_modules
 
 stop:
 	make stop-app
--- a/Makefile
+++ b/Makefile
@@ -10,9 +10,6 @@ start-app:
 
 init:
 	make start-wp
-	@echo "Waiting for 30s for containers to be ready..."
-	# Change 30 to a higher number if it fails to install
-	@sleep 30
 	make install
 	make env
 
--- a/Makefile
+++ b/Makefile
@@ -10,15 +10,27 @@ start-app:
 
 init:
 	make start-wp
-	@echo "Waiting for 15s for containers to be ready..."
-	# Change 15 to a higher number if it fails to install
-	@sleep 15
+	@echo "Waiting for 30s for containers to be ready..."
+	# Change 30 to a higher number if it fails to install
+	@sleep 30
 	make install
 
 install:
+	@echo "Setting up WP installation:"
 	sh ./bin/install.sh
+	@echo "Installing plugin dependencies:"
 	cd templates/vue && npm i
+	# TODO: bring down docker stack at this point?
+
+clear-db:
+	docker volume rm tapestry-wp_db_data
+
+uninstall:
+	@echo "Stopping application:"
+	make stop
+	make clear-db
 
 stop:
+	# TODO: prevent exit if lsof output is empty
 	kill $$(lsof -t -i:8080) && docker-compose down
 	
--- a/Makefile
+++ b/Makefile
@@ -13,10 +13,17 @@ init:
 	sh ./bin/await-containers.sh
 	make install
 	make env
+	@echo "All done!"
 
 install:
+	make install-wp
+	make install-app
+
+install-wp:
 	sh ./bin/install.sh
-	cd templates/vue && npm i
+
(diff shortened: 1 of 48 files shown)


Makefile
```diff
@@ -10,15 +10,29 @@ start-app:
 
 init:
 	make start-wp
-	@echo "Waiting for 15s for containers to be ready..."
-	# Change 15 to a higher number if it fails to install
-	@sleep 15
+	@echo "Waiting for 30s for containers to be ready..."
+	# Change 30 to a higher number if it fails to install
+	@sleep 30
```
wynnset, line 15: Can we somehow set this as a variable in one of our environment files so that it's adjustable locally and doesn't get accidentally committed?

rlnsy: Definitely

rlnsy: I've opted to load the value from a local shell script that exports the variable - that seemed like the easiest way given that we are in a Make/Shell environment. I've included 30s as a default value since that seems to work for me.

rlnsy:
@seanRong do you wanna give this a go on your machine?

shirleywbi:
**Running with Windows WSL:**

1. Came across an issue with the double brackets in `await-container.sh` (https://stackoverflow.com/questions/9666102/bash-double-bracket-issue). Resolved it by running `bash ./bin/await-containers.sh` instead of `sh ./bin/await-containers.sh`. Same issue with `stop-app.sh`. Resolved the same way.

2. I was able to get to the "Success: WordPress installed successfully." Had to install `node` to run the `cd templates/vue && npm i` so that'll need to be added to the docs to do that before calling `make init`.

3. Do I need PHP installed locally or is the phpmyadmin container sufficient? No nodes are appearing, thus unable to test step 3.
image
Settings and plugin activation look good.

4. Ran `make uninstall`. Afterward, tried running `make init` but got: "Error: Error establishing a database connection. This either means that the username and password information in your `wp-config.php` file is incorrect or we can’t contact the database server at `db`. This could mean your host’s database server is down." Tried running `docker system prune --all` and `docker volume prune` which did not resolve this issue.

@rlnsy Thoughts on what might cause this?

rlnsy:
@shirleywbi I've added fixes for your comments 1 and 2 above. 

Wrt.3, the php container should be sufficient if I'm not mistaken. Also, the nodes should only appear on a tapestry page right? They show up for me once I create a tapestry post.

4: Did you run `uninstall` before you ran`init` the first time? Or did you experience the db problem only on the second run-through?

Thanks,
Rowan.

shirleywbi:
@rlnsy Thanks.
3. That screenshot was a tapestry post...
4. No, I didn't have anything set up so didn't feel the need to run `make uninstall`. I only experienced the db problem the second time. Order of operations: `init`, `uninstall`, `init` (db error).
Thoughts?

rlnsy:
@shirleywbi Ah! The nodes not appearing is simply a consequence of not running `make start-app` after init. The testing instructions have been updated.

I'll think about the db issue and get back to you - thanks for testing!

rlnsy:
@shirleywbi. Hey! After extensive testing on my own machine, they only way I've been able to reproduce the db issue is by setting my container wait time to a lower value. Hence, my suspicion is that the number must be increased on your machine. Can you try copying `bin/config-sample.sh` into `bin/config.sh` and increasing the delay to something really big, like 180 seconds?

shirleywbi:
Coming across issues with the shell script:
When running `make init`:
It can't seem to find the config file. Works with `bash` instead of `sh`.
```
sh ./bin/await-containers.sh         //Does this work with bash on your end?
./bin/await-containers.sh: 3: source: not found
Waiting  seconds for containers to be ready
sleep: missing operand
Try 'sleep --help' for more information.
make: *** [Makefile:13: init] Error 1
```

Error when running `init`, `uninstall`, then `init` **resolved**.
(Though as a note, I had to bump it up to 300s because it failed with 180s the second time).

When running `make start-app`:
```
ERROR in ./src/services/LinkPreviewApi.js
Module not found: Error: Can't resolve '../../config' in 'myfilepath/tapestry-wp/templates/vue/src/services'
```
It's trying to read the config file that no longer exists in LinkPreviewApi (line 2). We'll need to remove this line.

rlnsy:
@shirleywbi Will the wait script work on windows if we simply switch the interpreter to bash instead of sh?

shirleywbi:
@rlnsy Yes. I was only having issues with `sh ./bin/await-containers.sh`, but it worked fine as `bash ./bin/await-containers.sh`.

rlnsy:
@shirleywbi Made the changes. If you don't mind giving this one more spin when you get the chance then it's ready for review. - Thanks!

shirleywbi (approved):
Good to go!

narendrasss (approved):
LGTM!

wynnset:
I must agree with @narendrasss ... mad scripting skillz @rlnsy ! Well done.

PR #761, opened 2020-09-21
741 - Unable to open the settings modal more than one time

seanRong:
After opening the settings modal for the first time, subsequent clicks on the gear icon will not open the modal.
This rectifies that issue, as well as removing the error with mutating the 'show' prop directly that used to be bound to the b-modal visible state.


shirleywbi:
Tested. Looks good to me!

wynnset (approved):

wynnset:
@shirleywbi in the future, please also don't forget to do a code review when you are testing and if it looks good Approve the changes. Thanks!

PR #771, opened 2020-09-29
748 - Bug Fix: Node not displaying after changing connections

### LINKED ISSUE #748: Node not displaying after changing connections

shirleywbi:
Closes #748 

Updates _pathIsAllowed implementation to use breadth-first-search. It only uses cached _pathIsAllowed when it returns true. Otherwise, if a node did not have permission to be read, the cache would return false before looking at other possible paths.

### DIFF
--- a/classes/class.tapestry.php
+++ b/classes/class.tapestry.php
@@ -531,8 +531,8 @@ class Tapestry implements ITapestry
         $queue->enqueue($from);
         while (!$queue->isEmpty()) {
             $node = $queue->dequeue();
-            if (TapestryCache::exists(__METHOD__, [$node, $to, $superuser_override, $userId])) {
-                return TapestryCache::get(__METHOD__, [$node, $to, $superuser_override, $userId]);
+            if (TapestryCache::get(__METHOD__, [$node, $to, $superuser_override, $userId])) {
+                return true;
             }
             if (TapestryHelpers::userIsAllowed('READ', $node, $this->postId, $superuser_override, $userId)
                 || (TapestryHelpers::userIsAllowed('ADD', $node, $this->postId, $superuser_override, $userId)
--- a/classes/class.tapestry.php
+++ b/classes/class.tapestry.php
@@ -484,7 +484,7 @@ class Tapestry implements ITapestry
         if ($tapestry->settings->superuserOverridePermissions && $roles->canEdit($this->postId)) {
             return $tapestry;
         } else {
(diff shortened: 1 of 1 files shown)


wynnset:
I tested this new implementation with the following tapestry: https://ubc.tapestry-tool.com/psyc102/tapestry/psyc102-98a-summer-2020-tapestry/#/

The tapestry took around 2 minutes to load. With the original implementation, the same tapestry was loading in under 10s. Given we need to make sure things are optimized for very large tapestries, this implementation needs to be improved to work faster.

classes/class.tapestry.php
```diff
@@ -526,30 +526,34 @@ private function _filterNodeMetaIdsByPermissions($nodeMetaIds, $rootId, $superus
...
-            if ($from == $to) {
-                $pathIsAllowed = true;
-            } else {
-                $checked[] = $from;
+        $queue = new SplQueue();
+        $checked[] = $from;
+        $queue->enqueue($from);
+        while (!$queue->isEmpty()) {
+            $node = $queue->dequeue();
+            if (TapestryCache::get(__METHOD__, [$node, $to, $superuser_override, $userId])) {
```
wynnset, line 534: If you are just checking if something is cached, you should use the `exists` method, not the `get` method.

shirleywbi:
> I tested this new implementation with the following tapestry: https://ubc.tapestry-tool.com/psyc102/tapestry/psyc102-98a-summer-2020-tapestry/#/
> 
> The tapestry took around 2 minutes to load. With the original implementation, the same tapestry was loading in under 10s. Given we need to make sure things are optimized for very large tapestries, this implementation needs to be improved to work faster.

@wynnset Is this a tapestry I can get access to?

classes/class.tapestry.php
shirleywbi: Fair. This change is probably slowing it down because I'm only getting the cached value if it's true. I was coming across issues with a node without read permissions, caching as false and then a subsequent adjacent node retrieves that and thinks it is also false.
image
Will come back with a solution...

wynnset:
> @wynnset Is this a tapestry I can get access to?

I've created an account for you and you should have received an email about it.

shirleywbi:
Instead of doing a BFS on each node, I've reimplemented the filter to use a DFS starting from the root node (if it exists) and collected all the valid nodes. Also removed the cache since it now only traverses the graph nodes once.

wynnset (approved):

PR #779, opened 2020-10-02
777 - Update Cypress to work with new master

### LINKED ISSUE #777: Update Cypress to work on new master (post-482)

narendrasss:
Closes #777

- Updates the cypress tests so it passes after the #482 changes
- Removes unnecessary functions in `utils`
- Adds a new `data-qa` attribute that is used by selectors in Cypress
  - This is meant to be an incremental replacement to `data-testid` - **why?** because I can never remember if it's `data-testid` or `data-test-id` or `data-test-ID` or ...

**Note**

The tests caught a bug on `master` where clicking on the "Add Root Node" button doesn't do anything (addressed in PR #781). While testing this, if that test fails but everything else passes, then it's all good.

In general, if you want to check whether a test failing is because of a bug in the code or a bug in the test code, then you should try replicating the steps in the browser. If it works there, then it's probably a bug in the test code.


wynnset (approved):

rlnsy:
@narendrasss @shirleywbi  I'm getting the following trace:
```
TypeError: The following error originated from your application code, not from Cypress.

  > Cannot read property 'top' of undefined

When Cypress detects uncaught errors originating from your application it will automatically fail the current test.

This behavior is configurable, and you can choose to turn this off by listening to the `uncaught:exception` event.

https://on.cypress.io/uncaught-exception-from-application

Because this error occurred during a `before each` hook we are skipping the remaining tests in the current suite: `User
```
For 10 tests, and the rest are being skipped as a result. Is this consistent with other testing? If so do we need to update the branch?

narendrasss:
@rlnsy Is this `user.spec.js`? I didn't update that suite because it was testing things that were more or less covered by everything inside `authoring/`. Try deleting that file and see if it works.

rlnsy:
@narendrasss whoops I'm a goon and didn't actually start the vue app. I shall report back

rlnsy:
Unfortunately actually starting the app didn't change the outcome. However, it turns out that using the test GUI gives me this error instead:
```
Removing cookie failed for: {"url":"http://localhost/wp-content/plugins","name":"wordpress_cd9b744c619529c4988e0e94344eaf12"}

Because this error occurred during a before each hook we are skipping the remaining tests in the current suite: Node content
```

narendrasss:
@rlnsy It's possible the master merge broke the tests again. I'll take a look!

narendrasss:
@rlnsy Just ran the tests and they all pass on my end. I did update Cypress versions in this PR, can you try deleting `node_modules/`, reinstalling and rerunning the tests?

I also did a search for `sidebarPos` and I can't find any code with `sidebarPos` in our codebase - did Cypress say where that error came from?

rlnsy:
@narendrasss My apologies, I forgot to follow up after testing yesterday.  I was able to get **all tests to pass except for those in `user.spec.js`**., running the dev configuration alongside wordpress running in Docker. Two places where I ran into issues:
1. When running tests in the GUI, I still get some of those cookie saving errors (which might not cause tests to fail) and 7 tests fail which is more than I get when runnning from command line.
2. I can't seem to get my MAMP setup to work with the tests. Changing the port from 8000 seems to break things, and there seems to be some kind of redirect to port 8888 that is stumping me.

Conclusion: Tests work fine using docker and the Cypress CLI. Importantly, I would expect them to work once merged in with the Githubb testing workflow. Maybe you could try replicating the other issues just to make sure it's not user error on my end? Otherwise these could probably be addressed in later PRs.

narendrasss:
@rlnsy No worries, thanks!

1. It's strange that the tests fail when using the GUI. Can you provide some of the error messages that show up when using the GUI? Nevertheless, if it _consistently_ passes in headless mode, then I think it's not too big of a deal.
2. Strange, Cypress shouldn't be dependent on what your local configuration is. How were you changing the ports?

rlnsy:
@narendrasss Following up on my previous reply... I've done some more testing (including a fresh installation on a new machine) and this is what I have now:

0. Both of my machines (both Macs) pass all tests successfully in non-GUI mode.

1. When running with the GUI, all tests pass when using both Google Chrome and Electron, but when using Firefox cypress fails exactly 4 tests with the following error:
```
Removing cookie failed for: {"url":"http://localhost/wp-content/plugins","name":"wordpress_<string>"}

Because this error occurred during a before each hook we are skipping the remaining tests in the current suite: <suite name>
```
(And only passes 5 tests). Since this is happening in a hook, I don't have a way to try and replicate the error in Firefox. I should also note that Firefox is my personal browser and I frequently have an additional instance running on my machine while testing, but I also ran the tests with my personal Firefox closed and got the same errors.

2. When using MAMP, I have the following cypress configuration:
```
{
  "env": {
    "ADMIN_USERNAME": "admin",
    "ADMIN_PASSWORD": "secret",
    "SUBSCRIBER_USERNAME": "subscriber",
    "SUBSCRIBER_PASSWORD": "subscriber"
  },
  "baseUrl": "http://localhost:8000"
}
```
but when running cypress tests in any mode I get:
```
Error: connect ECONNREFUSED 127.0.0.1:8888
```
which is weird because the config is set to use port 8000.

Furthermore, when I visit the main wordpress dashboard at :8000 in a browser - I get redirected to port 8888 and get a connection refused! This is a non-critical issue but I would be curious to find out why it happens.

**Conclusion**:
Maybe you can run the tests in GUI mode with firefox and see what happens? Otherwise I think these can be addressed later as the tests are passing (the objective of this PR).

Thanks for reading!
Rowan

wynnset:
Master merge broke it

PR #783, opened 2020-10-02
782 - Bug fix: Zero depth makes all nodes grandchild nodes

### LINKED ISSUE #782: 0 depth makes all nodes grandchild nodes
When the current depth is 0 all nodes look like grandchild nodes (no progress bar, no title, no buttons).

The expected behaviour is that 0 depth turns off depth → all nodes should be visible.

narendrasss:
Closes #782 

This PR fixes a bug where setting depth to zero in the url makes all nodes look like grandchild nodes. To test:

1. Create a Tapestry with a reasonable depth (> 3)
2. Change the depth query parameter in the url to `depth=0` (if it's not there, add `?depth=0` to the url)
3. All nodes should be visible and the same size (except the selected node, which as regular behaviour should be larger)

### DIFF
--- a/templates/vue/src/components/TapestryDepthSlider.vue
+++ b/templates/vue/src/components/TapestryDepthSlider.vue
@@ -116,6 +116,23 @@ export default {
     ...mapMutations(["updateNode"]),
     updateNodeTypes() {
       const depth = parseInt(this.currentDepth)
+
+      if (depth === 0) {
+        const nodesToUpdate = Object.values(this.nodes).filter(
+          node => node.nodeType !== "child"
+        )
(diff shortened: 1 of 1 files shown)


shirleywbi (commented):
Blocked by 780 - Unable to create new tapestry. Looks fine using existing tapestry though.

shirleywbi (approved):
Tested this with the @780 root node fix. There's an issue with the depth slider but for the context of just the grandchild nodes, it's good.

shirleywbi:
@narendrasss The depth slider isn't changing the URL. I created a bug in Asana.

wynnset (approved):

PR #791, opened 2020-10-08
788 - Add missing Cypress tests to achieve better coverage

### LINKED ISSUE #788: Add missing Cypress tests to achieve full coverage

narendrasss:
Closes #788 

This PR improves code coverage by around ~18% from 43% to 61% statement coverage. It also refactors some of the tests to better adhere to best practices. Specifically:

- Minimize using the node modal to make changes to the node
  - Using the node modal takes time and leads to needless repetition. This new method means we test the modal happy path once, and other tests programmatically make updates to the nodes (through Vuex actions).
- Removes the `applyModalChanges` utility function so it's more obvious what we're actually testing

### Other Notes

- The test runs are approaching 10 minutes per run. I think we should look into integrating the [Cypress Dashboard](https://dashboard.cypress.io) and implementing [parallelization](https://github.com/cypress-io/github-action#parallel) to cut this time down.
- What kind of coverage are we looking to achieve? 
- I was using the "Given-When-Then" methodology to name tests, but I quickly found that it became tedious to create clear test names - especially since in E2E tests there are a lot of different assertions and events happening in one run. In one of the spec files, `settings.spec.js`, I used the normal convention but I didn't change the naming for the other tests that were already written.


narendrasss:
Look into programmatically adding H5P content.

templates/vue/src/utils/Helpers.js
```diff
@@ -137,6 +137,27 @@ export default class Helpers {
     return outObject
   }
 
+  static deepMerge(source, other) {
+    const out = { ...source }
+    for (const key in other) {
+      const value = other[key]
+      if (typeof value === "object" && !Array.isArray(value)) {
+        out[key] = Helpers.merge(out[key], value)
```
wynnset, line 145: Wouldn't it be better if this is also calling `Helpers.deepMerge`?

wynnset (changes requested):
Thank you for writing all these tests - they look really good! I skimmed this a bit as I think I'll need to look at it in more detail later, but I found just a couple of minor things you might want to look at, but otherwise this is good to get merged after we merge the base branch into master first. You can mark it as ready to merge once you have double-check the items below.

templates/vue/cypress/integration/settings.spec.js
```diff
@@ -0,0 +1,68 @@
+import { TEST_TAPESTRY_NAME } from "../support/constants"
+
+describe("Settings", () => {
+  beforeEach(() => {
+    cy.fixture("stump.json").as("tapestry")
+    cy.setup("@tapestry")
+    cy.getByTestId("settings-button").click()
+  })
+
+  it(`should be able to set default depth`, () => {
```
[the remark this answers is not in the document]
narendrasss: Yes, I mentioned this in the description:

> I was using the "Given-When-Then" methodology to name tests, but I quickly found that it became tedious to create clear test names - especially since in E2E tests there are a lot of different assertions and events happening in one run. In one of the spec files, `settings.spec.js`, I used the normal convention but I didn't change the naming for the other tests that were already written.

What are your thoughts?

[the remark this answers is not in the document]
wynnset: Oh sorry I didn't see that. I think it would make sense to use the same naming convention, so I suggest either changing the old names or keep following Given-When-Then. That said, if you think it takes too long and it doesn't matter for some reason, I trust your judgement.

wynnset:
@narendrasss please fix merge conflicts and reassign back to me.

templates/vue/cypress/integration/link-authoring.spec.js
```diff
@@ -1,74 +1,114 @@
...
-    `)
-
-    it(`
-      Given: An accordion node and a regular node
-      When: A link is added from the accordion to the node
-      Then: A confirmation should appear that, when cancelled, adds only the link
-    `)
+  beforeEach(() => {
+    cy.fixture("stump.json").as("stump")
+    cy.setup("@stump")
```
wynnset, line 4: Could we name this something less generic than `stump` in case in the future we want to add other types of tapestries? Something like `base-tapestry` or something?

templates/vue/cypress/integration/node-appearance.spec.js
```diff
@@ -1,65 +1,32 @@
...
-    Given: A Tapestry node
-    When: Its progress bar is hidden
-    Then: It should hide its progress bar
-  `, () => {
-    setupThenTest({ hideProgress: true }, node => {
-      cy.getByTestId(`node-progress-${node.id}`).should("not.exist")
-    })
-  })
+      const url =
+        "https://image.shutterstock.com/z/stock-photo-colorful-flower-on-dark-tropical-foliage-nature-background-721703848.jpg"
```
wynnset, line 19: Could we use a creative commons image (e.g. Wikimedia) and make sure the URL is reliable enough that it won't break in the future?

wynnset (commented):
I can make these updates, but just wanted to first get your feedback about it @narendrasss

templates/vue/cypress/integration/link-authoring.spec.js
narendrasss: There's already other types of tapestries under `fixtures`, I think this `stump` name refers to the fact it's a root node with two nodes so it's like a  "mini-tree" 😄

Feel free to rename it to something more appropriate though.

templates/vue/cypress/integration/node-appearance.spec.js
narendrasss: Yes that sounds good!

PR #797, opened 2020-10-14
796 - Disable autofill for filter

### LINKED ISSUE #796: Autofill gets in the way of using tapestry filter

seanRong:
Removes that pesky autofill getting in the way of selecting dropdown items when using the Tapestry filter.

Also, in the future we can just pass a prop into Combobox to disable autofill when needed.

closes #796


templates/vue/src/components/Combobox.vue
```diff
@@ -6,6 +6,7 @@
       :placeholder="placeholder"
       :size="size"
       :style="inputStyle"
+      autocomplete="autocomplete"
```
narendrasss, line 9: Is this supposed to be:
```suggestion
      :autocomplete="autocomplete"
```
Otherwise I think it's passing the literal string "autocomplete".

narendrasss (changes requested):

templates/vue/src/components/Combobox.vue
seanRong: Thank you!

rlnsy:
@seanRong any tips for reproducing the original problem - i.e. getting autofill to pop up?

jaskwong:
@rlnsy you can try checking out 708-node-status-filter and running a couple searches for status, then it should start to pop-up, if you can't recreate it let me know and I can test it since I have this issue

yungalyx:
I'm no longer getting the native autofill boxes (black dropdowns) 👍 but am I able to filter by anything other than just author atm?

seanRong:
No, 708 isnt merged in yet so only authir

rlnsy:
@seanRong can confirm that this eliminated the issue for me as well.

wynnset (approved):

PR #806, opened 2020-10-17
805 - add link error handling

jaskwong:
- catch errors returned from backend and display error message notification


jaskwong:
Right now, I just caught the error for link adding, should I add try/catch for the other requests, I wasn't sure about the scope of this issue on asana

wynnset:
Yes, please add it in for everything. It would also be nice if we show a user an alert saying what went wrong (I believe the backend sends an error message back in the response that we can show). In terms of implementation: 
- We might want to create a helper class for this so that we do not repeat code where possible
- It would be nice if we do not use alerts, but something less intrusive like [Bootstrap toasts](https://bootstrap-vue.org/docs/components/toast)

wynnset (commented):
Very elegant solution, @jaskwong good job! I think we could improve this by doing more than just showing the error if the request fails, but that's probably outside the scope here. The exception is areas where there is already something implemented - those should be transferred to `actions.js`. For example, where `updateNodeCoordinates` is being called, it is resetting the location of the node - this should be transferred to `actions.js` along with any other instances (if any) where we are handling the exception more than just showing the error.

templates/vue/src/utils/errorHelper.js
```diff
@@ -0,0 +1,27 @@
+    export default class ErrorHelper {
+    /**
+   * Return true if response is an error
+   *
+   * @param {Object} response
+   *
+   * @return {Boolean}
+   */
+    static isError(response) {
+        return response.status !== 200
```
rlnsy, line 10: @jaskwong Should we support other successful response codes in the 200 range?

jaskwong: good call, I'll update it, but right now this function is never called so I might delete it later

rlnsy:
@jaskwong From what I can see, video errors from the backend are not being caught with this setup.  @shirleywbi has been working on video URL validation in the frontend, but we should probably add a catch-all handle with the same structure as you have here. Idk if this is outside the scope though.

rlnsy:
Also, let me know if you want any help with the error-handling!

jaskwong:
@rlnsy I think that would probably need a new error structure, the error handling on this branch is primarily for errors that happen as a result of the REST api when we try to update the database, also thanks for the offer, I think most of the error structure here is finished, and there's just a couple bugs that need to be worked out

templates/vue/src/components/TapestryNode.vue
```diff
@@ -266,60 +268,52 @@ export default {
...
+              originalCoordinates.x == coordinates.x &&
+              originalCoordinates.y == coordinates.y
+            ) {
+              continue
             }
+            this.updateNodeCoordinates({
+              id,
+              coordinates,
+              originalCoordinates,
+            }).then(() => {
```
jaskwong, line 314: @narendrasss I'm not sure if you've had the chance to test yet, if not, do you mind changing this to async/await syntax I should've caught that earlier

narendrasss (approved):
Thanks @jaskwong this looks good! There was a small bug where the svg takes a long time to update when you drag nodes around. I pushed a small fix to address this.

Otherwise I think it's good to go 😄 We should make a note to add the appropriate tests for this in the near future.

wynnset (approved):

templates/vue/src/utils/errorHelper.js
wynnset: I've removed this function since I agree with @rlnsy and also it wasn't being used.

PR #810, opened 2020-10-19
809 - fixes for dragselect in sidebar and filter button

seanRong:
Why I feel like this is needed:

Dragselect activating when trying to use the filter is a bit annoying.
Dragselect activating also interferes with using sidebar at times.
You can also activate dragselect by clicking literally anywhere on the page (including in the wp admin bar at the top for example)

Lingering questions:
I tried to force dragSelect to only activate when clicking in a certain region, ie under the tapestry div. Thoughts?


templates/vue/src/utils/dragSelectModular.js
```diff
@@ -51,4 +51,13 @@ export default class DragSelectModular {
...
   }
+
+  static initializeDragSelectZone(mainElementId) {
+    let allowedArea = document.getElementById(mainElementId)
+    document.body.addEventListener("mousedown", function(event) {
+      if (!allowedArea.contains(event.target)) {
+        DragSelectModular.pauseDragSelect()
+      }
+    })
+  }
```
wynnset, line 62: Can't you just use the `area` parameter to set the drag area? https://thibaultjanbeyer.github.io/DragSelect/DragSelect.html

wynnset: Alternatively, I think there's also a `vue-drag-slider` package that we could use instead that automatically has a wrapper for the area.

seanRong: I decided to use your first suggestion!

yungalyx:
i think opening the sidebar may interfere with the coordinates of dragselect 
ezgif com-gif-maker

wynnset:
@seanRong just FYI this issue seems to be happening due to this line: https://github.com/wynnset/tapestry-wp/blob/b30a5e77663855a4568ae0833e7bdb1001d97615/templates/vue/src/components/TapestryApp.vue#L243

wynnset:
@yungalyx that issue should be fixed now. Can you test again please?

jaskwong (approved):

jaskwong:
Tested works great!

@yungalyx I didn't mean to overstep, I just tested it since I saw it'd been 6 days

templates/vue/src/components/TapestryApp.vue
```diff
@@ -229,29 +233,33 @@ export default {
...
   transform: scale(1);
   transform-origin: top left;
   transition: all 0.2s ease-out;
+  width: 100%;
   z-index: 0;
 
   @media screen and (min-width: 500px) {
     &.sidebar-open {
-      transform: scale(0.7);
+      width: calc(100% - max(340px, 30%));
```
narendrasss, line 245: Just a note on using `width` instead of `transform` here. Animating an element's width triggers an [expensive browser process called reflow](https://pqina.nl/blog/animating-width-and-height-without-the-squish-effect/), which can cause jittery animations for slower computers and/or larger Tapestries.

That being said, the sidebar opening/closing isn't a frequent animation so it might not be that big of a deal. I noticed you wrote the commit @wynnset, so if it's all good for you then it's ok with me.

templates/vue/src/components/TapestryApp.vue
```diff
@@ -41,6 +41,7 @@
             @dragend="updateViewBox"
             @mouseover="handleMouseover(id)"
             @mouseleave="activeNode = null"
+            @nodemounted="updateSelectableNodes"
```
narendrasss, line 44: Could we perhaps rename this event to `mounted`? I think it reads a bit better.

```suggestion
            @mounted="updateSelectableNodes"
```

templates/vue/src/components/TapestryNode.vue
```diff
@@ -259,7 +258,7 @@ export default {
     },
   },
   mounted() {
-    DragSelectModular.updateSelectableNodes()
+    this.$emit("nodemounted")
```
narendrasss, line 261:
```suggestion
    this.$emit("mounted")
```

seanRong: This is also a lifecycle event name though, so I wanted to differentiate it. I could camelCase it I guess?

wynnset: I don't see an issue with following Narendra's suggestion as it's simpler. I don't think it matters that it's also a lifecycle event name, does it?

templates/vue/src/components/TapestryApp.vue
wynnset: I see. I couldn't find another way to make the drag-select selection rectangle appear in the correct size. If we can find a way to keep this as `transform` and make the drag-select selection work, that would be great. Do you think you could give it a look yourself quickly and let me know if you can find a way @narendrasss ?

narendrasss: I tried out a few things for around 30 mins and couldn't find an easy workaround. Might be a case of pre-optimization here; I wouldn't spend too long on it because it's only a brief animation.

templates/vue/src/components/TapestrySidebar.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-  <div ref="wrapper" :class="['sidebar-container', { closed: closed }]">
+  <div
+    ref="wrapper"
+    :class="['sidebar-container', { closed: closed }]"
+    @mousedown="pauseDragSelect"
```
wynnset, line 5: Hey @seanRong is this needed anymore? I tried removing it and it didn't seem to have any effect.

PR #811, opened 2020-10-19
717 - submit/accept/reject: review notification system

seanRong:
Adds a notification system in the top right to alert reviewers of nodes that could be reviewed.

They will be "new" until viewed once, and then they will no longer be new (however they would still count towards total yet to be reviewed).


seanRong:
NOTE: this needs to get updated with 680 or 712 such that router is completely integrated or else it just opens the lightbox instead of reviewform

seanRong:
Question: should there be forced rerendering whenever a student submits a node? Currently, the admin has to refresh their page. However, we don't seem to currently support just updating a singular node on the tapestry - the entire tapestry has to rerender.

narendrasss:
@seanRong This seems a bit related to the synchronous Tapestry thing we talked about over the summer, which would be tough to implement if we want to do live updates.

narendrasss:
Closing this in favour of #903

PR #812, opened 2020-10-19
804 - Bug Fix: Delete child nodes when links swapped

shirleywbi:
Closes #804

PR #815, opened 2020-10-19
813 - Bug Fix: Logged in users are allowed to delete links that dismember part of the tapestry

### LINKED ISSUE #813: Logged in users are allowed to delete links that dismember part of the tapestry

shirleywbi:
Closes #813 

Updated to use BFS against root. Old implementation did not check if clusters were connected.


yungalyx:
tested with method specified in #812 , LGTM! loved the queue and set implementation 👍

PR #820, opened 2020-10-22
818 - admin private nodes are not hidden

### LINKED ISSUE #818: Draft nodes by admin are not private

seanRong:
Closes #818 

Makes sure to not override the status filtered node result with the permissions filtered node result when not using superUser override, instead, we use their intersection

### DIFF
--- a/config/php.conf.ini
+++ b/config/php.conf.ini
@@ -3,3 +3,5 @@ memory_limit = 64M
 upload_max_filesize = 64M
 post_max_size = 64M
 max_execution_time = 600
+log_errors = On
+error_log = /dev/stderr
--- a/classes/class.tapestry.php
+++ b/classes/class.tapestry.php
@@ -491,7 +491,6 @@ class Tapestry implements ITapestry
         if (!isset($tapestry->settings->superuserOverridePermissions)) {
             $tapestry->settings->superuserOverridePermissions = true;
         }
-
         $tapestry->nodes = $this->_filterNodesMetaIdsByStatus($tapestry->nodes);
 
         if ($tapestry->settings->superuserOverridePermissions && $roles->canEdit($this->postId)) {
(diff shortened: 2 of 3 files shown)


config/php.conf.ini
```diff
@@ -2,4 +2,6 @@ file_uploads = On
 memory_limit = 64M
 upload_max_filesize = 64M
 post_max_size = 64M
-max_execution_time = 600
\ No newline at end of file
+max_execution_time = 600
+log_errors = On
+error_log = /dev/stderr
```
wynnset, line 7: Should this be here?

wynnset (approved):

config/php.conf.ini
seanRong: This allows error logging to actually happen, I don't believe it changes anything for the user

I was a bit surprised that it wasnt there already actually, how did everybody else debug?

shirleywbi: I set up XDebug in my vs code.

shirleywbi (changes requested):
Bug fixed but viewing large tapestries for public users is either broken or extremely slow >2.5min.

seanRong:
@shirleywbi how many nodes? I have one with 50 and its loading almost instantly

shirleywbi:
@seanRong Sent the tapestry json to you via Slack.

seanRong:
So strange, thanks!

shirleywbi (approved):
It's loading again!

shirleywbi:
@wynnset FYI. Looks good to go.

config/php.conf.ini
wynnset: I am still not using the docker setup. I look at the debug log for Wordpress generally which doesn't require this (I don't think).

PR #824, opened 2020-10-26
823 - Bug fix: Cannot use Ctrl/Cmd + A to select all nodes if logged in

### LINKED ISSUE #823: Cannot use Ctrl/Cmd + A to select all nodes if logged in

seanRong:
fixes #823 

Change from using mounted to using bv::modal to decide when to remove DragSelect's even listener in the settings modal in particular, as now the settings modal mounts immediately for logged in users and cannot be used to determine the visibility of the actual modal.


shirleywbi (approved):
LGTM

narendrasss (changes requested):
Code looks good, can you add tests?

seanRong:
As discussed, we should leave the tests for now

wynnset (approved):

PR #836, opened 2020-10-30
835 - Refactor and improve test utilities

### LINKED ISSUE #835: Improve Jest testing utils
Trying to render the TapestryApp in a Jest test will fail with a bunch of console errors related to Vue Router. This is somewhat related to [this asana task](https://app.asana.com/0/1126491658233864/1198971352239205), so both issues should be addressed.

narendrasss:
Closes #835 

## Why

Rendering the `TapestryApp` component generates lots of errors making it impossible to test the app by loading this component. Most of these errors result from two places:

- Vue Router not being configured correctly
- Progress not being initialized through Jest tests

## What

This PR fixes these problems by:

- Updating the test utilities module to match the `@testing-library/vue` API -> allows us to hook in Vue Router
- Refactor the dataset parsing to be self contained in the dataset module (instead of being spread out in `mutations` and `dataset`) -> allows reusability for testing


templates/vue/src/utils/test.js
```diff
@@ -1,7 +1,35 @@
...
 import { store } from "@/store"
-import { parse } from "./dataset"
+import { routes } from "@/router"
+import { parse, makeMockProgress } from "./dataset"
+import Helpers from "./Helpers"
 
+export function render(
+  component,
+  { fixture = null, ...options } = {},
+  callback = Helpers.noop
```
wynnset, line 10: Can we not just do this to keep it more simple:
```suggestion
  callback = ()=>{}
```

templates/vue/src/utils/Helpers.js
```diff
@@ -187,4 +187,10 @@ export default class Helpers {
 
     return false
   }
+
+  /**
+   * A "no-op" - does nothing.
+   * Useful as a default value when passing functions as arguments.
+   */
+  static noop() {}
```
wynnset, line 195: Not sure if we need this since calling it takes more work than just implementing it in-place.

templates/vue/src/utils/test.js
narendrasss: Sure.

templates/vue/src/utils/Helpers.js
narendrasss: I'm not sure if there would be much difference performance wise. Your other suggestion would make this part unnecessary though so I can take it off.

wynnset (approved):

PR #839, opened 2020-10-31
838 - Accordions display child nodes twice

### LINKED ISSUE #838: Accordion children display twice

jaskwong:
Closes #838

### DIFF
--- a/templates/vue/src/store/mutations.js
+++ b/templates/vue/src/store/mutations.js
@@ -121,7 +121,6 @@ export function updateRootNode(state, newNodeId) {
 
 // nodes
 export function addNode(state, node) {
-  console.log(state)
   Vue.set(state.nodes, node.id, node)
 }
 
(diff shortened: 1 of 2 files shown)


shirleywbi (approved):
LGTM.

templates/vue/src/store/mutations.js
```diff
@@ -83,6 +83,7 @@ export function setDataset(state, dataset) {
       Object.values(value).forEach(node => {
         // Has to call this so `state.nodes` is reactive
         Vue.set(state.nodes, node.id, node)
+        node.childOrdering = [...new Set(node.childOrdering)]
```
wynnset, line 86: Why is this needed? 🤔

wynnset (approved):
Looks good to me. I'd like @narendrasss to also take a look before merging this in just to make sure.

Update: With Jasmine's latest changes, I think this is pretty straightforward now so I just merged it in.

templates/vue/src/store/mutations.js
jaskwong: I think my intention was that if there was already a duplicate (ie. childOrdering contained the same id twice), then it stayed a duplicate even after reloading so this line was supposed to fix that for already affected nodes, I can test again maybe it's not needed

jaskwong: Update: I retested, turns out this line is unnecessary, that was an oversight on my part sorry about that

wynnset: No worries thanks for taking a look!

PR #840, opened 2020-11-02
825 - optimize thumbnails for Tapestry

### LINKED ISSUE #825: Utilize WP built-in functions to optimize thumbnails

seanRong:
## Changes

This PR overhauls the current thumbnail system for nodes. Large images take a long time to load and are often overkill given the size of a Tapestry Node. Instead, we now save another smaller/compressed version of the image uploaded and display that instead. 
In order to maintain backwards compatibility, a new button has been added in the settings modal to update all tapestries that use the old thumbnails to use optimized thumbnails. Old tapestries should still work even if they do not choose to optimize thumbnails.

## Issue Linkage

Closes #825
## PR Dependency

N/A

## Testing Notes
Remember that some images will not be compressed if they are already very small.


templates/vue/src/components/FileUpload.vue
```diff
@@ -35,20 +35,6 @@
...
-        <b-form-input
-          name="text-input"
-          :placeholder="placeholder"
-          :value="value"
-          :data-qa="inputTestId"
-          :disabled="isUploading"
-          required
-          @input="$emit('input', $event)"
-        />
-      </b-col>
```
wynnset, line 51: You should keep this in and hide/show it using a prop. We still need it for everything else (e.g. Video URLs, External Links, etc.)

seanRong: Thanks for the catch

seanRong:
Test failing because there's no add thumbnail by URL option anymore.
@narendrasss would you happen to know if we could simulate drag-drop upload?

rlnsy:
@seanRong I like what you've done here. One thing I noticed is that I can't seem to get this to work for locked-node thumbnails - can you verify this on your end?

seanRong:
@wynnset Remember how we said that we would discuss the locked thumbnails in the future?

The elephant in the room here is that a post could only have a single featured image/thumbnail without having to use other plugins. I was also thinking to perhaps do another PR to address locked to prevent this one from getting too old.

rlnsy:
> It doesn't work b/c not implemented :)
> 
> @wynnset Remember how we said that we would discuss the locked thumbnails in the future?
> 
> The elephant in the room here is that a post could only have a single featured image/thumbnail without having to use other plugins. I was also thinking to perhaps do another PR to address locked to prevent this one from getting too old.

Ah makes sense!

rlnsy:
@seanRong I thought I was going crazy, but it seems like something on this branch is giving me a bug. Basically, when I add a locked-node thumbnail to a node, I can no longer edit the node and clicking the edit button produces a bunch of errors in the console (I can share those if you'd like). The entire tapestry app hangs. I tested and this doesn't happen on master, only when I switch to this branch and re-build. Do you mind trying to replicate this? My process was basically to create a node with a locked thumbnail, lock it to a future date, then click the edit button on the node itself.

seanRong:
Can you post the errors please @rlnsy ? thanks!

rlnsy:
> Can you post the errors please @rlnsy ? thanks!

Here's the error I'm getting now. This time it occurred when ticking the "show a different thumbnail when locked" box, and prevented the thumbnail file upload from displaying at all:
Screen Shot 2020-12-17 at 2 41 26 PM

seanRong:
This is one of the weirdest bugs I've seen. Apparently one of the tags used for testing was making it hang.

wynnset:
> This is one of the weirdest bugs I've seen. Apparently one of the tags used for testing was making it hang.

I checked your commit and you had an extra quotation mark in another quotation mark inside `AppearanceForm` - that's probably what was causing the issue.

seanRong:
I see - sometimes VSCode decides to add stuff in there all on its own haha

wynnset:
One of the tests relating to the thumbnail is failing that we need to fix. The way the test was set up previous (forcing a wait) was not the correct way (cypress linter was complaining).

wynnset (approved):

wynnset:
@seanRong do you think you can fix this test? Otherwise this PR is good to go.

seanRong:
OK I seriously have no idea why the node appearance spec is failing on GA
Screen Shot 2021-01-02 at 10 25 30 PM
, this is running completely fine locally for me...

shirleywbi:
Seems like there might be an issue with the upload on CI since the image doesn't appear:
image
Like Sean, this works fine locally. Haven't been able to figure out what's wrong yet.

narendrasss:
> OK I seriously have no idea why the node appearance spec is failing on GA

I'm not super sure what's going on here, but I removed the url regex check since I don't think it's a good idea to check on a generated URL (it could change any time). Looks like it passes now, but it doesn't explain why the image doesn't show up in CI.

I'm going to leave it here for now.

shirleywbi:
@narendrasss Thanks. Not sure what the issue was. The imageURL for the node was an empty string and the Reddit image appears in the other tests.

wynnset (approved):

PR #845, opened 2020-11-05
844 - Bug Fix: Save root node as draft

jaskwong:


yungalyx:
Once I create a draft root node and add more nodes off of the root, I cannot convert the draft into a published node. (publish icon does not exist)

yungalyx:
looks good!

wynnset:
Just thinking about this again... I don't know if we should add this in because 2 people might create draft root nodes at the same time and when one of them publishes that root node, the other person's root node will be lost basically because it's not connected to anything. We might be able to add this in the future once we add "free-floating nodes"

templates/vue/src/components/NodeModal.vue
```diff
@@ -535,6 +536,12 @@ export default {
         })
       }
       await this.updateLockedStatus()
+      this.updateSelectedNode(this.node.id)
+      this.$router.push({
+          name: names.APP,
+          params: { nodeId: this.node.id },
+          query: this.$route.query,
+        })
```
jaskwong, line 544: @wynnset That sounds good, I think these lines might be a fix that we want to add regardless (including lines 525-527 and 479)

wynnset:
@jaskwong could you please move the relevant lines into a separate PR when you get a chance?

wynnset:
Will soon be resolved in #1222

Issue #841, opened 2020-11-03
Video Nodes with a Well-Formed but Nonexistent Video URL Hang on Submission


rlnsy:
One instance of this is when a user uploads a file, but modifies the URL then submits. E.g.
```
www.example.com/video.mp4
```
is well-formed and passes the checks introduced in #822 but causes our video components to error out.

PR #846, opened 2020-11-05
768/841 - Show errors when submitting a node with invalid video or if there are backend errors

rlnsy:
Since errors can occur in both the WP API and other places while submitting new nodes, we wish to provide better information about these errors to the user doing the submission. Furthermore, one example of a non-API error is the video URL problem described in #841.  This PR does the following:

- [x] displays a message to the user in the node submission box telling them an error has occurred

- [x] handles other errors related to video node submission that are currently not caught.

- [x] tests this new behaviour using cypress e2e tests

Some things to test:
1. See the new cypress tests and replicate the steps there if needed
2. Try opening the node modal, then logging out in a different window, then clicking submit. An error should appear in the modal.


jaskwong:
- when submitting after being logged out, it creates 2 identical toast alerts (I assume it should just create one)
- maybe when submitting with error it should scroll to the top of the modal so the user can see the error, because sometimes the error isn't in view (ex. if trying to submit with an invalid video url)

rlnsy:
>     * when submitting after being logged out, it creates 2 identical toast alerts (I assume it should just create one)
> 
>     * maybe when submitting with error it should scroll to the top of the modal so the user can see the error, because sometimes the error isn't in view (ex. if trying to submit with an invalid video url)

@jaskwong Thanks for the insight! I'm not actually doing anything related to the error toasts so I'm confused why this would introduce that behavior. I'll make sure to test it though.

jaskwong:
@rlnsy I think a toast is created everytime the “apierror” property of the state changes, I’m not sure if you’re changing it or not but this might be why

templates/vue/cypress/support/commands.js
```diff
@@ -166,6 +166,14 @@ Cypress.Commands.add("submitModal", () => {
...
 })
 
+Cypress.Commands.add("submitModalWithError", () => {
+  cy.server()
+  cy.route("PUT", `**/nodes/**/permissions`).as("editPermissions")
+
+  cy.getByTestId("submit-node-modal").click()
+  cy.getByTestId("node-modal", { timeout: 10000 }).should("be.visible")
+})
+
```
narendrasss, line 176: I believe this command would still run regardless if the modal has an error or not, because the node modal is visible until the action completes (so the assertion on line 174 will immediately pass). If you want to submit the modal without waiting, I believe just calling `cy.getByTestId("submit-node-modal").click()` should suffice.

templates/vue/src/components/__tests__/NodeModal.spec.js
```diff
@@ -1,63 +1,63 @@
-import { fireEvent } from "@testing-library/vue"
```
narendrasss, line 1: Was this an auto format thing? I don't see any differences.

narendrasss (changes requested):
@jaskwong @rlnsy just to chime in here - I've seen this behaviour on master as well, and it was happening whenever multiple api calls error out.

I do realize this wasn't marked ready for review yet, but I notice a few things in the code:

templates/vue/cypress/support/commands.js
rlnsy: @narendrasss yeah that's probably better, as the only reason I created the new command was to prevent waiting

rlnsy:
@jaskwong Yeah just ran a quick test on master and it happens there as well :(

templates/vue/src/components/__tests__/NodeModal.spec.js
rlnsy: @narendrasss I a little embarrassed to admit that I have no idea why this is showing as a change. I checked and the content of the files is exactly the same, even white-space. Any tips on helping git out here?

rlnsy:
@jaskwong I looked into adding an auto-scroll feature for the modal errors and I think we can use logic similar to what is already in the tapestry sidebar component - so it shouldn't be too hard. That being said, this should apply to the existing form errors as well, so maybe it should be addressed in a new PR. What are your thoughts?

jaskwong:
@rlnsy Are you referring to the toast errors? I think toast errors automatically show up at the top of the view area regardless of the scroll (ie. I think the css is position: fixed)

rlnsy:
> @rlnsy Are you referring to the toast errors? I think toast errors automatically show up at the top of the view area regardless of the scroll (ie. I think the css is position: fixed)

@jaskwong No, sorry, I mean the errors alerts that appear when a field item is missing, for example. Those already exist in master and I just copied the functionality for my alerts, so I would have to apply scrolling for both of them.

jaskwong:
@rlnsy ah got it, yea that makes sense, a new pr is probably better

rlnsy:
> @rlnsy ah got it, yea that makes sense, a new pr is probably better
> 
> **Edit: actually I think it depends how big of a change it would be, if its small enough I think it would be suitable to just do it w/n this one**

@jaskwong  Okay fair enough. I think it would be pretty smalll

rlnsy:
@jaskwong I couldn't figure out how to do this after spending a couple of hours on it. I'm thinking I'll create a separate ticket for now just to reduce the number of open PRs. Does that sound okay?

templates/vue/cypress/integration/node-authoring.spec.js
```diff
@@ -131,5 +131,63 @@ describe("Node Authoring", () => {
         })
       })
     })
+
+    const videoErrorMsg = "Invalid mp4 Video URL: please re-upload or check the URL"
```
wynnset, line 135: "Please" should be capitalized.

templates/vue/cypress/integration/node-authoring.spec.js
```diff
@@ -131,5 +131,63 @@ describe("Node Authoring", () => {
...
+        // check that node is not created
+        cy.contains(/cancel/i).click()
+        cy.contains(`/${nodeName}/i`).should("not.exist")
+
+        // check that error does not persist to new creation
+        cy.openModal("add", parent.id)
+        cy.changeMediaType("video")
+        cy.getByTestId(`node-title`).type(nodeName)
+        cy.getByTestId(`node-video-url`).type(
+          "https://www.youtube.com/watch?v=nMhua5LJRWg"
```
wynnset, line 164: Can we move this to the top of the file with the other constants?

templates/vue/src/components/NodeModal.vue
```diff
@@ -418,6 +431,17 @@ export default {
         }
       },
     },
+    hasSubmissionApiError() {
+      if (this.apiError) {
+        this.submissionError =
+          "An unexpected error occurred and the node could not be submitted"
```
wynnset, line 437: Why don't we show the actual API error instead rather than this ambiguous message?

wynnset (changes requested):
I made some visual updates to make our errors look a little better. Also asking for a few small changes below. 

Lastly, can we hide the toasts if they show up as a result of node submission and then instead just show them in the node submission modal instead? We could set a state when submitting the node and use it to determine whether to show toasts.

rlnsy:
@wynnset I made the changes requested. Let me know what you think of the solutions, and I'm happy to make further adjustments.

templates/vue/src/components/NodeModal.vue
```diff
@@ -10,16 +10,18 @@
...
     <b-container fluid class="px-0" data-qa="node-modal">
       <b-overlay :show="loading" variant="white">
-        <b-tabs card>
+        <div v-if="submissionError || formErrors.length" class="error-wrapper">
+          <h5>Node cannot be saved due to the following error(s):</h5>
+          <ul>
+            <li v-for="error in formErrors" :key="error">{{ error }}</li>
+            <li v-if="submissionError" data-qa="modal-submit-error">
+              {{ submissionError }}
+            </li>
```
wynnset, line 21: Can we add the submissionError as a part of formErrors so we avoid having two variables to keep track of errors to display? We can just amend the API error to the formErrors array maybe? You may need to move `setTapestryErrorReporting` functionality to the `handleSubmit()` section (not sure).

wynnset (changes requested):
Thanks @rlnsy this looks really good! Just one minor nit and we are golden. Feel free to set it as `ready to merge` after you are done.

wynnset (approved):

PR #847, opened 2020-11-05
792 - Improve search bar UX and allow searching by title

### LINKED ISSUE #708: Submit/accept/reject - Update search bar to filter by node status

### LINKED ISSUE #792: Search bar unintuitive

narendrasss:
## Changes

This PR overhauls the search bar to have more search features and to be more intuitive to use overall. It does this by:

- Changing styling completely for better UX
- Scrapping the `Combobox` component altogether in favour of a _much_ more accessible `v-select` component from the [Vue Select](https://vue-select.org/) library
- Integrating the [match-sorter](https://github.com/kentcdodds/match-sorter) library for an intuitive searching method (i.e. more than just doing a `startsWith` or exact match search)

## Screenshots

search

## Issue Linkage

Closes #792
Closes #708

## PR Dependency

N/A

## Automated Testing

- Search via author and title
  - Searching for author via id
  - Reset the search value if user changes the search type
- Search via node status
  - Hide rejected status by default
  - Show rejected status if `showRejected` setting
  - Reset value to 'All' if search type is changed to status
- Searching by visiting a search url directly
  - Show an error if an unauthorized user visits the url directly
- Limit authors to only those with public contributions
- Show loading indicator when superuser override is off
- Hide search bar for unauthorized users
- Searching for nodes with non-unique titles

## Other Notes

- Fixtures are being duplicated between cypress and Jest (an oversight on my part when setting those up). I think we should reconcile them into one fixture folder.


templates/vue/cypress/integration/search-bar.spec.js
```diff
@@ -0,0 +1,17 @@
...
+  /* beforeEach(() => {
+    cy.fixture("multi-author.json").as("tapestry")
+    cy.setup("@tapestry")
+  }) */
+
+  it("should be able to search nodes by its title", () => {
+    // click search button
+    // assert search by title by default
+    // type in 'hello'
+    // assert correct nodes are visible
```
narendrasss, line 11: I'm assuming that our "search results" won't be any different than it is now i.e. nodes that match the title in some way will have the regular opacity and nodes that don't will be more transparent.

There's other ways to handle search results, but keeping it the same would be the easiest. Some alternatives:

1. Highlight the parts of the title that match the query (on top of reducing opacity)
2. Show a dropdown of search results that, when an item is clicked, would "pan" the Tapestry to that particular node (this one's a bit of a stretch, but would look nice):

Screen Shot 2020-11-05 at 14 05 57

narendrasss (commented):
@wynnset Can you take a look at this? I also think adding a "waiting for approval" tag or something that shows the specification needs review would be nice.

templates/vue/cypress/integration/search-bar.spec.js
wynnset: I think we can keep it the same. There was not any negative feedback regarding the way we show results.

Just keep 2 things in mind here:
1. The selection of search criteria (e.g. "Title" in your mockup) should have an indicator that shows it's a select element (e.g. a down-pointing triangle to its right-handside)
2. If "Status" is selected as the search criteria, we should offer a select element, not auto-complete (default should be "All")

narendrasss: Ok - should there be a dropdown as well if we're searching by title?

seanRong:
Overall looks good, 
some minor issues:

- When searching by user. It lists out the usernames of all authors, even if that author has never made a public node, which could be confusing (ie it's also searching through draft nodes!). To add to this, if you start typing out the name of an author who hasn't made a public node, it will say "Sorry, no matching options" which is kind of confusing.

- Non - admins should not be able to see the search bar. However, if you abuse vue-router and type in an address with a query in it, like `http://127.0.0.1:8000/tapestry/privtest2/#/nodes/852?depth=0&search=title&q=`, you can 'hack' into the searchbar. This is probably bad security.

seanRong:
Also the text in the searchbox jumps a bit sometimes (you can see that in your GIF @narendrasss )

seanRong:
Off of a quick look,

708 also does filtering for accepted nodes, and also shows counts for the # of nodes of each type (also visually and programmatically disabling options that have 0).

narendrasss:
> Off of a quick look,
> 
> 708 also does filtering for accepted nodes, and also shows counts for the # of nodes of each type (also visually and programmatically disabling options that have 0).

@wynnset can you confirm these are required?

wynnset:
> > Off of a quick look,
> > 708 also does filtering for accepted nodes, and also shows counts for the # of nodes of each type (also visually and programmatically disabling options that have 0).
> 
> @wynnset can you confirm these are required?

@narendrasss yes, please.

narendrasss:
@seanRong I'm done with my portion of the work for this PR. Do you want to take over?

wynnset:
Seeing a bunch of these errors when I click on the searchfield and get the suggestions. I have some nodes that are titled the same, FYI. image

seanRong:
Ohh that's an edge case I didn't have locally. I'm not sure how much we can separate the keys from the dropdown options since I didn't code that part (@narendrass did)

seanRong:
Did github eat my comment or something? I might've already posted this but...

Going to pass this back to you @narendrasss, I'm still getting the text jumping in the search bar and I have no idea how to fix it. The failing tests is because I've altered the text in the searchbar for status to have a number displayed beside it.

Otherwise I've integrated everything from 708. Something that was added quite differently was the ability to see rejected nodes, we're doing this via a toggle in the settings modal this time instead, I thought this was more clear than having them appear when you clicked on rejected (the way it is now, since the implementations in 680 have changed since, you can't even click on rejected initially because it'll show 0 rejected by default).

As another consequence of the new 680 changes, some of the filtering has to be done a little differently or else the publish will return everything that's published (even accepted nodes).

wynnset:
A couple of other issues I found:
- [ ] When I search by title, let's say I type in "a", if I click outside, it should keep my search term. Currently, it clears it.
- [ ] Can we make sure the status labels are labelled "Accepted", "Submitted", "Rejected", "Draft", and "Published"

seanRong:
That last commit changed the names of the statuses globally. I think it could potentially cause some backward compatibility issues if somebody has, say, an accepted node on their tapestry because "accept" is now "accepted". I thought that we should probably just go all in and change everything while 680 is kind of new but I'm also open to the idea of just changing the names displayed in the filter. 

I'm also not really sure why the filter has the statuses capitalized because then I have to stick in a`toLowercase()` 🤷, although its nbd.

wynnset:
@seanRong review status names could be changed since #680 is still not pushed live and we can push this as part of this release still if we can get it out soon. Otherwise, we should just change the labels.

wynnset:
Couple more things:
- [ ] The search field should not have a placeholder, at least for "title" (currently it says "Node title" like in the screenshot below)
- [ ] Height of inputs should be reduced a bit (maybe to 38px) and the whole thing should probably be placed inside of the same container that the search icon is in (visually) something like this:
image

narendrasss:
> Couple more things:
> 
>     * [ ]  The search field should not have a placeholder - it should be empty initially or when there is no search query - at least for "title" (currently it says "Node title" like in the screenshot below)
> 
>     * [ ]  Height of inputs should be reduced a bit (maybe to 38px) and the whole thing should probably be placed inside of the same container that the search icon is in (visually) something like this:
>       image

I had the placeholders there because I believe it's better for accessibility to have those. Maybe have it more greyed out? In the screenshot you showed it does look like "Node title" is the search value instead of the placeholder.

seanRong:
@wynnset Will changing "publish" -> "published" cause any issues then? That's definitely already live.

wynnset:
> @wynnset I had the placeholders there because I believe it's better for accessibility to have those. Maybe have it more greyed out? In the screenshot you showed it does look like "Node title" is the search value instead of the placeholder.
@narendrasss if it's for accessibility, can we instead use aria-label? And it should say "Type in title or select from list" or something similarly instructional. I'm not sure if this is even necessary since we already have the other dropdown for selecting the element.

wynnset:
> @wynnset Will changing "publish" -> "published" cause any issues then? That's definitely already live.

We should definitely NOT edit the status attributes. Only reviewStatus attributes. publish should still be showing as "Published" when displayed to the user using a value->label association.

shirleywbi (commented):
Test Results: 
**Reset the search value if user changes the search type**
Only seems to work if value is selected from drop down but not if typed.
image
**Status missing in filter when selected**
image
**Author unable to view filter**
Is this expected?
image
**Toolbar is showing empty elements for filter and settings for subscriber/public**
image
**If I reject a node, should "All" show Total - Rejected or Total? with show rejected setting as off**
Count is normal after refresh.
image
**Submitted nodes are visible to editor**
Not sure if it's the scope of this ticket but should editor be able to view the submitted nodes when it doesn't have the permission to edit?
image

shirleywbi:
@narendrasss Came across some potential bugs listed above. Let me know if you have any questions regarding them.

templates/vue/src/utils/constants.js
```diff
@@ -29,4 +29,13 @@ export const licenses = {
   },
 }
 
+export const nodeStatus = {
+  ALL: "All",
+  DRAFT: "Draft",
+  SUBMITTED: "Submitted",
+  REJECTED: "Rejected",
+  ACCEPTED: "Accepted",
+  PUBLISHED: "Publish",
```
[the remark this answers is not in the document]
narendrasss: This was intentional I believe. I wanted to copy over the string literals used in #683 (iirc we use the literal `"publish"` for the publish status), but I think these are incorrect after looking at the code more.

templates/vue/src/components/TapestryFilter.vue
```diff
@@ -1,144 +1,330 @@
...
+
+const filterTypes = {
+  AUTHOR: "Author",
+  TITLE: "Title",
+  STATUS: "Status",
+}
 
-const filterOptions = {
-  AUTHOR: "author",
+const statusMap = {
```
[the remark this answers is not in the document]
narendrasss: Yes - the idea was `nodeStatus` lets us be consistent for status naming, and `statusMap` maps _display values_ to those statuses.

templates/vue/cypress/integration/search-bar.spec.js
```diff
@@ -0,0 +1,176 @@
...
+        path,
+        query: {
+          search: "Title",
+          q: "first node",
+        },
+      })
+
+      cy.url().should(url => expect(url.endsWith(path)).to.be.true)
+      cy.contains(
+        `You don't have access to the search bar for this Tapestry.`
```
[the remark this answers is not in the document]
narendrasss: I think the cypress test loaded the page first _then_ visits the invalid URL, showing the error. Good catch on visiting the url directly - I'm not sure how easy for this to fix on load.

narendrasss:
@shirleywbi Thanks for the review! Regarding your comments:

- I believe the toolbar being hidden is intentional behaviour. Right now only users who have edit access to the Tapestry itself can see the toolbar.
-  _Submitted nodes are visible to editor_ - this is also intentional because editors in WordPress have similar access privileges as administrators.

shirleywbi:
@narendrasss I see. 
In that case:

- Should the toolbar be completely out of sight versus the small empty squares that are still on the screen?
- I think there might be an issue with the editor because they can't actually do anything to the submitted nodes (e.g., edit/open the node).

wynnset:
> * I believe the toolbar being hidden is intentional behaviour. Right now only users who have edit access to the Tapestry itself can see the toolbar.

Only the author filter should be limited to those who have edit access. Search by title and status should be open to everyone.

templates/vue/src/utils/constants.js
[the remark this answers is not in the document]
wynnset: "Published" should be what the user sees in the frontend. "publish" should be what we save the status as. Reminder that the only VALID Wordpress post status are "publish" and "draft".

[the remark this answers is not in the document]
seanRong: This is correct: the strings here need to match the stauses of the nodes exactly

These are not what is displayed in the dropdown,  theres a seperste translation there. These are the strings we are matching on when filtering

narendrasss:
> **Reset the search value if user changes the search type**
> Only seems to work if value is selected from drop down but not if type

@wynnset do you think it's ok to leave it as is? It's not straightforward to adjust the search bar value because of the nature of the `v-select` library.

wynnset:
> > **Reset the search value if user changes the search type**
> > Only seems to work if value is selected from drop down but not if type
> 
> wynnset do you think it's ok to leave it as is? It's not straightforward to adjust the search bar value because of the nature of the `v-select` library.

@narendrasss yes that is ok.

wynnset:
@narendrasss I'm not sure if you are aware of this, but if I type something in the box and then select a value, it doesn't populate the selected value in the search box, although it does update the URL. This is weird behavior and should be fixed. If we select a value, it should show that value in the search box.

narendrasss:
> @narendrasss I'm not sure if you are aware of this, but if I type something in the box and then select a value, it doesn't populate the selected value in the search box, although it does update the URL.

Is this what you're referring to?

no_select

templates/vue/src/components/SettingsModal.vue
```diff
@@ -39,6 +39,19 @@
             </b-form-checkbox>
           </b-form-group>
           <b-form-group
+            label="Show me rejected nodes"
+            description="If enabled, you will be able to see all submitted nodes that have previously been rejected. Mind the decreased performance"
```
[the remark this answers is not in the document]
narendrasss: Thanks! As a tip for small changes like these you can [suggest changes](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/commenting-on-a-pull-request) by adding a code block with the suggestion tag like:

```
```suggestion
 // your suggestion
```

This lets me commit it directly:

Screen Shot 2020-12-09 at 14 19 09

narendrasss:
@wynnset @shirleywbi Thanks both, your comments should be addressed now. Let me know if there's anything else.

wynnset:
> > @narendrasss I'm not sure if you are aware of this, but if I type something in the box and then select a value, it doesn't populate the selected value in the search box, although it does update the URL.
> 
> 
> 
> Is this what you're referring to?
> 
> 
> 
> no_select
> 
> 

Yes!

shirleywbi (commented):
Thanks for the changes @narendrasss. A couple/questions more issues:

Should we be able to toggle open/close the search bar? If I open the search bar, everything is greyed out and the only way for me to view all the nodes again is to select status: all.

What should be the expected behaviour of clicking the search icon? I expected it to toggle the search bar but got an error message:
image

Unable to close search bar by removing part of the URL when converting from:
http://localhost/tapestry/792-2/#/nodes/4230?depth=2
to
http://localhost/tapestry/792-2/#/nodes/4230?depth=2&search=title
ezgif com-gif-maker (1)

After adding "Find this" node and clicking the search, is it expected to get duplicate items? (Refresh resolves this issue)
image

Is this clear selected icon supposed to be there? If so, it's near invisible unless you hover on it. It also doesn't seem to do anything and only appears with Author and not Title.
To get this button to appear: Make a selection (e.g., admin)
image
image

shirleywbi:
Also, not sure if this is part of the scope but if I reject a node in editor, the node doesn't seem to appear in admin.
ezgif com-gif-maker (2)

narendrasss:
> What should be the expected behaviour of clicking the search icon? I expected it to toggle the search bar but got an error message

@shirleywbi Oh thanks, not sure how I missed this - should be fixed now.

Regarding your other comments:

- I'm unable to replicate the duplicate admin issue. I tried adding a node both with the search bar closed and open.
- Good catch. I believe this is part of the dropdown library I used. Since we're not using it here I've set the css to `display: none` so it's not confusing.

> Also, not sure if this is part of the scope but if I reject a node in editor, the node doesn't seem to appear in admin.

IIRC this is intentional behaviour since rejecting a node makes it disappear for reviewers (I don't believe multiple reviewers can review the same node).

shirleywbi (approved):
LGTM!

wynnset:
@narendrasss this is great, we are almost there! Draft count seems to be wrong. It may be taking into account all the draft nodes rather than the number of nodes this current user has access to. I'm not sure, but maybe this is also happening with the other statuses. We gotta make sure the count reflects the current user, so if the user doesn't have access to some nodes, those nodes should not be included in the counts.

templates/vue/src/components/TapestryApp.vue
```diff
@@ -1,11 +1,6 @@
...
   <div id="app-container" :class="{ 'sidebar-open': isSidebarOpen }">
-    <div
-      v-if="
-        canEdit || (!settings.renderMap && maxDepth > 1 && settings.defaultDepth > 0)
-      "
-      class="toolbar"
-    >
+    <div v-if="!settings.renderMap" class="toolbar">
       <tapestry-filter style="z-index: 10;" />
       <div class="slider-wrapper">
```
wynnset, line 5: The restrictions that were removed above should probably be added to this, otherwise it's appearing as an empty square, which is not what we want.

templates/vue/src/utils/constants.js
```diff
@@ -29,4 +29,12 @@ export const licenses = {
   },
 }
 
+export const nodeStatus = {
+  PUBLISH: "publish",
+  DRAFT: "draft",
+  SUBMIT: "submitted",
+  REJECT: "reject",
+  ACCEPT: "accept",
```
wynnset, line 37: To keep consistent with everywhere else, can we keep these as `rejected` and `accepted`?

wynnset (changes requested):

templates/vue/src/utils/constants.js
narendrasss: These were the strings used in #680, so I kept it as is in case there are any nodes in prod with these statusses.

narendrasss:
> Draft count seems to be wrong. It may be taking into account all the draft nodes rather than the number of nodes this current user has access to.

@wynnset can you describe your steps to reproduce this? I can't replicate the problem.

Right now I have an admin user with one draft node, and a subscriber with two draft nodes. On the subscriber side it's correctly showing 2 draft nodes:

Screen Shot 2020-12-14 at 15 50 20

Likewise on the admin side it's correctly showing 1 draft node:

Screen Shot 2020-12-14 at 15 51 09

templates/vue/src/utils/constants.js
wynnset: That released was skipped as per Steven's request, so you can safely change these. Thanks for thinking of that though!

wynnset:
@narendrasss it might be a data issue because it's only behaving that way with an old tapestry. On a newly created tapestry, I don't have those issues.

wynnset:
I just pushed some visual and behaviour changes. I've made it so that we don't clear the search when users select an item, and have the v-select displayed value always match what's in the route query parameter. I think it feels the most intuitive this way, although it feels a bit hacky. @narendrasss can you take a quick look at my changes when you get a chance just to double-check my changes are valid and see if you have any ideas for improving it? This still requires more testing.

seanRong:
Perhaps more of a suggestion here, but since this PR deals with UX anyways:
Should we perhaps allow the user to press ESC to get out of the search feature?

seanRong:
Notable edge cases that I tested:
Attempting to fill in the search bar using terms in the URL (this actually works)
Attempting to navigate to an invalid search parameter (ie status = "blargh" or filter by "asdfasdf"), both these cases are caught by the app
Adding new nodes/changing status does in fact update the filter counts live.

Things that I found were broken:
The show rejected option for admin was broken. After a lot of digging I found that we still weren't consistently using the reviewStatus names as outlined in constants. This has since been updated and now it all works.

all that being said I think this PR is looking really solid!

seanRong:
P/S to any further testers - check if the depth slider is appearing for subscribers! I had it disappear briefky for some unknown reason

shirleywbi:
> Perhaps more of a suggestion here, but since this PR deals with UX anyways:
> Should we perhaps allow the user to press ESC to get out of the search feature?

Tagging @narendrasss @wynnset. Thoughts?

wynnset:
Thanks for fixing that @seanRong ! I thought I just couldn't find it because I was doing something wrong. 

> > Should we perhaps allow the user to press ESC to get out of the search feature?

If this is easy to add, then I don't see why not. But at the same time, this PR is getting really old so we should try and merge it ASAP. 

Is everything else working?

narendrasss:
> Should we perhaps allow the user to press ESC to get out of the search feature?

I think I would like to defer that to another PR just because I don't feel this should be delayed any further. Good idea though.

wynnset:
> check if the depth slider is appearing for subscribers! I had it disappear briefly for some unknown reason

@seanRong the depth slider does not show up when I'm logged out or logged in as non-admin, although this seems to also be the case on master too! I'll create a separate bug report for that.

wynnset (approved):

PR #849, opened 2020-11-06
848 - Bug fix: thumbnail external link not updating

yungalyx:


wynnset:
@yungalyx this PR is marked as "waiting for answer" but I do not see a question. Can you please clarify what's going on?

yungalyx:
@wynnset hey, just confirming if this is actually a bug, so im waiting on Bita's response through Asana. I already created the PR for it so I thought it might be better to wait for her response before I close it.

wynnset:
Ok thanks! Next time please leave a comment here to keep the PR updated.

yungalyx:
Issue closed. Bug resulted from user error. i.e, adding an image url into external link instead of thumbnail url link.

PR #854, opened 2020-11-07
853 - Export from database

### LINKED ISSUE #853: Export from database

jaskwong:
- adds export endpoint to get tapestry from database
- updates import functionality to accept new and old tapestry exports

Closes #853 

- tests endpoint is returning non-null response
- tests importing form file (old and new format) and that node titles exist


yungalyx:
@jaskwong hey i think some of the settings arent being saved, try toggling the checkboxes in the settings modal and refreshing the page.

templates/vue/cypress/integration/import-export.spec.js
```diff
@@ -1,5 +1,43 @@
...
+    cy.route("GET", "**/tapestries/**/export").as("export")
+
+    cy.get('.settings-button').click()
+    cy.contains(/advanced/i).click()
+    cy.get('#export-button').click()
+
+    cy.wait('@export')
+    .then((xhr) => {
+        assert.isNotNull(xhr.response.body, "response body not null")
+    })
```
narendrasss, line 15: I'd suggest to use the Cypress assertion for this to keep it consistent. You can do something like:

```suggestion
    cy.wait('@export').its("response.body").should("not.be.null")
```

templates/vue/src/components/modals/SettingsModal/index.vue
```diff
@@ -378,15 +379,11 @@ export default {
...
       this.isExporting = true
-      let filteredTapestry = this.tapestryJson
-      filteredTapestry.nodes = filteredTapestry.nodes.filter(
-        node => node.status === "publish"
-      )
-      const tapestry = filteredTapestry
-      tapestry["site-url"] = wpData.wpUrl
-      const blob = new Blob([JSON.stringify(tapestry, null, 2)], {
+      const exportedTapestry = await this.exportTapestryStore()
+      console.log(exportedTapestry)
```
narendrasss, line 385: I think you meant to remove this?

templates/vue/src/services/TapestryAPI.js
```diff
@@ -58,6 +58,12 @@ class TapestryApi {
     }
   }
 
+  async exportTapestry() {
+    var url = `${apiUrl}/tapestries/${this.postId}/export`
```
narendrasss, line 62: I think a better name for this would be `getTapestryExport` since the function itself isn't doing the "exporting".

Also let's avoid using `var` for declarations and use `const` or `let` instead. Keeps things consistent.

templates/vue/src/store/actions.js
```diff
@@ -204,6 +204,14 @@ export async function deleteNode({ commit, dispatch }, id) {
   }
 }
 
+export async function exportTapestry({ dispatch }) {
+  try {
+    return await client.exportTapestry()
```
narendrasss, line 209: You don't need the `async` and `await` here since `client.exportTapestry()` already returns a Promise. Also would probably be a good idea to rename this action too (as above).

```suggestion
export function exportTapestry({ dispatch }) {
  try {
    return client.exportTapestry()
```

narendrasss (changes requested):
Overall good stuff, thanks!

Just a couple notes on the code.

templates/vue/src/store/actions.js
jaskwong: @narendrasss I think because it's nested in a try-catch block, without await, the catch block won't be reached and it will just return the promise as received, but with await, a rejected promise will throw an error that can be caught

classes/class.tapestry.php
```diff
@@ -322,6 +323,40 @@ public function getAllContributors()
...
+     */
+    public function export()
+    {
+        $nodes = [];
+        foreach ($this->nodes as $node) {
+            $temp = (new TapestryNode($this->postId, $node))->get();
+            if (NodeStatus::DRAFT == $temp->status) {
+                continue;
+            }
+            $nodes[] = $temp;
```
shirleywbi, line 339: @wynnset @jaskwong How should we be handling submitted, accepted, and rejected nodes? Do those export too?

wynnset: DRAFT nodes should not be included (meaning submitted or rejected nodes). Accepted nodes should be included since they are not DRAFT.

shirleywbi (approved):
LGTM!

templates/vue/src/components/TapestryMain/RootNodeButton/index.vue
```diff
@@ -148,15 +148,33 @@ export default {
...
+        }, {})
+      const keys = Object.keys(data.settings.defaultPermissions)
+      for (let key in originalDefaultPerms) {
+        if (!keys.includes(key)) {
+          this.changes.permissions.add(key)
+        }
+      }
+      if (this.changes.permissions.size > 0) {
+        this.changes.noChange = false
+      }
```
wynnset, line 177: I find this function a bit hard to read. Would you mind adding some spacing and comments to clarify to the reader what's going on?

wynnset (changes requested):
Looking great! Thank you for the hard work on this. Comments are mainly nits and simple fixes.

wynnset:
Please mark as "ready to review" once comments are addressed.

endpoints.php
```diff
@@ -293,6 +293,13 @@
             'callback' => 'get_all_user_roles',
         ],
     ],
+    'GET_TAPESTRY_EXPORT' => (object) [
+        'ROUTE' => '/tapestries/(?P<tapestryPostId>[\d]+)/export',
+        'ARGUMENTS' => [
+            'methods' => $REST_API_GET_METHOD,
+            'callback' => 'exportTapestry',
+        ],
```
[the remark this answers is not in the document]
jaskwong: @wynnset What permissions does the user need to export a tapestry?

PR #867, opened 2020-11-20
866 - Bug fix: New text legibility feature has flickering animation

### LINKED ISSUE #866: New text legibility feature has flickering animation

wynnset:
## Changes
* Change animation to use transition component instead of CSS hover for smoother animation that doesn't flicker
* Simplify some of the code in the TapestryNode component
## Issue Linkage
Closes #866
## PR Dependency
N/A
## Automated Testing
N/A


yungalyx:
@wynnset Looks good! A small issue but sometimes the transition wouldn't animate at all. It's not exclusive to the root node.
ezgif com-gif-maker

wynnset:
Thanks @yungalyx ! @narendrasss would you know off the top of your head why it's not animating sometimes? I noticed that too.. it usually happens at the beginning if it does happen.

narendrasss:
@wynnset can't say for sure just from looking at the code briefly, but it might be something to do with wrapping the `circle` components in a `transition`. Would need a bit of investigation I think.

wynnset:
@narendrasss thanks. Would you be able to spend like 0.5hrs on this and let me know ASAP, please?

narendrasss:
@wynnset played around with this for a bit and I can't seem to replicate the problem.

PR #869, opened 2020-11-23
857 - Bug fix: CI Docker container does not have permission on file upload directories (Cypress testing)

### LINKED ISSUE #857: 857 - CI Docker container does not have permission on file upload directories (Cypress testing)

rlnsy:
## Changes
Gets all cypress tests labelled as **CI FAIL** to work properly in our github testing environment by ensuring that the wordpress installation has access to its own content directory for uploads.
## Issue Linkage
Closes #857 
## PR Dependency
N/A
## Automated Testing
N/A


rlnsy:
@narendrasss I figured out that the issue arises from the `wp-content` directory having ownership than all other wp-data directories in the github environment. See this screen cap from one of the previous runs:
Screen Shot 2020-11-23 at 1 17 13 PM
For comparison, my local project has all the ownership set to the same user (my main user):
Screen Shot 2020-11-23 at 2 15 56 PM
The simple fix (which I have implemented), is to use the github runner user in elevated privilege mode to modify the ownership of that directory specifically to the `www-data` user. This sort of feels like a hack though, and I'd be happy to continue looking into this for the root cause (i.e. what is different in the github env. to make this different). If this is good enough to merge as is, I'm happy with that too. Let me know your thoughts. Thanks!

wynnset:
Thanks @rlnsy for figuring this out! Just FYI, changing the directory permissions for the uploads folder is generally a part of the deployment procedures, so it might be OK to do that as part of our testing setup as well. You can take a look at [this article here](https://wordpress.org/support/article/changing-file-permissions/#permission-scheme-for-wordpress) that talks about the various approaches Wordpress suggests regarding changing permissions.

narendrasss (approved):
Thanks for figuring this out! It looks good.

@rlnsy were you planning to continue work on this? Just noticed it's assigned to me - not sure if you need anything other than my review.

rlnsy:
> Thanks for figuring this out! It looks good.
> 
> @rlnsy were you planning to continue work on this? Just noticed it's assigned to me - not sure if you need anything other than my review.

I was not. I think @wynnset just assigned you to review.

PR #871, opened 2020-11-24
870- Urgent fix: Unable to add root node

### LINKED ISSUE #870: Urgent fix: Unable to add root node

seanRong:
# Changes
Made a null check in the helper for hasPermission. The only time I believe the passed in node parameter would be null is when creating the root node, and a check for status on a null node will not work. This allows the root node to be created
## Issue Linkage
Closes #870

### DIFF
--- a/templates/vue/src/utils/Helpers.js
+++ b/templates/vue/src/utils/Helpers.js
@@ -140,17 +140,19 @@ export default class Helpers {
   }
 
   static hasPermission(node, action) {
-    if (node.status === "draft") {
-      if (wp.canEditTapestry() && node.reviewStatus === "submitted") {
-        return true
-      } else if (node.author && wp.isCurrentUser(node.author.id)) {
-        // authors cannot edit their submitted draft nodes
-        if (action == "edit" && node.reviewStatus === "submitted") {
+    if (node !== null) {
+      if (node.status === "draft") {
+        if (wp.canEditTapestry() && node.reviewStatus === "submitted") {
+          return true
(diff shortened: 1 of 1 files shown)


templates/vue/src/utils/Helpers.js
```diff
@@ -140,17 +140,19 @@ export default class Helpers {
   }
 
   static hasPermission(node, action) {
-    if (node.status === "draft") {
-      if (wp.canEditTapestry() && node.reviewStatus === "submitted") {
-        return true
-      } else if (node.author && wp.isCurrentUser(node.author.id)) {
-        // authors cannot edit their submitted draft nodes
-        if (action == "edit" && node.reviewStatus === "submitted") {
+    if (node !== null) {
```
wynnset, line 143: I suggest to revert your changes and _instead_, add something here like
```
if (node === null) {
  return wp.canEditTapestry()
}
```

wynnset, line 143: I suggest to revert your changes and _instead_, add something here like
```
if (node === null) {
  return wp.canEditTapestry()
}

```

wynnset (changes requested):
This fix works, but I don't like the code change because it's considering that if `hasPermission` is called, that the user `canEditTapestry`. Although this is true as of now, it would be better to code in a way that would prevent possible future changes. For example, a bit further down that function (second check), we are referring to the author of the node, which would error out if the code ever gets executed there with the `node` being `null`.

wynnset (approved):

PR #876, opened 2020-11-26
868 - Bug Fix: Superadmins should have full edit rights on all sites

### LINKED ISSUE #868: Superadmins should have full edit rights on all sites

shirleywbi:
## Changes
* Converted TapestryUserRoles to TapestryUser
* Updated canEdit to use WP capabilities instead of roles
## Issue Linkage
Closes #868 
## PR Dependency
Depends on: N/A
## Automated Testing
* N/A - Did not add tests as our local environment is not set up for multisite. As such, this was also not tested.


shirleywbi:
@wynnset As discussed during Sprint Planning, if you could test this, that would be great!

wynnset:
@shirleywbi From what I'm reading on the Wordpress site current_user_can will always true for superadmins. 

See https://developer.wordpress.org/reference/functions/current_user_can/

> Note: Will always return true if the current user is a super admin, unless specifically denied.

shirleywbi:
@wynnset I think this should fix it. The `canEdit` used by `userIsAllowed` has been updated to look at 'edit_post' capabilities.

wynnset (approved):
Tested this and it works! Thanks Shirley.

PR #886, opened 2020-11-30
872 - Save analytics event

jaskwong:
- Adds calls to record analytics in components
- Adds ajax action/function in tapestry.php 
- Initializes analytics table in db 
- Add toggle on/off for saving analytics under settings->advanced (default should be off)

Closes 872

- For testing, look at the database


yungalyx:
Yup I can find the db and everything updates correctly 👍

wynnset (changes requested):
Just a few minor changes please. In addition, can I ask you to add a new toggle switch under "Advanced" tab of the settings modal to enable Analytics? It should be disabled by default.

shirleywbi (commented):
@jaskwong @wynnset 
I had a couple questions regarding this PR:
- Are analytics only saved for people with edit access? I've checked admin/editor which saves screen clicks but nothing is saved for author/subscriber/public.
- Are we supposed to be saving logging the user doing these actions?

I'm getting a lot of h5p-video log events even though all I'm doing here is:
(1) Click play
(2) Closing the modal
Some of which have the same time. Is this expected?

Also, are we missing a column in the table for the action "click", "pause", etc.?
image

To confirm, does a html5-video event correspond to the video node? If so, it doesn't seem to be logging anything.

Since users that can edit view accordion's differently from users with no edit access, I don't think the accordion events will ever log since they appear as lightboxes.

jaskwong:
@shirleywbi I think @wynnset can probably answer this better, I'm not too sure myself which events should be logged or not

wynnset:
@jaskwong @shirleywbi 
> * Are analytics only saved for people with edit access? I've checked admin/editor which saves screen clicks but nothing is saved for author/subscriber/public.

Analytics events should be saved for everyone, whether they are logged in, logged out, regardless of their role.
> * Are we supposed to be saving logging the user doing these actions?

We should be logging these, yes.

> I'm getting a lot of h5p-video log events even though all I'm doing here is:
> (1) Click play
> (2) Closing the modal
> Some of which have the same time. Is this expected?

This is unclear - we should re-test this once the next issue below is fixed.
> Also, are we missing a column in the table for the action "click", "pause", etc.?

It seems that instead of saving the action, we are saving `tapestry_tool_log_event`. That column should be stating the action e.g. `click`, `play`, `pause`.

> To confirm, does a html5-video event correspond to the video node? If so, it doesn't seem to be logging anything.

I'm guessing @shirleywbi meant an `h5p-video` node. That corresponds to an H5P Video node. We should also be saving similar events for regular videos or youtube videos.
> Since users that can edit view accordion's differently from users with no edit access, I don't think the accordion events will ever log since they appear as lightboxes.

Users that can edit should still be able to view accordions by clicking on the media icon on the main accordion node. In that case, it should record accordion logs for them. Is there something not working here as expected?

shirleywbi:
> I'm guessing @shirleywbi meant an `h5p-video` node. That corresponds to an H5P Video node. We should also be saving similar events for regular videos or youtube videos.

I'm actually referring to the "html5-video" event located in the VideoMedia
https://github.com/wynnset/tapestry-wp/blob/df958291340b91a127c6dd6cff64f02fea2def5e/templates/vue/src/components/lightbox/VideoMedia.vue
```
        client.recordAnalyticsEvent("user", "play", "html5-video", this.node.id, {
          time: video.currentTime,
        })
``` 
It's currently not logging the "play" and "pause" events.

> Users that can edit should still be able to view accordions by clicking on the media icon on the main accordion node. In that case, it should record accordion logs for them. Is there something not working here as expected?

My bad! This is working as expected.

shirleywbi (changes requested):
@jaskwong A couple of bugs:
1. The user_guid changes every time. This should be the same for the same user, right?
image
2. Normal videos (non-H5P) don't seem to be recording any analytics.
3. Is there a difference between auto-play and play events? I seem to always get both when I play an H5P video.
 
Made a minor edit with the import syntax but otherwise, it's looking good.

jaskwong:
@shirleywbi 
1. I think the id's are randomly generated every time, I'm not sure what the expected behaviour is @wynnset 
2. I will look into it, was it an uploaded video or from url?
3.  Will fix

wynnset:
> 1. I think the id's are randomly generated every time, I'm not sure what the expected behaviour is @wynnset 

Good catch @shirleywbi. Yes this is supposed to be the same for the same user. The way it used to function, the guid was generated the first time and then it was saved (I think to a cookie) and used/reused over and over again. 

If this has changed, we should fix it as it's crucial for the analytics.

shirleywbi:
> 2. I will look into it, was it an uploaded video or from url?
Video URL. I used a Youtube link.

tapestry.php
```diff
@@ -354,3 +354,93 @@ function prefix_title_entity_decode($response)
...
+        $rand = base_convert ($rand, 10, 16);
+        return strval($rand);
+      }, $uuid);
+    return $uuid;    
+}
+
+function create_tapestry_guid_cookie() {
+    if (!isset($_COOKIE['tapestry_guid'])) {
+        $uuid = create_tapestry_guid();
+        setcookie('tapestry_guid', $uuid, time()+31556926);
```
shirleywbi, line 387: @jaskwong @wynnset Now that we're storing cookies, will we need to add one of those warning/banners for the user to approve of storing cookies?

shirleywbi (commented):
@jaskwong What's the difference between "app" play vs. "user" play?
image

Otherwise, everything else looks good to me.

jaskwong:
> @jaskwong What's the difference between "app" play vs. "user" play?
> image
> 
> Otherwise, everything else looks good to me.

@shirleywbi Looking at the code it looks like VideoMedia logs as “app” for play and pause but h5pvideo logs as “user” (unless it’s autoplay or the lightbox is closed) but I’m not sure what the original intent was @wynnset

wynnset:
@jaskwong @shirleywbi The 'actor' is who initiates/causes an action. If the user has played the video, it's different than if the app auto plays the video. Those are two different actors.

tapestry.php
wynnset: I don't think we need that. To my best knowledge, those notices are only needed when third-party cookies are utilized in the EU, whereas these are our own cookies used for statistical reasons.

shirleywbi (approved):
LGTM!

wynnset (approved):

PR #888, opened 2020-11-30
862 - Reorganize component folder structure

seanRong:
## Changes
Many refactors related to component structure
## Issue Linkage
Closes (issue #862 )
## Testing
This doesn't actually change anything so you don't really need to test outside of checking to see it compiles.
However: please review the current changed folder structure and see whether or not it is logical. Many things have been moved, and since this is basically a dev QOL PR we should attempt to arrange things in a way that makes the most sense to us when working on Tapestry.

For future reference: casing matters a lot here for resolving pathnames, so be careful


seanRong:
I've made all the moves and renames as stated in Asana.

Some things that could be up for discussion are the things that are going into the `components/common` folder, and whether or not we should create any further subfolders under `components`. One thing that came to mind was that maybe we should place even the `LockedTooltip `component inside `components/tapestry-node` but I felt like because it was needed in `TapestryApp` it made less sense.

shirleywbi:
@seanRong A couple of thoughts:
- I thought it was a bit odd for `SettingsModal` and `SettingsModalButton` to be outside of a folder called settings-modal. Likewise with NodeModal node-modal. Should we either rename the folder or move the components in? Why keep them out?
- I don't think Lightbox.vue has been renamed yet (as per Asana).
- Should `TapestryAccordion` be placed in media > accordion? 
   - Reason: Accordion components use it.
- Should `TapestryActivity` be placed in media > activity?
   - Reason: Used by Question which is located there

seanRong:
> I thought it was a bit odd for SettingsModal and SettingsModalButton to be outside of a folder called settings-modal. Likewise with NodeModal node-modal. Should we either rename the folder or move the components in? Why keep them out?

I wanted to put them in there too. But right now it seems like those folders are actually for "components that are used by settings-modal" and "components that are used by node-modal". @narendrasss @wynnset since you two have been working on the code the longest was that the reason for those folders and should we change it as above?

> I don't think Lightbox.vue has been renamed yet (as per Asana).

Again I wasn't sure if that component in particular should be renamed but I wouldn't mind doing so

The last two points I will implement 👍

narendrasss:
> But right now it seems like those folders are actually for "components that are used by settings-modal" and "components that are used by node-modal"

Yes this is correct. It was inspired by some folder structures in React apps but I'm open to changing it. One thing to note though is that moving e.g. TapestryNode.vue into `/tapestry-node` means we would have to `import TapestryNode from "./tapestry-node/TapestryNode.vue"` which feels a bit awkward in my opinion.

- We can rename `TapestryNode.vue` into `index.vue` and the `tapestry-node` folder into `TapestryNode` to maintain the import as `import TapestryNode from "./TapestryNode"` however.

Another approach that's common is to split up the folder into the various "views" we have in the app. Typically these views correspond to pages but since we don't really have that in Tapestry I was thinking to split it up into the major building blocks i.e. the App itself (the svg), the Node Modal, the Settings Modal, the Lightbox, and the Sidebar.

Since the app encompasses everything, the folder would then look like:

```
/components
  /shared
  /NodeModal
  /SettingsModal
  /Lightbox
  /Sidebar
  /App <-- everything only used in App
  App.vue
```

Thoughts? @seanRong @shirleywbi @wynnset

shirleywbi:
> Another approach that's common is to split up the folder into the various "views" we have in the app. Typically these views correspond to pages but since we don't really have that in Tapestry I was thinking to split it up into the major building blocks i.e. the App itself (the svg), the Node Modal, the Settings Modal, the Lightbox, and the Sidebar.
> 
> Since the app encompasses everything, the folder would then look like:
> 
> ```
> /components
>   /shared
>   /NodeModal
>   /SettingsModal
>   /Lightbox
>   /Sidebar
>   /App <-- everything only used in App
>   App.vue
> ```

Since our components are pretty specific to each "view", I think treating them more like views is a good suggestion.

seanRong:
GHA really doesnt do well with capitalization does it??

templates/vue/src/components/TapestryNode/index.vue
```diff
@@ -121,15 +121,15 @@
...
 import Helpers from "@/utils/Helpers"
 import * as wp from "@/services/wp"
-import AddChildButton from "./tapestry-node/AddChildButton"
-import ProgressBar from "./tapestry-node/ProgressBar"
-import StatusBar from "./tapestry-node/StatusBar"
-import NodeButton from "./tapestry-node/NodeButton"
+import AddChildButton from "@/components/TapestryNode/AddChildButton"
+import ProgressBar from "@/components/TapestryNode/ProgressBar"
+import StatusBar from "@/components/TapestryNode/StatusBar"
+import NodeButton from "@/components/TapestryNode/NodeButton"
```
narendrasss, line 132: I see the idea behind changing these to absolute paths, but I think for these components that live in the same folder relative imports are nicer e.g. `./AddChildButton`instead.

narendrasss (changes requested):
A small nitpick on import convention. @wynnset would like your thoughts on this as well.

templates/vue/src/components/TapestryNode/index.vue
wynnset: Agreed. Although not the biggest deal.

seanRong:
Fyi, still working on splitting TapestryApp (frustratingly close to done)

wynnset:
@seanRong is this still needing some work?

seanRong:
All ok now. @wynnset

wynnset:
@seanRong can you please squash your commits to reduce merge conflicts?

wynnset:
@seanRong actually, I'll do that don't worry about it. I found a really easy way to do it (see this [stackoverflow answer](https://stackoverflow.com/a/53622018/7318884))

seanRong:
Neat, bookmarked that one

shirleywbi (approved):
LGTM!

wynnset (approved):

PR #895, opened 2020-12-08
889 - Add automated testing specific to TYDE

### LINKED ISSUE #889: Add automated testing specific to TYDE

rlnsy:
## Changes
1. Adds automated tests related to the following:
- Modules
- Stages
- Subtopics / Question sets
- Completion / progress for each of the above
-  Cockpit functions (favorite, profile, clicking on modules)
2. Also makes some source code tweaks to allow for testing, including a small extension to the tapestry API.
## Limitations
- Had troubles with jest tests, so the tests here are purely end-to-end. I'm open to suggestions for elements that would be better tested in Vue testing.
- Code coverage is quite good, but there are some random bits of code that I could not get to. Certain components could use more testing in the future, namely the activities summary (I was blocked on this due to a creation bug)
## Issue Linkage
Closes #889 
## PR Dependency
N/A
## Testing/Reviewing
Would love suggestions on test cases, code quality, optimizations etc.


templates/vue/cypress/integration/tyde.spec.js
```diff
@@ -0,0 +1,260 @@
...
+        cy.submitModal()
+        cy.getBySrc(earnedIcon).should("not.exist")
+        cy.getBySrc(notEarnedIcon).should("exist")
+      })
+    })
+
+    it("Should allow creation of a regular node (subtopic) from a question set", () => {
+      cy.fixture("tyde/one-question-set.json").as("oneQuestionSet")
+      cy.setup("@oneQuestionSet")
+      cy.getSelectedNode().then(node => {
```
shirleywbi, line 75: Submitted a bug report for this but I think there's an issue with getSelectedNode since it seems to be selecting Root, instead of the question set.
image

shirleywbi (commented):
@rlnsy I think you have the major behaviour captured. The only thing I can think of is whether we need a test for the profile page.

endpoints.php
```diff
@@ -520,7 +520,8 @@ function importTapestry($postId, $tapestryData)
         // Construct ID map and add nodes to new Tapestry
         foreach ($tapestryData->nodes as $node) {
             $oldNodeId = $node->id;
-            $newNode = $tapestry->addNode($node);
+            $parentId = $node->import_parent_id;
```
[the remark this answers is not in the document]
rlnsy: Yes it is. There was a bug where virtually all of my fixtures were being rejected, and including this extra field is a quick hack to make it work.

templates/vue/cypress/integration/tyde.spec.js
rlnsy: Oh good catch! Super easy to get false positives with cypress. I can actually just convert this to get the node by title.

templates/vue/cypress/support/commands.js
```diff
@@ -194,3 +194,21 @@ Cypress.Commands.add("getByTestId", (testId, ...args) =>
...
+
+Cypress.Commands.add("getBySrc", (srcURL, ...args) =>
+  cy.get(`[src="${srcURL}"]`, ...args)
+)
+
+Cypress.Commands.add("getByBackground", (imageURL, ...args) =>
+  cy.get(`[style^='background-image: url("${imageURL}")']`, ...args)
+)
+
+Cypress.Commands.add("openSpaceship", () => {
```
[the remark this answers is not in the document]
rlnsy: I think something like open menu might work.

templates/vue/src/components/node-modal/SpaceshipPartForm.vue
```diff
@@ -3,31 +3,36 @@
     <h6 class="mb-3 text-muted">Planet View Icon</h6>
     <b-form-group label="Not earned">
       <file-upload
+        input-test-id="tyde-planet-not-earned-icon"
```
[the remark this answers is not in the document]
rlnsy: Hmm it should be but it seems to be working with this old version somehow. I'll make the change.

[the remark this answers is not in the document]
rlnsy: @shirleywbi I changed it and it broke - I think because `input-test-id` is used specifically for text entry fields, so I'm gonna use this for now.

templates/vue/cypress/integration/tyde.spec.js
```diff
@@ -0,0 +1,307 @@
...
+        cy.openLightbox(node.id)
+        cy.getByTestId("record").click()
+        cy.contains(/done/i).click()
+        cy.contains(/submit/i).click()
+        cy.contains("Your response has been recorded.").should("be.visible")
+        cy.closeLightbox()
+      })
+      cy.openTydeMenu()
+      cy.getByTestId("tyde-menu-profile-button").click()
+      cy.contains("What's your name?").should("exist")
```
rlnsy, line 292: @wynnset This assertion fails right now. Can you confirm what is actually supposed to happen here? After that is figured out, this is ready for review.

wynnset: @rlnsy In the profile, the answer the user provided to the given question is supposed to show up. If you are answering using audio, then an audio element should exist with the recorded audio. Here you are checking for the question, but I think you need to be checking for the answer.

shirleywbi (approved):
LGTM!

rlnsy:
@shirleywbi @wynnset It looks like the failing test above (after merging the base branch) is an actual break in functionality that my tests detected. That is, I am unable to add a subtopic node to a question set when testing locally as well. Should I file this as a bug on Ansana?

shirleywbi:
@rlnsy If it's a functionality issue and not a test issue, please file the bug. Thank you.

wynnset:
@rlnsy I agree with @shirleywbi . Could you please update the status of this task?

templates/vue/cypress/integration/tyde.spec.js
```diff
@@ -0,0 +1,318 @@
...
+        "uDzIp1Ohyg",
+      earned:
+        "https://www.clipartmax.com/png/middle/29-292956_world-globe-clip" +
+        "-art-clipart-2-wikiclipart-wikiclipart-transparent-background-ea" +
+        "rth-clipart.png",
+      hover:
+        "https://www.clipartmax.com/png/middle/29-292956_world-globe-clip" +
+        "-art-clipart-2-wikiclipart-wikiclipart-transparent-background-ea" +
+        "rth-clipart.png",
+    },
```
wynnset, line 25: Could we use some images that are from a more reliable source like Wikimedia?

wynnset (changes requested):
Mostly looking good. Just one minor change... Please mark as `ready to merge` when changes are committed.

templates/vue/cypress/integration/tyde.spec.js
rlnsy: Good call

wynnset (approved):

PR #902, opened 2020-12-17
901 - Bug Fix: Subscriber with permissions cannot add

### LINKED ISSUE #901: Subscriber unable to add nodes when node permissions allow for adding

shirleywbi:
## Changes
* Update user permissions to only check node source when creating link for a new node
* Allow users to edit and add if they are the author and the node is not in the review process.
   * Permitting authors to add because users with edit access can change their permissions. This also matches the behaviour of `Helpers.hasPermission`.
## Issue Linkage
Closes #901
## PR Dependency
Depends on: #935
## Automated Testing
* As subscriber: should be able to add a child node then edit that child node using the node modal


templates/vue/cypress/integration/node-authoring.spec.js
```diff
@@ -235,4 +238,32 @@ describe("Node Authoring", () => {
...
+
+        cy.openModal("add", parent.id)
+        cy.getByTestId(`node-title`).type(child.title)
+
+        cy.changeMediaType(child.mediaType)
+        cy.getEditable(`node-text-content`).type(child.typeData.textContent)
+
+        cy.submitModal()
+        cy.contains(child.title).should("exist")
+        cy.get(".links").children().should("exist")
```
narendrasss, line 265: I think a better assertion for this would be to check if the link between the child and the parent exists. You can use the `cy.link` command to get the link given the parent id and child id, and the `cy.getNodeByTitle` command to get the id of the child node. Something like:

```suggestion
        cy.getNodeByTitle(child.title).its("id").then(childId => {
          cy.link(parent.id, childId).should("exist")
        })
```

narendrasss (changes requested):
Looks like this solved the initial problem of adding a node, but once that node is added I'm not able to edit it (as a subscriber):

Screen Shot 2020-12-22 at 11 10 16

I also had a small comment on the cypress test.

shirleywbi:
> Looks like this solved the initial problem of adding a node, but once that node is added I'm not able to edit it (as a subscriber):
> 
> Screen Shot 2020-12-22 at 11 10 16

@narendrasss Does this happen when the default tapestry allows subscribers to add/edit? The new node created should have the default node permissions so if the Tapestry doesn't allow it, the subscriber won't be able to edit even though they're the authro.
@wynnset Can I get what the expected behaviour should be for this?

narendrasss:
> Does this happen when the default tapestry allows subscribers to add/edit?

The default permissions only allowed read permissions for subscribers, so if that's expected behaviour then it's ok.

Having said that though I think previously we allowed the user complete permissions if they were the author, so not sure if that was changed at any point.

wynnset:
I'm not sure if I understand the issue here correctly, but:
- All users should be able to add draft nodes
- All users should be able to edit their own draft nodes and submit that node
- Once a node is submitted, the node author should not be able to edit it
- If the submitted node gets rejected, the author should be able to edit it again
- If the submitted node gets accepted, the author should no longer be able to edit it

The default permissions don't have anything to do with what the permissions of a node are. The only thing that determines the permissions for a node are the permissions as defined for each particular node (in the Access tab of each given node).

If this doesn't answer your question, feel free to message me on Slack so we can have a quick call to resolve it.

jaskwong:
@shirleywbi Works well! When I was reviewing the code, it seems to have a couple of different parts, but would it be possible to just add a case to the `addTapestryLink` function in endpoints.php (ex. checking if the user has add permission for the source node and is the author of the target node), I think this could lead to a simpler solution but perhaps there's something I'm missing

shirleywbi:
@jaskwong I opted to add the author permissions to the node since I thought it'd be better for the Node Permissions (front-end) to show that authors have permissions to it. Thoughts?

jaskwong:
> @jaskwong I opted to add the author permissions to the node since I thought it'd be better for the Node Permissions (front-end) to show that authors have permissions to it. Thoughts?

I was assuming that the newly added node should have the default permissions (or the permissions set by the author), and so they would only need temporary add permission to create the link, and then afterwards would not have add permission to the new node unless by default or explicitly stated, so I assume it would depend on the desired behaviour

shirleywbi:
> @shirleywbi I was assuming that the newly added node should have the default permissions (or the permissions set by the author), and so they would only need temporary add permission to create the link, and then afterwards would not have add permission to the new node unless by default or explicitly stated, so I suppose it would depend on the desired behaviour

@wynnset Thoughts on suggested behaviour above versus current behaviour of setting the author permissions?

wynnset:
@shirleywbi & @jaskwong let's have a chat about this on Monday or Wednesday please.

classes/class.tapestry.php
```diff
@@ -134,8 +134,12 @@ public function getNodeIds()
     public function addNode($node)
     {
         $tapestryNode = new TapestryNode($this->postId);
-        $tapestryNode->set($node);
 
+        // If user is author and node is not in review process, grant author edit permissions
+        $userId = wp_get_current_user()->ID;
+        $isAuthor = $userId === $tapestryNode->getAuthor()->id;
```
wynnset, line 140: Wouldn't this always be true?

endpoints.php
```diff
@@ -679,7 +679,9 @@ function addTapestryLink($request)
         if (!TapestryHelpers::userIsAllowed('ADD', $link->source, $postId)) {
             throw new TapestryError('ADD_NODE_PERMISSION_DENIED');
         }
-        if (!TapestryHelpers::userIsAllowed('ADD', $link->target, $postId)) {
+        $isNewLink = $link->value == 1;
```
wynnset, line 682: This line is a bit confusing and looks hacky. Why does this test on `link->value` show that it's a new link? And what exactly does it mean for a node to be new in this case?

classes/class.tapestry.php
shirleywbi: Good catch. If a user is adding a node, they're going to be the author. I'll remove it.

endpoints.php
shirleywbi: Originally, I was basing it off newLink in the NodeModal/index.vue to differentiate it from creating a link via add button:
```
          const newLink = {
            source: this.parent.id,
            target: id,
            value: 1,
            type: "",
          }
```
I added a property `addedOnNodeCreation` to clear that up. Let me know if that's still confusing.

shirleywbi:
With the fix from #935, the operation and sidebar tests is passing locally.
image
The only test that is failing now is the first `node-permissions` test. The test fails because public users can view the node when "User-0" is checked. Since User-0 has edit access, bulk unchecking fails. User-0 seems to correspond to "Public". As of right now, I am unsure why there is a User-0 and where that is initialized. 
image

jaskwong (approved):

jaskwong:
lgtm!

utilities/class.tapestry-helpers.php
```diff
@@ -219,6 +219,12 @@ public static function userIsAllowed($action, $nodeMetaId, $tapestryPostId, $sup
         $groupIds = self::getGroupIdsOfUser($userId, $tapestryPostId);
         $user = new TapestryUser($userId);
 
+        // If user is author and node is in review, user is not allowed to edit node
+        $inReview = isset($node->reviewStatus);
```
wynnset, line 223: This is not accurate. Only nodes with an `accepted` or `submitted` review status are not allowed to be edited by their author. Nodes that are in review but have a review status of `rejected` should still be editable by the original author.

utilities/class.tapestry-helpers.php
```diff
@@ -219,6 +219,12 @@ public static function userIsAllowed($action, $nodeMetaId, $tapestryPostId, $sup
...
+        if ($action === "EDIT" && $inReview && $user->isAuthorOfThePost($nodePostId)) {
```
wynnset, line 224: I'd replace `$user->isAuthorOfThePost($nodePostId)` with `!$user->canEdit($tapestryPostId)` since super admins should still be able to edit nodes, regardless of whether they are the author or not.
```suggestion
        if ($action === "EDIT" && $inReview && !$user->canEdit($tapestryPostId)) {
```

wynnset (changes requested):
Great job figuring out the details of this bug and putting in a fix @shirleywbi ! Just a couple of quick changes needed and we're good. You can mark it as `ready for merge` once those edits are made.

utilities/class.tapestry-helpers.php
shirleywbi: @wynnset Since this is returning false for isUserAllowed, if I change it to canEdit, it means that a user who is editing a node in review **cannot** edit the node.

utilities/class.tapestry-helpers.php
shirleywbi: Thanks for catching that. Changed!

shirleywbi:
@wynnset FYI. I only made one of the suggested changes and had to change when full author permissions were added. Otherwise, it should be ready.

utilities/class.tapestry-helpers.php
wynnset: @shirleywbi notice the `!` so it is a negative check. It means if the user canNOT edit the tapestry, it will return false. If the user can edit the tapestry, that conditional return won't execute.

wynnset (approved):

PR #903, opened 2020-12-17
717 - Add notification bar

### LINKED ISSUE #717: Submit/accept/reject - Show submitted nodes to reviewers to accept/reject
The author of the Tapestry (or any reviewer) should see a notification button on the top right corner that would open a list of nodes that are awaiting approval.

Clicking on a submitted node entry would select that node in the tapestry (not open it for viewing) and also show the review comment history and a textarea just below where the submitted node entry appears with two buttons:
"Accept and Add" - this would add that node and make it visible to everyone.
"Reject" - this would mark the node as rejected

At the top, it would be great to also have:
A search bar (to search for a node or a user)
A filter for status (All, Pending, Accepted, Rejected) - by default only the pending items will show

(Maybe this could go in the sidebar?)

Some questions:
Can the author comment outside of an action? NO
Can students reply to comments? YES
Should comments persist? YES
Should other admins be able to view all comments? Can they contribute to those comment threads? YES

narendrasss:
## Changes

Adds a new notification button to the toolbar that shows a list of nodes awaiting review.

- Each item in this list is a link to the review form for that particular node
- Max height is set to `32rem`. When there are more nodes, the list becomes a scrolling view.
- When empty, the list shows a "No nodes awaiting review" message.

There are a few more nice-to-haves that I plan to include in this PR, which includes:

- [x] Having the menu close when you click outside of the menu

Nevertheless I'm putting this PR up for testing since the main functionality is done.

## Screenshot
Screen Shot 2020-12-18 at 13 02 25

## Issue Linkage
Closes #717

## PR Dependency
Depends on: #864 

## Automated Testing
N/A


shirleywbi (commented):
Are these only clickable if there is at least one node? The icons circled below don't seem to be clickable.
image
image

When I click on the notification, the no nodes awaiting for review appears. But, the notification is only there when I click and hold. Would this be better as a toggle?
image

Is the icon supposed to in greyscale when there is a pending review?
image

Menu title seems to be missing. Same with the "Submitted time by author".
image

shirleywbi (commented):
Are these only clickable if there is at least one node? The icons circled below don't seem to be clickable.
image

Minor styling: Should scrollbar be visible (on Windows, Chrome)? Also, missing bottom padding?
image

Otherwise, it's looking good to me.

narendrasss:
@shirleywbi Thanks for checking the root node case! Should be good now.

shirleywbi (approved):
LGTM!

templates/vue/src/components/NotificationBar.vue
```diff
@@ -0,0 +1,227 @@
+<template>
```
wynnset, line 1: Since this is only for reviews, should we name it something like `ReviewNotifications` or something instead? Also, after the component re-organization is done, this should probably be grouped in a folder with the toolbar.

wynnset (approved):

PR #904, opened 2020-12-18
Fixes from 844

jaskwong:
Adds router push after update selected node in nodemodal.vue

Testing: Should update the selected node after adding a node


templates/vue/src/components/modals/NodeModal/index.vue
```diff
@@ -655,9 +655,14 @@ export default {
...
-            this.updateSelectedNode(id)
-          }
+        }
+        if (this.node.status == "draft") {
+          this.updateSelectedNode(id)
+          this.$router.push({
+            name: names.APP,
+            params: { nodeId: id },
+            query: this.$route.query,
+          })
```
wynnset, line 665: @jaskwong can you explain what this fixes exactly? What was the bug we were having before this fix?

jaskwong: I think it was just to add the router push after updating the selected node, although it’s been a while since the other PR so maybe it’s not necessary anymore

wynnset:
As discussed, @jaskwong will create a task to take a more comprehensive approach to fixing the issue at hand here (the duplicate functionality provided by `updateSelectedNode` and the router push) and we can address that later as it's not urgent.

PR #908, opened 2020-12-18
907 - Bug fix: Depth slider not appearing for non-admins

### LINKED ISSUE #907: Depth slider not appearing for non-admins

narendrasss:
## Changes

Fixes the depth slider not appearing for non-admins.

- Moved the conditional down from the toolbar to the specific toolbar components
- Changed `v-if` to `v-show` for the depth slider because the depth slider updates the `maxDepth` on mount. Using a `v-if` that depends on `maxDepth` means the depth slider never mounts and the `maxDepth` never updates.

## Issue Linkage
Closes #907

## PR Dependency
Depends on: N/A

## Automated Testing
N/A


wynnset:
Just one issue... we sometimes end up having an empty toolbar and it looks weird. That's why we had the `v-if` there before:
image
image

I think it would be nice if we create some computed values to determine whether each part is showing so then we can more easily use them in the wrappers if needed.

narendrasss:
@wynnset should be good now.

wynnset (approved):

PR #914, opened 2020-12-22
913 - Bug Fix: Edit button in sidebar does not work

### LINKED ISSUE #913: Edit button in sidebar does not work

shirleywbi:
## Changes
* ~~Replace edit-node emit with router update in TapestrySidebar component~~
* Capture vue edit-node and open-node events in TapestryApp
* Update jest one-node with a different description and license
## Issue Linkage
Closes #913
## PR Dependency
Depends on: N/A
Conflicts with: #864 which fixes the edit button in sidebar.
## Automated Testing
* Check sidebar renders lightbox and modal upon clicking view and edit


templates/vue/src/components/TapestrySidebar.vue
```diff
@@ -179,6 +187,13 @@ export default {
         query: this.$route.query,
       })
     },
+    editNode() {
```
rlnsy, line 190: Forgive my lack of knowledge here but is there a reason we have to create a new route? I'm wondering if there's any way to use the existing edit route logic.

templates/vue/src/components/TapestrySidebar.vue
```diff
@@ -28,11 +32,15 @@
       <header class="sidebar-header">
         <h1 ref="info" data-name="info" class="content-title">{{ node.title }}</h1>
         <div class="button-container">
-          <b-button v-if="node.accessible || canEdit" @click="viewNode">
+          <b-button
```
rlnsy, line 35: Same thing here. Is there anyway to combine this with the conditions in the regular edit button?

rlnsy (commented):
@shirleywbi Works great! Just see me comments below.

templates/vue/src/components/TapestrySidebar.vue
shirleywbi: I'll take a look. I think we might be able to use mixins to avoid duplicating the call in TapestryNode or add something along the lines of this when mounting TapestryNode:
```
      this.$root.$on("edit-node", id => {
        this.$router.push({
          name: names.MODAL,
          params: { nodeId: id, type: "edit", tab: "content" },
          query: this.$route.query,
        })
      })
```

shirleywbi: @rlnsy Thoughts on the update?

rlnsy:
@shirleywbi I like the changes! This looks good to me.

wynnset (approved):

PR #917, opened 2020-12-22
916 - TYDE CoS (Master)

### LINKED ISSUE #916: CoS #1: Design data structure

narendrasss:
## Changes
Adds a specification for the data requirements for the TYDE Circle of Support activity. Currently only includes a `README` and a skeleton `index.vue` file. You can view a [rendered view of the README here](https://gist.github.com/narendrasss/388edaec897e3985e923561f703a8556).

**Note**: This is my interpretation of the Circle of Support activity, so let me know if anything in the README is incorrect.

## Issue Linkage
Closes #916 

## PR Dependency
Depends on: N/A

## Automated Testing
None required


wynnset (commented):
This looks great! The only thing I would say is I'm not sure if we need to have BOTH the `communities` array in `Connection` AND the `connections` array in `Community`. We should probably decide on one and the other one would be computed. Alternatively, we could also keep a separate array to keep track of this relationship. This would be similar to how we store links in a tapestry. They are not stored with the nodes, but rather separately in their own array.

Btw I can always look at the rendered output of markdown files in file changes by using the rich diff toggle, so no need for the gist (although appreciate the thoughtfulness!) 
image

narendrasss:
> The only thing I would say is I'm not sure if we need to have BOTH the communities array in Connection AND the connections array in Community. We should probably decide on one and the other one would be computed.

Yeah I wasn't sure about this either. I think given that Connection-Community is a many-to-many relationship, a separate array (or set perhaps) to keep track of the relationship would make sense. Not sure what to call it though - members? ConnectionCommunities?

> Btw I can always look at the rendered output of markdown files in file changes by using the rich diff toggle, so no need for the gist (although appreciate the thoughtfulness!)

Awesome! Wasn't aware of this.

wynnset (approved):

wynnset:
This looks good. Let me know if you want me to merge it.

wynnset:
@narendrasss could you fix the merge conflicts, please?

PR #922, opened 2020-12-27
921 - Allow submitted nodes to be moved by the author

yungalyx:
## Changes
* adds new permission checks to Helpers.hasPermissions to allow authors to move their submitted nodes (but prevents editing after submission)
* adds additional guards in endpoint.php
## Issue Linkage
Closes (issue #921 )
## Automated Testing
* N/A
~~* After creating submitted node, author will move node and assert that node has changed positions~~
    * After multiple attempts of trying to get the drag and drop to work based on the implementations listed below, I was unable to successfully get it working. As such, no tests have been added for this issue. This may be a d3 specific issue.
        * https://github.com/cypress-io/cypress/issues/3441
        *  https://github.com/cypress-io/cypress/issues/845
        * https://stackoverflow.com/questions/54027884/testing-d3-js-drag-events-with-cypress-js


narendrasss (approved):

templates/vue/src/utils/Helpers.js
```diff
@@ -188,6 +188,14 @@ export default class Helpers {
       return wp.canEditTapestry()
     }
 
+    // Check 2: User submitted node, can't edit, but can move
+    if (node.author && wp.isCurrentUser(node.author.id) && action == "move") {
+      if (node.reviewStatus !== nodeStatus.ACCEPT) {
+        return true
+      }
+    }
```
wynnset, line 196: This should be moved down below into the `if (node.status === nodeStatus.DRAFT)` conditional and the comment block for that conditional should also be updated.

templates/vue/src/utils/Helpers.js
```diff
@@ -188,6 +188,14 @@ export default class Helpers {
       return wp.canEditTapestry()
     }
 
+    // Check 2: User submitted node, can't edit, but can move
```
wynnset, line 191: Delete this comment
```suggestion
```

templates/vue/src/components/TapestryMain/TapestryNode/index.vue
```diff
@@ -12,7 +12,10 @@
         'has-title': !node.hideTitle,
       }"
       :style="{
-        cursor: node.accessible || hasPermission('edit') ? 'pointer' : 'not-allowed',
+        cursor:
+          node.accessible || hasPermission('edit') || hasPermission('move')
+            ? 'pointer'
+            : 'not-allowed',
```
wynnset, line 18: Would be nice to have the `move` icon if it's only moving.
```suggestion
        cursor:
          node.accessible || hasPermission('edit') ? 'pointer' :
            (hasPermission('move') ? 'move' : 'not-allowed'),
```

endpoints.php
```diff
@@ -1114,7 +1114,11 @@ function updateTapestryNodeCoordinates($request)
             throw new TapestryError('INVALID_NODE_META_ID');
         }
         if (!TapestryHelpers::userIsAllowed('EDIT', $nodeMetaId, $postId)) {
-            throw new TapestryError('EDIT_NODE_PERMISSION_DENIED');
+            if (TapestryHelpers::userIsAllowed('MOVE', $nodeMetaId, $postId)) {
+               // Authors are allowed to move their nodes but not edit after submission
+            } else {
+                throw new TapestryError('EDIT_NODE_PERMISSION_DENIED');
+            }
```
wynnset, line 1121: Reverse conditional to avoid empty statements
```suggestion
            if (!TapestryHelpers::userIsAllowed('MOVE', $nodeMetaId, $postId)) {
                throw new TapestryError('EDIT_NODE_PERMISSION_DENIED');
            }
```
I would also suggest changing `EDIT_NODE_PERMISSION_DENIED` to a newly created `MOVE_NODE_PERMISSION_DENIED` error to be more accurate.

wynnset (changes requested):

templates/vue/src/components/TapestryMain/TapestryNode/index.vue
shirleywbi: @wynnset Do we want the cursor to be `'move'` even on hover or just when we click and press down on a node? This change makes it so it's a move cursor is present on hover, even when hovering on the lightbox, add, and edit buttons.

wynnset: @shirleywbi on hover please. This will never happen when hovering on add / edit / view buttons because those buttons don't appear in the given conditional i.e. the negation of `node.accessible || hasPermission('edit')`.

shirleywbi: @wynnset I think there is one case for hovering on add. For instance, if a user is logged in, they can create draft nodes/submitted nodes even though they don't have permission. Also, with #937 allowing anyone to move the node (without saving it), I'm not sure what the expected behaviour should be in terms of the `move` cursor.

wynnset: Ok yes I guess we can just leave it as is then.

wynnset (approved):

PR #925, opened 2020-12-29
924 - Add abstraction layer between Tapestry and Tyde fixtures

### LINKED ISSUE #924: Add abstraction layer between Tapestry and Tyde fixtures

shirleywbi:
## Changes
* Update Cypress setup to add copilot to non-Tyde specific fixtures.
## Issue Linkage
Closes #924 
## PR Dependency
Depends on: N/A
## Automated Testing
N/A. Testing not necessary. Please review code.

### DIFF
--- a/templates/vue/cypress.gha.json
+++ b/templates/vue/cypress.gha.json
@@ -3,7 +3,8 @@
     "ADMIN_USERNAME": "admin",
     "ADMIN_PASSWORD": "secret",
     "SUBSCRIBER_USERNAME": "subscriber",
-    "SUBSCRIBER_PASSWORD": "subscriber"
+    "SUBSCRIBER_PASSWORD": "subscriber",
+    "TAPESTRY_ENV": "TYDE"
   },
   "baseUrl": "http://localhost:8000",
   "retries": 1
--- a/templates/vue/cypress.gha.json
+++ b/templates/vue/cypress.gha.json
@@ -3,8 +3,7 @@
     "ADMIN_USERNAME": "admin",
     "ADMIN_PASSWORD": "secret",
     "SUBSCRIBER_USERNAME": "subscriber",
-    "SUBSCRIBER_PASSWORD": "subscriber",
-    "TAPESTRY_ENV": "TYDE"
+    "SUBSCRIBER_PASSWORD": "subscriber"
   },
   "baseUrl": "http://localhost:8000",
   "retries": 1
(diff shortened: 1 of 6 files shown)


templates/vue/cypress/support/commands.js
```diff
@@ -3,9 +3,20 @@ import roles from "./roles"
 import { deepMerge } from "./utils"
 import { API_URL, TEST_TAPESTRY_NAME } from "./constants"
 
-Cypress.Commands.add("setup", { prevSubject: false }, (fixture, role = "admin") => {
+Cypress.Commands.add("setup", { prevSubject: false }, (fixture, role = "admin", tyde = false) => {
   if (fixture) {
     cy.get(fixture).then(tapestry => {
+      // Add copilot permissions for Tyde
+      if (tyde && !fixture.startsWith("tyde/")) {
```
narendrasss, line 10: I think a better way to do this would be to use a `TAPESTRY_ENV` variable inside the cypress config that we switch between the `tyde` and `master` branches accordingly:

```
{
  ...
  "env": {
    "TAPESTRY_ENV": "TYDE"
  }
  ...
}
```

You would then change this line to:

```suggestion
      if (Cypress.env("TAPESTRY_ENV") === "TYDE") {
```

And remove the `tyde` parameter at the top. This way, we don't have to define a new `tydeSetup` command, keeping the test code the same.

I also don't think we need the second check in the conditional. I think `!fixture.startsWith("tyde/")` will always be true since we're using aliases which always start with `@`.

narendrasss (commented):
Thanks for getting this started @shirleywbi! I have a different approach in mind that I wanted to discuss before getting this in:

templates/vue/cypress/support/commands.js
shirleywbi: Thanks for the suggestion. I didn't think to use env variables. I'll go ahead and make the switch. Let me know if there are any further changes/approaches you were thinking.

narendrasss (approved):
Awesome — thanks!

templates/vue/cypress.gha.json
```diff
@@ -3,7 +3,8 @@
     "ADMIN_USERNAME": "admin",
     "ADMIN_PASSWORD": "secret",
     "SUBSCRIBER_USERNAME": "subscriber",
-    "SUBSCRIBER_PASSWORD": "subscriber"
+    "SUBSCRIBER_PASSWORD": "subscriber",
+    "TAPESTRY_ENV": "TYDE"
```
wynnset, line 7: Not sure we need this. When would this not be `TYDE`?

wynnset (changes requested):

templates/vue/cypress.gha.json
shirleywbi: Good point. I guess in this branch, it would always be `TYDE`. I'll remove it and update the conditional mentioned above.

PR #927, opened 2021-01-04
926 - optimize locked thumbnail

### LINKED ISSUE #926: Optimize locked node thumbnail

seanRong:
## Changes
Applies the same changes that 825 applied to node thumbnails, but this time for locked node thumbnails. The thumbnail optimizer now optimizes both normal node thumbnails, as well as locked node thumbnails.
## Issue Linkage
Closes #926
## PR Dependency
Depends on: #840
## Testing Notes
Another reminder that some images may not be optimized if they are small enough. Suggest to test with larger images. Also the failing test is same as 825


shirleywbi (approved):
Made a couple of typo fixes, removed some unused code and cleaned up the optimizeThumbnail endpoint to reduce repetition but otherwise, it looks good to me.

wynnset (approved):

PR #929, opened 2021-01-04
898- Ask user to refresh page or login if session timed out

seanRong:
## Changes
* Uses heartbeat API to determine when a user has been logged out, and will display a modal asking them to either log in again or refresh and remain logged out.
## Screenshot
image
Clicking login brings you here and will return you back to tapestry page once logged in:
image

## Issue Linkage
Closes (issue #898)

## Automated Testing
N/A. To test, I had another window open where I would log the user in/out of to see if the tapestry browser would display the modal.


seanRong:
Unfortunately, I've run out of time to fully complete this, however it's pretty close to done

- the redirection to the login page hasn't been implemented yet
- could use some styling

Note that it took me ages to figure out why the heartbeat API wouldn't work with vanilla JS. I still don't know, so I had to implement a bit of jQuery in there. It's a bit messy and the linter doesn't like it, but it is the best working solution I could come up with.

templates/vue/src/App.vue
```diff
@@ -57,6 +90,12 @@ export default {
   },
   methods: {
     ...mapMutations(["init"]),
+    refresh() {
+      this.$router.go()
+    },
+    redirectToLogin() {
+      window.location.href = `${window.location.origin}/wp-login.php?redirect_to=${window.location.pathname}`
```
wynnset, line 97: This should ideally include the `#` and everything after it too so that we can take the user back to exactly where they were in the given tapestry.

templates/vue/src/App.vue
```diff
@@ -38,7 +52,26 @@ export default {
...
+    loggedIn(newValue) {
+      if (!newValue) {
+        this.$bvModal.show("loggedOutModal")
+      }
+    },
+  },
   mounted() {
+    if (isLoggedIn()) {
+      var that = this
+      jQuery(function($) {
```
wynnset, line 65: Can we avoid using jQuery by any chance? Can we use vanilla JS to check for the event, etc? jQuery will add quite a bit of load to our application.

templates/vue/src/App.vue
```diff
@@ -38,7 +52,26 @@ export default {
       return Object.keys(this.nodes).length === 0
     },
   },
+  watch: {
+    loggedIn(newValue) {
```
wynnset, line 56: Instead of `newValue` can we call this something more meaningful like `isStillLoggedIn`

wynnset (changes requested):

templates/vue/src/App.vue
shirleywbi: @wynnset Unfortunately, it doesn't seem like we're able to use vanilla JS.
https://wordpress.org/support/topic/heartbeat-api-in-vanilla-js-addeventlistener/
https://stackoverflow.com/questions/54008235/wp-heartbeat-api-in-vanilla-js-addeventlistener-for-custom-events

wynnset: After looking into this and trying a few things, I don't think we have a way of getting around this since WP Heartbeat API is fully relying on jQuery.

templates/vue/package.json
```diff
@@ -31,6 +31,7 @@
     "bootstrap-vue": "^2.13.1",
     "d3": "^5.16.0",
     "dragselect": "^1.13.1",
+    "jquery": "^3.5.1",
```
wynnset, line 34: @shirleywbi I don't think we actually need to include this in our build since Wordpress already loads jQuery and this functionality is dependent on Wordpress. Was there a specific reason you included it?

templates/vue/src/App.vue
```diff
@@ -38,7 +52,26 @@ export default {
...
+  },
   mounted() {
+    if (isLoggedIn()) {
+      var that = this
+      jQuery(function($) {
+        wp.heartbeat.interval("fast")
+
+        $(document).on("heartbeat-tick", function(event, data) {
+          that.loggedIn = data["wp-auth-check"]
+          console.log(that.loggedIn)
```
wynnset, line 70: Might want to remove this console log.

templates/vue/package.json
shirleywbi: @wynnset I think you're right. Just tested it without the package and it works fine. Removing it.

wynnset (approved):

PR #931, opened 2021-01-07
919 - TYDE CoS #2 - Add CoS endpoints

### LINKED ISSUE #919: Cos #2 - Add endpoints to save data

narendrasss:
## Changes
* Adds a new **`/activities/cos`** endpoint for reading and saving CoS data, which is kept in the user's `usermeta` table.
* CoS data is versioned by day and is checked on a per-request basis.
* Adds a new **`CircleOfSupport`** class that handles reading and saving logic. Currently only has two public methods **`get()`** and **`save($circleOfSupport)`**.

## Screenshot
N/A

## Issue Linkage
Closes #919

## PR Dependency
Depends on: (PR # or N/A)

## Automated Testing
N/A

### DIFF
--- a/classes/activities/class.circle-of-support.php
+++ b/classes/activities/class.circle-of-support.php
@@ -9,7 +9,11 @@ class CircleOfSupport
 
     public function __construct($userId = 0)
     {
-        $this->userId = $userId || wp_get_current_user()->ID;
+        if (!$userId) {
+            $userId = wp_get_current_user()->ID;
+        }
+
+        $this->userId = $userId;
         $this->versions = $this->_getVersions();
     }
 
@@ -25,6 +29,10 @@ class CircleOfSupport
             return null;
         }
 
+        if (0 === count($this->versions)) {
+            return null;
+        }
+
         return $this->versions[count($this->versions) - 1];
     }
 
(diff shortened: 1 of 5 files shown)


classes/activities/class.circle-of-support.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+class CircleOfSupport
+{
+    private $userId;
+    private $versions;
+
+    const META_KEY = 'circle_of_support';
```
wynnset, line 8: Please prepend this with `tyde_` so it doesn't potentially conflict with other meta keys that might be named similarly by other plugins.

templates/vue/src/components/tyde/activities/CircleOfSupport/index.vue
```diff
@@ -3,13 +3,24 @@
...
+        circles: [],
+        communities: [],
+        connections: [],
+        members: {},
+        timestamp: moment(),
+      },
     }
   },
+  async mounted() {
+    this.cos = await client.getCosActivity()
```
wynnset, line 23: This would just return null if there is no existing CoS saved already. Maybe you want to create a base case here (e.g. `if (retVal !== null)...`) otherwise your data variable would always get overwritten.

wynnset (changes requested):

wynnset (approved):

PR #947, opened 2021-01-28
765 - Bug Fix: Old H5P Nodes, When Edited, Show Up as "Video" Content Type

### LINKED ISSUE #765: editing h5p nodes shows up as video content type

jaskwong:
Creates a new tapestry-settings.php file for the tapestry settings page with one database command to clean old h5p nodes.
Filters nodes on loading to ensure h5p media formats and the media type are compatible 

Closes #765 

image

Testing: 
Check that there's a page under settings->tapestry
Should update posts with `mediaFormat == "h5p"` and `mediaType == "video"` to `mediaType = "h5p"`
When loading the tapestry, even if posts have  `mediaFormat == "h5p"` and `mediaType == "video"`,  the "Content Type" in the node modal should show h5p

### DIFF
--- a/classes/class.tapestry-node.php
+++ b/classes/class.tapestry-node.php
@@ -516,6 +516,9 @@ class TapestryNode implements ITapestryNode
 
     private function _formNode()
     {
+        if ($this->mediaType == "video" && $this->mediaFormat == "h5p") {
+            $this->mediaType = "h5p";
+        }
         return (object) [
             'id' => $this->nodeMetaId,
             'postId' => $this->nodePostId,
--- a/tapestry-settings.php
+++ b/tapestry-settings.php
@@ -1,0 +1,67 @@
+<?php 
+
+add_action('init', 'run_db_commands');
+add_action( 'admin_init', 'tapestry_settings_init' );
+add_action( 'admin_menu', 'add_tapestry_settings_page' );
+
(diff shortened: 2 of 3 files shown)


shirleywbi (approved):
Tested and LGTM! @wynnset, @jaskwong Is a refresh enough feedback that the task is complete?

shirleywbi:
Regarding the flaky test, it started appearing after the branch update. Tried to look into the cause but it doesn't seem like I'm able to reproduce the issue locally after 4 attempts.

wynnset:
> Tested and LGTM! @wynnset, @jaskwong Is a refresh enough feedback that the task is complete?

Not sure what you mean by refresh? We need to make sure that old h5p nodes that were showing as "Video" show up correctly as H5P nodes when editing the node.

Thinking about this further, I think that additionally, we should add some logic that when a Node Modal is opened to edit a node, we should check and see if the media type is Video and media format is H5P, and if it is, we should change the media type to H5P. Better yet, we should do this in the backend when loading nodes.

jaskwong:
@wynnset I think she means, once the command is run on the settings page, it just refreshes, but maybe we should add some sort of confirmation message like “The commands were successfully run” 

- will work on the other stuff

shirleywbi (approved):
Tested! Looks good.

PR #949, opened 2021-01-30
910 - Add automated testing for H5P

### LINKED ISSUE #910: Add automated testing for H5P

rlnsy:
## Changes
* Adds a fixture management system that uses docker at the volume level to autonomously include data that is outside the scope of Tapestry (e.g. H5P content)
* Modifies our CI workflow to run WP with an H5P fixture
* Tests basic H5P node functionality using 
## Issue Linkage
Closes #910
## PR Dependency
Depends on: N/A
## Testing
If you use docker for development, it is worth making sure the fixture loader works properly for you. This can be done as follows:
1. From the root directory, run `./bin/load-fixture.bash h5p_content_<platform>` where platform is one of "mac" or "linux" and should reflect the os of your host system. The mac version should work on Windows, but this is untested. This will not replace any existing data you have set up.
2. Run `make start-fixtures`, to start the app with the data from the fixtures directory. Navigating to H5P content in the WP admin dashboard should show a single content item called "Math Quiz".
3. All Cypress tests should pass, particularly `h5p.spec.js`.


rlnsy:
@narendrasss I spent a lot of time today trying to test the actual content within the H5P iframe, but couldn't get it to work (see the cypress `commands.js` file and my new test). Do you have any experience doing this?

narendrasss:
> @narendrasss I spent a lot of time today trying to test the actual content within the H5P iframe, but couldn't get it to work (see the cypress `commands.js` file and my new test). Do you have any experience doing this?

I've never done tests within an iframe before, but the blog post suggests you need to add the following line to the config (I don't see this change):

```
{
  "chromeWebSecurity": false
}
```

Maybe try that first?

templates/vue/cypress/support/commands.js
```diff
@@ -196,3 +196,18 @@ Cypress.Commands.add("getByTestId", (testId, ...args) =>
...
+// FIXME: subsequent finds don't seem to work
+Cypress.Commands.add("getIFrame", id =>
+  cy
+    .get(`iframe[id="${id}"]`)
+    // 1
+    .its("0.contentDocument")
+    .should("exist")
+    .its("body")
+    .should("not.be.undefined")
+    .then(cy.wrap)
```
narendrasss, line 212: I'm not sure if it's safe to do `.then(cy.wrap)` because if `cy.wrap` uses `this` inside its implementation it's not going to work as expected. I believe `.should` already yields the element, so I think this would work just fine:

```suggestion
  return cy
    .get(`iframe[id="${id}"]`)
    // 1
    .its("0.contentDocument")
    .should("exist")
    .its("body")
    .should("not.be.undefined")
```

rlnsy:
@narendrasss Thanks for the suggestions. I tried with the security setting, but it doesn't change anything since the iframe being loaded is not from a remote source. Also, removing `cy.wrap` causes errors when calling `.contains` on the resulting object. I'll have to look into this further.

rlnsy:
@wynnset Updating you on this PR... As I mentioned last week, what I thought was the hard part of this task (creating and loading content in the tests) is working just fine using my docker volume layer. However, I've encountered some limitations with Cypress with respect to iFrames (how the h5p content is included) that prevents me from doing very strong testing on the way the content interacts with the rest of the app. See my comments in the test and the `commands.js` file related to this. Let me know if you have any suggestions for assertions we can make given the technology we have at the moment - otherwise we might have to put this on the back-burner.

wynnset:
@rlnsy thanks for the update and for your investigation here. I guess if that's what you've found out, we can put this on the back burner as you've suggested. Is there any sort of tests we can add in that doesn't deal with what's inside the iframe?

rlnsy:
@wynnset Not that I know of in my limited experience with frontend testing, unfortunately. Some assertions about dimensions and specific visibility of elements would be great but I don't know how to do that without doing end-to-end testing on the content itself.

narendrasss:
@rlnsy @wynnset I think this is a good use case for visual testing since we only care if the H5P looks good within its container. I've never done it before, but might be worth looking into:

- https://medium.com/norwich-node-user-group/visual-regression-testing-with-cypress-io-and-cypress-image-snapshot-99c520ccc595
- https://percy.io/
- https://docs.cypress.io/guides/tooling/visual-testing.html

wynnset:
@rlnsy I'm not sure if we decided to completely put this on hold or if there are still things we want to do. Either way, can you please update the status / labels of this PR? It's currently marked as "waiting for answer" and assigned to you. Thanks!

wynnset:
Obsolete PR

PR #954, opened 2021-02-08
945 - Bug Fix: Delete draft child nodes from other users upon node deletion

### LINKED ISSUE #945: Delete draft child nodes from other users upon node deletion

shirleywbi:
## Changes
* Added warning modal if there are draft nodes (including rejected nodes) from other users attached to node being deleted
    * As stated in the ticket, only the node is deleted.
* Modified logic for root deletion to check for draft nodes
## Screenshot
With draft node attached:
ezgif com-gif-maker (2)
image
## Issue Linkage
Closes #945 
## PR Dependency
Depends on: N/A
## Automated Testing
* Given that a node has a draft node from another user attached to it, tests that warning modal renders, can cancel the delete, and can delete through the modal

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,46 +1,48 @@
 # tapestry-wp
 
-This is a plugin for Wordpress that allows creating non-linear, collaborative, and interactive content. 
+This is a plugin for Wordpress that allows creating non-linear, collaborative, and interactive content. This plugin adds a new post type to Wordpress called "Tapestry" and allows presentation and authoring in the frontend using Vue JS.
 
-Visit [tapestry-tool.com](https://www.tapestry-tool.com) for more info
+Visit [tapestry-tool.com](https://www.tapestry-tool.com) for more info.
 
-## Usage
+## How to install this plugin for Wordpress
 
-To install this plugin locally:
-- Ensure you have Wordpress installed locally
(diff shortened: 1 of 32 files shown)


wynnset:
@shirleywbi instead of dealing with rejected nodes differently, can we do it the same way as regular draft nodes? So allow deletion of the node but after giving them a warning and confirming with them.

jaskwong (approved):

jaskwong:
Functionality looks good, just a really small thing is when deleting a node with draft nodes attached, it still has the error message under the delete button "Only nodes with a single connection can be deleted." which might be a little misleading

shirleywbi:
> Functionality looks good, just a really small thing is when deleting a node with draft nodes attached, it still has the error message under the delete button "Only nodes with a single connection can be deleted." which might be a little misleading

@jaskwong Thanks. It should be fixed now. The error message just wasn't updating.

narendrasss (approved):

wynnset (changes requested):
@shirleywbi sound logic here and everything looks good. Just one small nit - can we be more accurate with our wording. Instead of calling it neighbour types, it might make more sense to call it "node has draft children" or something like that. This would help other readers more easily understand what's going on without needing to look at the implementation of your backend function. This should be changed in all places where we use a variation of "neighbour type" as the naming in this PR. Thanks!

Please mark as "ready for merge" when these changes are implemented.

wynnset (approved):

PR #957, opened 2021-02-09
956 - Bug Fix: Prevent Console Error for Redundant / Duplicate Router Navigation

jaskwong:
- overwrites the default router push function to catch duplicated navigation errors

Closes 956

Testing:
Console should no longer log router duplicate navigation errors (ex. when opening the settings and then closing it)

### DIFF
--- a/templates/vue/src/router.js
+++ b/templates/vue/src/router.js
@@ -20,6 +20,14 @@ const router = new VueRouter({
   routes,
 })
 
+// sourced from https://stackoverflow.com/questions/58634914/getting-navigation-duplicated-error-while-the-route-is-being-replaced-on-vue-rou
+const originalPush = VueRouter.prototype.push
+VueRouter.prototype.push = function push(location) {
+  return originalPush.call(this, location).catch(err => {
+    if (err.name !== "NavigationDuplicated") throw err
+  })
+}
+
 router.beforeEach((to, from, next) => {
   const nodes = Object.keys(store.state.nodes)
   if (to.matched.length === 0 && nodes.length > 0) {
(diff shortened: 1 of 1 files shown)


jaskwong:
@wynnset This solution isn't the same as the one suggested on Asana, but I think it's simpler since it doesn't require checking all the routes, it simply catches the error when navigation is duplicated.  Do you think it would be more valuable to check the routes before pushing?

templates/vue/src/router.js
```diff
@@ -20,6 +20,14 @@ const router = new VueRouter({
   routes,
 })
 
+// sourced from https://stackoverflow.com/questions/58634914/getting-navigation-duplicated-error-while-the-route-is-being-replaced-on-vue-rou
```
wynnset, line 23: Can you remove everything after the question number in the URL? The last part is not needed for the URL to work.

templates/vue/src/router.js
```diff
@@ -20,6 +20,14 @@ const router = new VueRouter({
...
 })
 
+// sourced from https://stackoverflow.com/questions/58634914/getting-navigation-duplicated-error-while-the-route-is-being-replaced-on-vue-rou
+const originalPush = VueRouter.prototype.push
+VueRouter.prototype.push = function push(location) {
+  return originalPush.call(this, location).catch(err => {
+    if (err.name !== "NavigationDuplicated") throw err
+  })
+}
+
```
wynnset, line 30: Genius!

wynnset (approved):

wynnset:
> @wynnset This solution isn't the same as the one suggested on Asana, but I think it's simpler since it doesn't require checking all the routes, it simply catches the error when navigation is duplicated. Do you think it would be more valuable to check the routes before calling `router.push`?

@jaskwong this is good, thanks!

PR #959, opened 2021-02-09
958 -  Add Setting Toggles to Allow Draft Nodes and Submitting Nodes

jaskwong:
- Creates a toggle under settings->access to enable/disable users from creating draft nodes
- Creates a toggle under settings->access to enable/disable users from submitting their nodes for review (only visible if above toggle is enabled)
- When disabled, users should see buttons to save drafts/submit nodes for review

image

Closes #959

Automated Testing
- When drafts are disabled, create draft button is not visible and users without add permission cannot see the add button 
- When submit for review is disabled, users cannot see the button to submit drafts for review

### DIFF
--- a/templates/vue/src/components/TapestryMain/TapestryNode/index.vue
+++ b/templates/vue/src/components/TapestryMain/TapestryNode/index.vue
@@ -91,9 +91,7 @@
             <tapestry-icon :icon="icon" svg></tapestry-icon>
           </node-button>
           <template
-            v-if="
-              (isLoggedIn && this.settings.draftNodesEnabled) || hasPermission('add')
-            "
+            v-if="(isLoggedIn && this.settings.draftNodesEnabled) || canEditTapestry"
           >
             <add-child-button
               v-if="!isSubAccordionRow"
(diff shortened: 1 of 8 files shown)


wynnset:
Found a bug... Even though I turned off ability to submit, I was able to see the submit node button in the sidebar and submit the node:
image

We should disable all submission UI elements and furthermore disallow submission from the backend for security purposes.

narendrasss (approved):

templates/vue/src/components/TapestryMain/TapestryNode/index.vue
```diff
@@ -90,7 +90,9 @@
           >
             <tapestry-icon :icon="icon" svg></tapestry-icon>
           </node-button>
-          <template v-if="isLoggedIn">
+          <template
+            v-if="(isLoggedIn && this.settings.draftNodesEnabled) || canEditTapestry"
```
wynnset, line 94: This is not correct logic as it is going to prevent anyone who has been given add/edit access to a node not to be able to add/edit.

wynnset (commented):
I'm going to fix this @shirleywbi but just wanted to let you know.

PR #951, opened 2021-02-02
950 - TYDE CoS #4 - Create communities tab

narendrasss:
## Changes
* Implement the add communities tab of the TYDE Circle of Support activity

## Screenshots
https://user-images.githubusercontent.com/31267630/109707174-b50b1500-7b4e-11eb-9491-78c10a185a4a.mp4

https://user-images.githubusercontent.com/31267630/107430271-c4f88180-6ad9-11eb-8e38-77ab2115201b.mp4

## Issue Linkage
Closes #950 

## PR Dependency
Depends on: #948, #960

## Automated Testing
N/A

PR #960, opened 2021-02-11
952 - TYDE CoS #5 - Add smart positioning and responsiveness for communities

### LINKED ISSUE #952: CoS #5 - Add smart positioning and responsiveness for communities

narendrasss:
## Changes
* Adds the community view with smart placement algorithm

## Screenshot
https://user-images.githubusercontent.com/31267630/109707174-b50b1500-7b4e-11eb-9491-78c10a185a4a.mp4

## Issue Linkage
Closes #952

## PR Dependency
Depends on: #951 

## Automated Testing
N/A


shirleywbi (commented):
@narendrasss Took a preliminary look.
1. When a community has no color selected, looking at SPY, should the dot be centered or should a grey dot be in place?
image
2. Will we be adding a label to indicate what this page is about/what these buttons do at any point? Out of context, I wouldn't know that + was to create a community. 
3. Also, can we fix the offset for the face?
image
4. Not sure if this is a non-issue but I made the name for a community too long (continuous string) and I now it cuts off in the Adding Connection View. I also can't edit it...
image
image
5. Should we limit it so that you can only see either the connection popup or the community popup?
image
6. Should the community popup also have a close button?
image
7. When we add an 11th community, should there be a warning or message that says it won't show up on screen? And to hide that extra piece popping up here.
image
8. Communities get cut off when you have too many. Should I be able to scroll?
image

narendrasss:
Thanks for the thorough look @shirleywbi!

> Will we be adding a label to indicate what this page is about/what these buttons do at any point? Out of context, I wouldn't know that + was to create a community.

There is an onboarding task for this — see CoS 7 on Asana.

> When we add an 11th community, should there be a warning or message that says it won't show up on screen? And to hide that extra piece popping up here.

I believe the maximum communities is 10 communities. We should disable adding any new ones when we reach that point though, good catch.

@wynnset Can you clarify some of Shirley's comments? Namely:

1. Should we force users to pick a color for their community?
2. Should there be a character limit for the community name? Or should it just overflow if it gets too long?

wynnset:
@shirleywbi those are all really good points and I think most of them except # 2 need to be addressed.
@narendrasss to answer your questions:
> 1. Should we force users to pick a color for their community?

The default color should be grey. Users can leave it be grey if they want. There should not be any "no color" option.
> 2. Should there be a character limit for the community name? Or should it just overflow if it gets too long?

Let's add a limit - something that makes sense and allows maybe 25-40 characters max (you be the judge here as to what makes sense).

narendrasss:
@shirleywbi @wynnset Thanks both — most of the issues should be fixed now.

For the emojis, it looks like we're using different emoji sets so I can't confirm if the icon is centered on your side @shirleywbi. Also keep in mind that there's currently no way to delete communities in this PR; both me and @wynnset agreed to leave that off for a separate ticket since this ticket is taking a while as is.

One thing that's left is to set a limit for connection names as well — I think this makes sense given the limit for communities, and long connection names also break the UI. Not sure how many characters would make sense though given that the space for connections is much more limited.

shirleywbi (commented):
@narendrasss Thanks for addressing the previous concerns. Just a few more testing notes:
9. The new character count goes negative if you have too many characters. The error message only appears when you click outside the input box.
image

10. Is there a limit on how many connections can be in a community? It starts getting cut off after 4. Is this supposed to be the PR where you can click on the community to expand it?
image

narendrasss:
> The new character count goes negative if you have too many characters. The error message only appears when you click outside the input box.

This is intentional, I modelled it off how Twitter shows character limits. @wynnset Let me know if you want it to look different.

> Is there a limit on how many connections can be in a community? It starts getting cut off after 4. Is this supposed to be the PR where you can click on the community to expand it?

The community's meant to show an expand icon as soon as there are connections that overflow the container, so this is a bug. Good catch.

wynnset:
As per yesterday's discussions, please have that error appear on keyup if possible but use your best judgement if need be. Thanks

shirleywbi (commented):
@narendrasss 
1. Similar to the dot handling in add connections, looks like it's missing a grey dot.
image

2. Should there be a cap on how many communities a connection can be assigned to? Or space these out/wrap the dots? Otherwise, we get overlapping.
image

3. The functionality has been disabled but should we also disable the button?
image

4. Similar to above. Update CSS for this button so that on-hover doesn't do anything when it should be disabled? Also, UI looks different between this button and the Add Community button. Should these be the same?
image

5. The clickable still doesn't seem to be working for me. I can't seem to get this to expand. 
image

narendrasss:
@shirleywbi Thanks, everything should be addressed other than the clickable thing. I can't seem to replicate it, everything works fine on my end. Can you describe exactly what you did that it doesn't show up?

shirleywbi:
> @shirleywbi Thanks, everything should be addressed other than the clickable thing. I can't seem to replicate it, everything works fine on my end. Can you describe exactly what you did that it doesn't show up?

So I'm under the impression that if I click a community, it will expand if it is full. 
Steps: 
1. Create 7 connections without a community.
2. Create 1 community.
3. Assign the 7 connections to a community through the :) tab.
4. Clicked around the community to try and get it to expand.
image

narendrasss:
@shirleywbi Hmm yeah strange. I tried replicating the steps and the expand arrow was able to show up. Any console errors?

shirleywbi:
@narendrasss Also, I've noticed that when I'm clicking the tabs, it only triggers if I click within the box. Can this be updated so it opens if I click anywhere in the tabs for both connection and community?
image

shirleywbi:
> @shirleywbi Hmm yeah strange. I tried replicating the steps and the expand arrow was able to show up. Any console errors?

No console errors. Anything local that might not be pushed?
image
Is it supposed to set an activeCommunity?

narendrasss:
> Anything local that might not be pushed?

@shirleywbi Just checked the code, everything's pushed.

So the important piece of state here is the `clickables` state. It's an object that determines if a community is expandable or not. The expand button is set to show whenever its community is expandable. Whenever the community list rerenders, we check if the community is overflowing (has hidden connections) and if it is, we set the clickable state to true.

shirleywbi:
> So the important piece of state here is the `clickables` state. It's an object that determines if a community is expandable or not. The expand button is set to show whenever its community is expandable. Whenever the community list rerenders, we check if the community is overflowing (has hidden connections) and if it is, we set the clickable state to true.

@narendrasss Seems like the issue might be scrollHeight and clientHeight. It's returning the same value for the element.
image
image

narendrasss:
@shirleywbi Just checked, looks like on chrome it doesn't appear but works fine on firefox. Will investigate, thanks!

narendrasss:
@shirleywbi That bug was super interesting! Turns out it's only not appearing for the south-facing communities:

CleanShot 2021-03-18 at 12 24 08@2x

And it's happening because in Chrome it's not possible for content to overflow _upwards_ whereas in Firefox it's possible (therefore the `scrollHeight === clientHeight` in Chrome). This eventually led me to a rabbit hole of CSS writing modes, and basic TLDR is that it's impossible to scroll a container that uses `flex-wrap: wrap-reverse` to overflow its content because it goes against the CSS writing mode.

Currently looking for a workaround, will update this PR when I get one working.

narendrasss:
@shirleywbi Should work now. Thanks again for catching it!

shirleywbi (commented):
@narendrasss @wynnset Is it an issue if connections get cut off if the screen width is too small? Or too many connections? Everything else looks good to me!
image
image

wynnset:
> @narendrasss @wynnset Is it an issue if connections get cut off if the screen width is too small? Or too many connections? Everything else looks good to me!

I don't think we need to worry about too many connections as long as we can fit at least 15 connections per community. @narendrasss on smaller screens, can we have more connections show up? I noticed it's just 1 column vs. 2 columns on larger screens - can we fit more than 1? Seems like there's a lot of padding that could be reduced potentially.

narendrasss:
> can we fit more than 1?

Made the connection width responsive to the viewport, should look better now.

CleanShot 2021-03-23 at 10 13 59@2x

shirleywbi (approved):
LGTM!

PR #962, opened 2021-02-17
953 - Change accordion to multi-content node

### LINKED ISSUE #953: Change accordion to multi-content node

shirleywbi:
## Changes
**Multi-Content Node: Changes to authoring / Node Modal**
* Update content type to "multi-content" and downgrade all "accordion" to presentation style.
* Implement multi-content in-place addition of sub-items with SubItemTable
    * Users must save nodes before adding to SubItemTable 
* Added presentation style to MultiContentForm (default: accordion) 
    * Only accordion is complete. Page and slideshow to be implemented in another PR.
* Rerouted closing of NodeModals:
    * Adding children to multi-content nodes brings you back to the parent node
    * Creating a multi-content node keeps the modal open but changes it from "Add" to "Edit"
* Bug Fix: Duplicate children upon initial render of multi-content node children
## Screenshot 
image
change.
image
image
## Issue Linkage
Closes #953
## PR Dependency
Depends on: N/A
## Automated Testing
* Add multi-content node in-place using save button
* Add/edit child using add sub-item table
* Node modal routing such that adding/editing/deleting children of a multi-content node returns to the parent node


wynnset (commented):
@shirleywbi did a quick scan of your code and it looks great thanks! The one thing I would add in here is something to update the existing nodes that are accordions so we maintain backwards compatibility. I think we can use something like [this](https://wordpress.stackexchange.com/a/144873) to automatically run this when the new version is installed. We can loop through all tapestry nodes and update the accordion ones (through PHP/Wordpress not the db). We can also utilize the tapestry settings page that @jaskwong added for this instead, but I think automatically updating is better.

narendrasss (approved):
Works well!

tapestry.php
```diff
@@ -383,4 +384,42 @@ function tapestry_tool_log_event() {
...
+        $post_id = get_the_ID();
+        $node_meta_data = get_post_meta($post_id,'tapestry_node_data', true);
+        if (isset($node_meta_data->mediaType) && $node_meta_data->mediaType == 'accordion') {
+            $node_meta_data->mediaType = 'multi-content';
+            $node_meta_data->presentationStyle = 'accordion';
+            update_post_meta( $post_id, 'tapestry_node_data',  $node_meta_data);
+        }
+    }
+}
+register_activation_hook( __FILE__, 'accordion_update' );
```
wynnset, line 414: Is it necessary to run this on activation? Aren't we running `tapestry_plugin_update` on init anyways?

tapestry.php
```diff
@@ -383,4 +384,42 @@ function tapestry_tool_log_event() {
...
+
+function tapestry_plugin_update() {
+    global $TAPESTRY_VERSION_NUMBER;
+    $installed_version = get_site_option('tapestry_version');
+    if ( $installed_version != $TAPESTRY_VERSION_NUMBER) {
```
wynnset, line 419: It might work best if we check the specific version number here instead to see if `tapestry_version` is older than the current version. That way, this wouldn't run on every version update.

tapestry.php
```diff
@@ -12,6 +12,7 @@
 
 // Used to force-refresh assets
 $TAPESTRY_VERSION_NUMBER = '2.41.0-beta';
```
wynnset, line 14: Please update this to '2.44.0-beta'

templates/vue/src/components/modals/NodeModal/forms/ContentForm/MultiContentForm/index.vue
```diff
@@ -24,13 +34,29 @@
...
+      type: String,
+      required: true,
+    },
+  },
+  data() {
+    return {
+      presentationStyle: [
+        { value: "accordion", text: "Accordion" },
+        { value: "slideshow", text: "Slideshow" },
+        { value: "page", text: "Page" },
```
wynnset, line 57: Should we remove these until we add those presentation styles?

templates/vue/src/components/modals/NodeModal/forms/ContentForm/MultiContentForm/index.vue
```diff
@@ -24,13 +34,29 @@
...
       required: true,
     },
+    actionType: {
+      type: String,
+      required: true,
+    },
+  },
+  data() {
+    return {
+      presentationStyle: [
```
wynnset, line 54: This should be plural?

wynnset (changes requested):
Overall looking great, just a few minor changes needed

tapestry.php
shirleywbi: Updated. Please let me know if this was what you're looking for.

tapestry.php
```diff
@@ -11,7 +11,8 @@
  */
 
 // Used to force-refresh assets
-$TAPESTRY_VERSION_NUMBER = '2.41.0-beta';
+$TAPESTRY_VERSION_NUMBER = '2.44.0-beta';
+add_option( "tapestry_version", $TAPESTRY_VERSION_NUMBER );
```
wynnset, line 15: Change the current version so the update kicks in. Also rename to tapestry_plugin_version so it's clear. 
```suggestion
add_option( "tapestry_plugin_version", "2.41.0-beta" );
```
I also suggest moving this to the update file itself.

tapestry-plugin-updates.php
```diff
@@ -0,0 +1,61 @@
...
+            $node_meta_data->isMultiContentChild = true;
+            update_post_meta( $post_id, 'tapestry_node_data',  $node_meta_data);
+        }
+    }
+}
+
+function tapestry_plugin_update() {
+    global $TAPESTRY_VERSION_NUMBER;
+    $installed_version = get_site_option('tapestry_version');
+    if ( version_compare($installed_version, $TAPESTRY_VERSION_NUMBER, '<') ) {
```
wynnset, line 54: Nice usage of `version_compare`! 

So what I was saying here is that we should check the `$installed_version` against the specific version where this feature was introduced so that the update doesn't run every time there's a new version, but only once. For example, with your current code, if the `$installed_version` is 2.41.0 and the user updates to 2.44.0, it will update. But then again when the next update comes out (e.g. 2.45.0) `$installed_version` is 2.44.0 and so it will run this update again, and again on any future update. Not sure if this is what you intended, because then you need to make sure in your updates that you don't accidentally make assumptions about the data, etc. If you change to the following, then the update will only run once:
```suggestion
    if ( version_compare($installed_version, '2.44.0-beta', '<') ) {
```
Let me know if any of that doesn't make sense or if I missed something!

tapestry.php
```diff
@@ -21,6 +22,7 @@
  */
 require_once dirname(__FILE__).'/endpoints.php';
 require_once dirname(__FILE__).'/tapestry-settings.php';
+require_once dirname(__FILE__).'/tapestry-plugin-updates.php';
```
wynnset, line 25: Could we please drop the `tapestry-` prefix in both these files? It's clear that it's regarding tapestry since we are in the tapestry plugin directory.

wynnset (changes requested):
Some clarifications about versions and updates.

tapestry-plugin-updates.php
shirleywbi: Good call. Definitely what I intended but I hadn't considered that $TAPESTRY_VERSION_NUMBER would continue to update.

wynnset (approved):

wynnset:
So I am getting some nodes that have their presentation style set to "accordion-row" in the frontend:
image

But checking in the database, it's set as "accordion":
image

I am flabbergasted... any clue why this is happening?

I tried deleting the tapestry_plugin_version option so the update runs again, but the update ran again and still same issue...

shirleywbi:
> Hey @shirleywbi I was just testing the upgrade to see if it's properly fixing everthing, but facing an issue:
> 
> I am getting some nodes that have their presentation style set to "accordion-row" in the frontend:
> 
> image
> 
> But checking in the database, it's set as "accordion":
> image
> 
> I am flabbergasted... any clue why this is happening?
> 
> I tried deleting the tapestry_plugin_version option so the update runs again, but the update ran again and still same issue...

@wynnset Looking into it. I think upon retrieving the node from the dataset, it assigns a presentationStyle of "accordion-row" to it (so I suspect this is happening for nested multi-content).

shirleywbi:
@wynnset Please let me know if the issue still persists. I opted to rename it rather than getting rid of it completely since #963 will be using this attribute.

wynnset:
> Please let me know if the issue still persists. I opted to rename it rather than getting rid of it completely since #963 will be using this attribute.

@shirleywbi looks good now! That sounds good.

wynnset (approved):

PR #965, opened 2021-02-23
963 - Multi-Content Node: Adding page display type

### LINKED ISSUE #963: Multi-Content Node: Adding page display type

shirleywbi:
## Changes
* Add Page style for MultiContent nodes.
    * Add Page component
    * Add SubPage component to handle "More Content". 
* Refactor AccordionMedia into MultiContentMedia, pulling out AccordionRows and PageRows as separate components
* Renaming several elements and functions to multi-content (from accordion)
* Bug Fix: Fix initial display when opening accordion and page LightBoxes
    * Added back missing activeIndex, setting accordion to show first child and page to show -1
* Bug Fix: Prevent autoplay when Multi-Content LightBox is opened for Youtube 
    * This is a work-around since `:autoplay=0` is not working.
* Bug Fix: Prevent autoplay when Multi-Content LightBox is opened for H5P Youtube
    * H5P youtube videos are based on Youtube Iframe API: https://developers.google.com/youtube/iframe_api_reference which is the same API used above.
    * I opted not to use setTimeout which could have also paused the video because there was a brief moment of sound that would play.
* Bug Fix: Console errors for when Multi-Content nodes have no children
* Add show title checkbox for page children

As of this PR, nested multi-content nodes appear with their parent node type (e.g., if I have a page MC, the sub-nodes will be page). This will be addressed in #971.

## Screenshot
ezgif com-gif-maker (5)
image
image
## Issue Linkage
Closes #963 
## PR Dependency
Depends on: N/A
## Automated Testing
- Checks that text content should render
- Checks that given a text node with a more content node, both shall render
- Checks that given a video and if rows are locked until previous rows are completed, subsequent rows are locked


narendrasss:
@shirleywbi @wynnset Is it intentional that youtube content doesn't have titles?

CleanShot 2021-03-03 at 11 45 56@2x

shirleywbi:
> @shirleywbi @wynnset Is it intentional that youtube content doesn't have titles?
> 
> CleanShot 2021-03-03 at 11 45 56@2x

@wynnset Should I add a title to video content?

wynnset:
Is this an issue just for YouTube? It should be consistent across various content types. I think we should allow the user to define whether they want to show the title for each subsection in a multi-content node.

shirleywbi:
> Is this an issue just for YouTube? It should be consistent across various content types. I think we should allow the user to define whether they want to show the title for each subsection in a multi-content node.

@wynnset It's because it's showing tapestry-media. Videos in the lightbox don't usually show the title. It only shows the title in the node before clicking. I can add the title that the user defined for the node if you'd like.

narendrasss:
Found an issue: the content doesn't seem to show when you change a node to a multi content node.

CleanShot 2021-03-30 at 13 03 27

"Root" was initially a video node, and I changed it to multi content.

shirleywbi:
> Creating a new multi content node and adding child nodes work as expected, but a few issues:
> 
> 1. the content doesn't seem to show when you change a node to a multi content node.
> 
> CleanShot 2021-03-30 at 13 03 27
> 
> "Root" was initially a video node, and I changed it to multi content.
> 
> 1. The show title checkbox isn't showing up for me:
> 
> CleanShot 2021-03-30 at 13 08 50@2x

@narendrasss Good call. I had the hardest time reproducing that first one since it only occurred when you had created non-multi-content nodes, refreshed, then changed one to multi-content since it dataset.js was hiding the childOrdering. The issues should be resolved now.

narendrasss (approved):
Looks good!

wynnset:
Locking mechanism is not working:
image
image

wynnset:
Locked node functionality is incomplete:
image
image

The locked icon is missing in the page menu. Also, if a node is locked, in the multi-content row for that node, we should show the following, centred, as the content:
- The lock icon
- "This content is current locked."
-  The conditions for its unlock (available currently when you hover over a locked node)  
image

shirleywbi:
@wynnset Added. Hope this is okay.
image

templates/vue/src/components/Lightbox/media/MultiContentMedia/AccordionRows.vue
```diff
@@ -0,0 +1,255 @@
...
+      )
+    },
+    rowBackground() {
+      if (this.isMultiContentContext) {
+        let rgb = 40
+        let colorOffset = this.level * 10
+        rgb = colorOffset > rgb ? 0 : rgb - colorOffset
+        return {
+          background: `rgb(${rgb}, ${rgb}, ${rgb})`,
+        }
```
wynnset, line 186: Cool!

templates/vue/src/components/Lightbox/media/H5PMedia/H5PIframe.vue
```diff
@@ -349,11 +352,23 @@ export default {
...
+                // Disable Autoplay For Youtube Workaround:
+                // There's a bug with the Youtube Video API such that you cannot disable autoplay.
+                // This is a workaround to stop the video upon loading.
+                if (!this.autoplay) {
+                  // As of April 2021, H5P has not implemented a stop functionality,
+                  // so there is no way to avoid the "More Videos" when paused.
+                  // This pause requires a timeout to avoid a spinner.
+                  setTimeout(() => {
+                    h5pVideo.pause()
+                  }, 1000)
```
wynnset, line 365: Instead of this, could we do a `setInterval` and check to see if the video is loaded and if it is loaded, then pause the video and clear the interval? If it's not loaded (i.e. cannot be paused), we should keep going until the video is paused. The reason I am suggesting is that for me the video loads faster than 1000ms and therefore I actually hear the video playing for a bit under a second.

wynnset (changes requested):
Just a few minor things:
- I updated the styling for a few things. I noticed you had the `<ul>` element repeating rather than `<li>` so I fixed that. I also simplified the styling syntax (I think there is some learning in there for you if you take a look at how I restructured that). Let me know if you notice any issues with any of that (I know the bullets got bigger but I like that look actually, and it’s ok that the locked menu items don’t get highlighted)
- [x] If I click to edit a child node directly in the tapestry, I expect that the close (X) button and cancel button would just close the modal, but they instead take me back up to the parent edit modal. Same thing if I save a child node. If a user is coming to the node modal from a parent node modal, we should keep track of that so we can take them back there. Otherwise, we should do the default behaviour of closing the modal.
- [x] Why is the styling for the title text of locked rows different than the ones not locked? Row titles should have consistent styling:  
image

shirleywbi:
@wynnset 
> Just a few minor things:
> 
> * I updated the styling for a few things. I noticed you had the `<ul>` element repeating rather than `<li>` so I fixed that. I also simplified the styling syntax (I think there is some learning in there for you if you take a look at how I restructured that). Let me know if you notice any issues with any of that (I know the bullets got bigger but I like that look actually, and it’s ok that the locked menu items don’t get highlighted)

Thanks. This was really helpful.

> * [X]  If I click to edit a child node directly in the tapestry, I expect that the close (X) button and cancel button would just close the modal, but they instead take me back up to the parent edit modal. Same thing if I save a child node. If a user is coming to the node modal from a parent node modal, we should keep track of that so we can take them back there. Otherwise, we should do the default behaviour of closing the modal.
> * [X]  Why is the styling for the title text of locked rows different than the ones not locked? Row titles should have consistent styling

Fixed. The styling differed because the title was being generated differently for locked and unlocked nodes. When locked, the title is generated by PageRows. When unlocked, the title is from the given TapestryMedia.

templates/vue/src/components/Lightbox/media/H5PMedia/H5PIframe.vue
shirleywbi: I tried using `setInterval` but it seems that by the time the H5P tells us the video has loaded as per loaded event found here: https://github.com/h5p/h5p-video, it may have already started playing. I opted to remove the `setTimeout` but there is a spinner.

wynnset (changes requested):
This is pretty much good to go now. I've gone through and updated the code and styling where needed myself except for a couple of things:
- I find the language "isSubMultiContentRow", "hasSubMultiContent", or "subMultiContent" really hard to understand... any way we can change that to something easier to grasp?
- I noticed that when I log in as a user with edit access to a multi-content node, I do not see the multi-content node's child nodes (I should). If I log in as the user who created the node, I can see those nodes.

Once these are addressed, please mark as "ready to merge". Thanks @shirleywbi !

shirleywbi:
> * I noticed that when I log in as a user with edit access to a multi-content node, I do not see the multi-content node's child nodes (I should). If I log in as the user who created the node, I can see those nodes.
> 
@wynnset I can't seem to replicate this. For example, I created the nodes using an admin account. I'm getting all my nodes as an editor and author.
image
image

wynnset:
@shirleywbi sorry never mind. I had only set the main node to be visible and not the subnodes, so it's ok that it didn't show up. Ignore that part of my comments!

shirleywbi:
@wynnset Since I made a couple of extra changes, I marked this as "ready for review".

PR #973, opened 2021-03-11
Merge master into TYDE (Mar 11 2021)

narendrasss:

PR #976, opened 2021-03-18
975 - Fix WP installation for CI

### LINKED ISSUE #975: Fix WP installation for CI

rlnsy:
## Changes
* Includes the WP docker environment variables in the CLI compose config. so it is consistent with the actual WP image. This addresses changes to the default values of these variables in the latest versions of the wordpress image.
## Issue Linkage
Closes #975.
## PR Dependency
Depends on: N/A

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,8 +38,6 @@ jobs:
         env:
           DOCKER_WP_INSTALL_DELAY: 60
         run: bash ./bin/await-containers.sh
-      - name: Show logs
-        run: docker-compose logs
       - name: Setup WP
         run: sh ./bin/install.sh
       - name: Setup link preview config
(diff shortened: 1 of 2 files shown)


rlnsy:
@wynnset @narendrasss this currently hard-codes the version of the wordpress image we use in docker. It's a hack, but it will fix all of the CI for the time being. The latest versions of the image changed the way that environment variables are loaded into the `wp-config.php` and it's breaking our system. I'm trying to find the root cause rn.

rlnsy:
@narendrasss @wynnset Update: found the root cause. See change description above. This is ready for review and merge in to fix CI problems throughout the project. Note that this same change will need to be done on the `tyde` branch, don't know if we should wait for the official TYDE merge, or create a duplicate PR with `tyde` as the base.

wynnset:
Thanks @rlnsy for finding the issue and fixing this! I think we can add this change as a new commit to the pending tyde merge PR #973 so it gets applied to tyde then. This looks good so I'm going to merge it in right away.

wynnset:
Update: I have committed this change to the tyde merge branch / PR #973

PR #978, opened 2021-03-22
967 - Add navigation to “Page” multi-content node

### LINKED ISSUE #967: Add navigation to “Page” multi-content node

shirleywbi:
## Changes
* Add page navigation bar to Page presentation style
    * Full-screen view shows full navigation until browser width is too small, then it converts to burger menu.
    * Lightbox view shows compressed burger menu
* Add navigation checkbox to authoring form
## Screenshot
Authoring changes:
image
image
image

Full screen view:
image
image

Light box view:
image
image

## Issue Linkage
Closes #967 
## PR Dependency
Depends on: #963
## Automated Testing
* Renders nested children titles in menu
* Given small viewport (500px), show burger menu


narendrasss:
@shirleywbi Some UX things (maybe for @wynnset):

1. Should the items on the sidebar be active depending on the currently visible node?
2. Should the items be clickable (to navigate to that node)?

CleanShot 2021-03-30 at 13 17 39

Some issues with full screen:

1. The first time I open the node in full screen, the sidebar covers the node content. It fixed itself when I resized the window.

CleanShot 2021-03-30 at 13 13 14

2. The responsive behaviour seems a bit weird, and opening the sidebar seems a bit jittery:

CleanShot 2021-03-30 at 13 15 03

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageNavigationBar/index.vue
```diff
@@ -0,0 +1,259 @@
...
+  },
+  data() {
+    return {
+      opened: false,
+      browserWidth: Helpers.getBrowserWidth(),
+    }
+  },
+  created() {
+    window.addEventListener("resize", Helpers.debounce(this.setBrowserWidth, 300))
+  },
```
narendrasss, line 72: Correct me if I'm wrong @shirleywbi, but from skimming the code it looks like the resize event is being used to size the navbar and the node contents? If so, this might be what's causing the full screen bug (because when you go full screen, the resize event doesn't trigger) and the weird resizing behaviour.

I think you could solve this UI requirement using CSS alone, which would make the code a lot simpler and offload the layout work to the browser. Here's a small example I made:

https://codesandbox.io/s/async-currying-fx4nk?file=/index.html

I might be missing something here, so I do apologize if you've already tried approach!

wynnset:
> @shirleywbi Some UX things (maybe for @wynnset):
> 
> 1. Should the items on the sidebar be active depending on the currently visible node?

Yes they should be
> 2. Should the items be clickable (to navigate to that node)?
> 

Yes they should be
> Some issues with full screen:
> 
> 1. The first time I open the node in full screen, the sidebar covers the node content. It fixed itself when I resized the window.
> 

Weird yes
> 2. The responsive behaviour seems a bit weird, and opening the sidebar seems a bit jittery:
> 

This looks fine to me, but maybe I'm missing something.

narendrasss:
> This looks fine to me, but maybe I'm missing something.

@wynnset I think I wasn't expecting there to be a delay between resizing and the elements updating, but I agree that's a bit nitpicky. It's also possible for the screen to look like this:

CleanShot 2021-03-30 at 13 57 10@2x

I left a comment above on why I think this is happening — I think ultimately if we change the code to use CSS instead of JS to do the positioning, it would be more resilient to bugs and easier to understand what's going on.

wynnset:
> > This looks fine to me, but maybe I'm missing something.
> 
> @wynnset I think I wasn't expecting there to be a delay between resizing and the elements updating, but I agree that's a bit nitpicky. It's also possible for the screen to look like this:
> 
> CleanShot 2021-03-30 at 13 57 10@2x

Yeah, this shouldn't happen.
> 
> I left a comment above on why I think this is happening — I think ultimately if we change the code to use CSS instead of JS to do the positioning, it would be more resilient to bugs and easier to understand what's going on.

I agree that we should be using CSS, not JS for styling unless absolutely necessary.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageNavigationBar/index.vue
shirleywbi: Good call. I was trying to do something similar but was having a lot of issues with `position: absolute`. There was some alignment in the code elsewhere that was making it behave differently. Should be better now. Thanks.

narendrasss (approved):

PR #986, opened 2021-03-30
977 - Add ability to create timed pop-up for Videos

### LINKED ISSUE #977: Add ability to create timed pop-up for Videos

narendrasss:
## Changes
- Allow authors to add pop-ups to video content, which includes H5P, URL-videos and YouTube videos.
- Refactor video nodes to conform to the following state machine:

VideoMedia-StateMachine(1)

### Remaining Todos

- [x] Fix state transitions for non-YouTube H5P videos
  - Specifically, try to see if we can get autoplay to work correctly. Previously, autoplay didn't work because it emits several paused/play events after the initial load, overriding the playing state transitioned to by the parent `VideoMedia`.

## Screenshot
image
image

## Issue Linkage
Closes #977 

## PR Dependency
Depends on: N/A

## Automated Testing
N/A


shirleywbi:
@narendrasss Coming across some issues:
1. Youtube: Autoplay is intermittent. Sometimes, it will start playing when I click to another part of the video (via progress bar). Other times, it doesn't.
image
2. Console errors for null in ContentForm
image
3. H5P: Skipping ahead/back pauses video. Cursor sometimes gets stuck on drag and when unpausing the video, it drags the progress bar.
4. H5P: I got stuck in a loop when trying to navigate with the progress bar/pressing play.
ezgif com-gif-maker (8)
5. 60 is a bit misleading. I assumed it was a default value. Also, should there be validation for this? I was able to submit it without entering any value and also invalid values (e.g., negative).
image
6. YT/H5P: This might be a separate bug ticket with progress update. "Continue" sometimes doesn't appear until I either wait a really long time or close my modal and reopen it.
image
7. H5P: Unable to leave gravity form to go back to video. I think the continue might be underneath the end form.
image
8. Should we be handling the case where there are two popup nodes with the same time (e.g., 5s)?

narendrasss:
> H5P: Skipping ahead/back pauses video. Cursor sometimes gets stuck on drag and when unpausing the video, it drags the progress bar.

@shirleywbi can you clarify this? This is the behaviour on my end:

https://user-images.githubusercontent.com/31267630/115604254-02297e80-a296-11eb-94b6-0f7504a5bc84.mp4

Do you mean the play screen is appearing?

shirleywbi:
> > H5P: Skipping ahead/back pauses video. Cursor sometimes gets stuck on drag and when unpausing the video, it drags the progress bar.
> 
> @shirleywbi can you clarify this? This is the behaviour on my end:
> 
>  CleanShot.2021-04-21.at.11.37.09.mp4 
> Do you mean the play screen is appearing?

Yes.

narendrasss:
@shirleywbi I added some fixes, namely adding validation and some UX improvements. I wasn't able to replicate most of the bugs you show though. Can you take another look?

Here's a video of me answering a quiz at the end of the video. Can you see if there's anything I'm missing here?

https://user-images.githubusercontent.com/31267630/115615061-4cfdc300-a2a3-11eb-9e88-430208e8b4f3.mp4

> YT/H5P: This might be a separate bug ticket with progress update. "Continue" sometimes doesn't appear until I either wait a really long time or close my modal and reopen it.

I added a loading indicator for this to make it more obvious that it's completing in the background.

> Yes. But I assume that's expected behaviour based on what you're showing me.

It's a side effect because the video gets paused by the browser when you seek. I agree it's not very elegant UX-wise though, and I was thinking if we even need the play screen in the first place since the video already has its own play/pause indicator. The play screen also blocks the user from seeking the progress bar if the video's paused. What do you think @wynnset?

templates/vue/src/components/modals/NodeModal/index.vue
```diff
@@ -843,6 +843,15 @@ export default {
         )
       }
 
+      if (this.node.popup) {
+        const { time } = this.node.popup
+        if (time == undefined) {
```
shirleywbi, line 848:
```suggestion 
        if (time === "") {
```
`time` appears as an empty string rather than undefined.

shirleywbi (commented):
@narendrasss Did some additional testing with the update. Seems like the flashing was probably just CPU issues on my end. I closed some tabs and it stopped flashing. 

General:
1. Should "Continue" wait until the form is finished loading before continue appears? Example uses a WP post as popup but may apply to any media that requires loading.
image
2. Clicking rewatch sometimes doesn't restart the video. 
image
3. Unable to continue for Activity popup in the *middle* of the video.
ezgif com-gif-maker (9)
4. Can we have video popups? If so, there's no way to navigate back to the original video.
image

H5P:
5. Seeking by dragging in a video is "sticky". 
ezgif com-gif-maker (10)

shirleywbi:
@narendrasss Also, for MP4s, the seeking behaviour seems to trigger popups unexpectedly. For instance, if I start the video at 0:00, click 0:20, I wouldn't expect the text popup at 5s. 
Example below:
image
ezgif com-gif-maker (11)
I've also tried extending the text node to appear at 7s in case the current time was too close to the popup time but I still get the text popup. If this is intentional in preventing people from skipping over the popups, then we might need to ensure that the subsequent popups (at 10/15/20s) are not skipped too.

Also came across an issue with updating the progress: 
When changing a node from one video to another video, the progress doesn't seem to be updating. It starts the video wherever the old video ended.

narendrasss:
> Also, for MP4s, the seeking behaviour seems to trigger popups unexpectedly. For instance, if I start the video at 0:00, click 0:20, I wouldn't expect the text popup at 5s.

Ah good catch, I've only tested it with dragging, not clicks.

> When changing a node from one video to another video, the progress doesn't seem to be updating. It starts the video wherever the old video ended.

I noticed this issue before as well, I think it's out of scope for this ticket.

templates/vue/src/components/modals/NodeModal/index.vue
narendrasss: Good catch!

narendrasss:
> Clicking rewatch sometimes doesn't restart the video.

Fixed! The YouTube player is a bit finicky, thanks for catching this.

> Should "Continue" wait until the form is finished loading before continue appears? Example uses a WP post as popup but may apply to any media that requires loading.

I think this is fine assuming the popup was previously completed. It would be a problem if the continue button showed up even though the popup isn't completed, however.

> Unable to continue for Activity popup in the middle of the video.
> Can we have video popups? If so, there's no way to navigate back to the original video.

Both of these issues should be fixed now. I've also hooked up the `close` button to continue the video.

> Seeking by dragging in a video is "sticky".

Strange - I don't have this problem. An H5P issue maybe?

> Also, for MP4s, the seeking behaviour seems to trigger popups unexpectedly.

Should be fixed now.

shirleywbi:
> > Should "Continue" wait until the form is finished loading before continue appears? Example uses a WP post as popup but may apply to any media that requires loading.
> 
> I think this is fine assuming the popup was previously completed. It would be a problem if the continue button showed up even though the popup isn't completed, however.

@narendrasss I think this one was happening regardless of whether it was completed or not.

templates/vue/src/components/Lightbox/media/VideoMedia/UrlVideoMedia.vue
```diff
@@ -0,0 +1,156 @@
...
+      @pause="handlePause"
+      @seeked="handleSeek"
+      @seeking="$emit('seeking')"
+      @timeupdate="updateVideoProgress"
+    ></video>
+  </div>
+</template>
+
+<script>
+import client from "@/services/TapestryAPI"
```
shirleywbi, line 19: Can we add a comment to explain what type of media this is? (e.g., how is this different from YoutubeMedia?)

shirleywbi: Seems like from `VideoMedia` this is the uploaded videos?

templates/vue/src/components/Lightbox/media/VideoMedia/index.vue
```diff
@@ -0,0 +1,343 @@
...
+    },
+    events() {
+      return VideoEvents
+    },
+    node() {
+      return this.getNode(this.nodeId)
+    },
+    videoComponent() {
+      switch (this.node.mediaFormat) {
+        case "mp4":
```
shirleywbi, line 149: Same as before, will we not be handling non-mp4 files?

shirleywbi (changes requested):
@narendrasss Came across a few more edge cases: 
1. Broken non-video H5P (e.g., 360 Virtual Tour).
image
image

2. Youtube video autoplaying in accordion with the screen overlay. Expected the video not to play.
image

templates/vue/src/components/Lightbox/media/TapestryMedia.vue
```diff
@@ -135,6 +112,9 @@ export default {
     node() {
       return this.getNode(this.nodeId)
     },
+    isVideoNode() {
+      return ["mp4", "youtube", "h5p"].includes(this.node.mediaFormat)
```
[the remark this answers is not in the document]
shirleywbi: @wynnset Can you confirm?

shirleywbi:
@narendrasss Retested these too. Most of them look good. Just a minor fix and follow-up on the strange H5P behaviour.
> > Unable to continue for Activity popup in the middle of the video.
> > Can we have video popups? If so, there's no way to navigate back to the original video.
> 
> Both of these issues should be fixed now. I've also hooked up the `close` button to continue the video.
> 
Looks almost perfect. Just need to fix a loading state after clicking "Done" and video continues.
image

> > Seeking by dragging in a video is "sticky".
> 
> Strange - I don't have this problem. An H5P issue maybe?
> 
Strange. It's possible. It could also be dependent on the browser. I'm only having this issue with H5P on Chrome. The behaviour seems fine on Firefox.

While looking into it, I also noticed that "timeupdate" events are being sent a lot faster compared to master branch. There are even events being sent while the video is paused. Since there's something listening to these events, it might affect performance. I'm not sure if this needs to be fixed...

templates/vue/src/components/Lightbox/media/TapestryMedia.vue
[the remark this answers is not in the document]
wynnset: I think that is fine for now.

wynnset:
I managed to fix the merge conflicts and get this PR working. Could someone please do some testing on this maybe @EldadZZipori @Frozemint ? At minimum do the following:
- [ ] Create some various types of video nodes including regular video upload, youtube video, H5P interactive video, and H5P interactive video with YouTube and test all those nodes and make sure the pop-ups work, and the play, pause, seek, and all other video functionality also works.
- [ ] Add some multi-content nodes, maybe some with videos in them with pop-ups in them and make sure that works
- [ ] I see that there are a lot of requests sent for saving the progress and h5p settings. Could you see if you can look into this? H5P settings save request should only be sent if the settings were changed. Video progress only needs to get updated every 2 seconds. Right now it's running multiple times per second which actually is a huge load on the server and causes the progress not to get saved properly.

Let me know if you need any help with this!

EldadZZipori:
**Regular video upload** 
- [ ] Text popup  
- [ ] Video popup - Functions weird, when second video pops up you cannot go out of the popup (youtube upload)
- [ ] External Link popup - both embedded and new window
- [ ] Multi-content popup with three sub-items, video-text-link

**Youtube video upload**
- [ ] Text popup  
- [ ] Video popup - Functions weird, when second video pops up you cannot go out of the popup (regular upload)
- [ ] External Link popup - both embedded and new window
- [ ] Multi-content popup with three sub-items, video-text-link

EldadZZipori:
Allowing multi-content node as a pop-up is now disabled.
Currently the multi-content nodes, with videos in them have several bugs related to the popups.

1. Popup appears as more content for the video and not only as a pop-up
Screenshot from 2021-06-01 09-23-55

2. Popup appears only from the second play
ezgif com-gif-maker

3. Cannot continue unless opening the sub-content
ezgif com-gif-maker(1)

EldadZZipori:
Test & Notes.

- For now we do not allow popups inside multi-content videos as it has some bugs (although some of it such as not displaying the popup as part of the accordion is fixed). We need to test that popups are not an option for videos and h5p inside multi-content.
- For now we also do not allow multi-content nodes to be used as popups - should also be tested.
- General functionality of all other nodes as popups for videos and h5p.

harry02031:
Youtube video(url) upload

- [ ]    Text popup - Working correctly

- [ ]    Video popup - Second video pop ups work as expected but the continue button only appear around the end of the video(around last two mins of the pop up video) during the first play of the video only. However, the continue button will always be visible in the viewings after the first play. 

- [ ]   External Link popup - new window and embedded links are working as expected(embedded link only supports HTTP sites not HTTPS sites)

- [ ]  WordPress Post type popup working

- [ ]  h5p type popup, the popup is working but here is no continue button visible, thus you cannot skip out of this popup window, you have to manually close the window and click on the youtube video again

- [ ]    Multi-content type popup (disabled as per Eldad stated above), however if you add a multi-content popup with multiple items, it will still pop up in the video around the start of the video (Bug)

harry02031:
Regular video upload
  - I can only upload a video of my own after uploading a youtube video. If I were to create a video type node, and try to upload a video of my own, it says invalid video url

- [ ]  Text popup - Working correctly

- [ ] video popup -  youtube video pop ups work as expected but the continue button only appear around the end of the video(around last two mins of the pop up video) during the first play of the video only. However, the continue button will always be visible in the viewings after the first play.  On my end, if you have another regular video upload as a popup, you cannot see the video progress bar anymore.


https://user-images.githubusercontent.com/30706624/120942820-2e7f4a00-c6e0-11eb-8f1f-1440fc462ec0.mov



- [ ]  External Link popup - new window and embedded links are all working as expected(embedded link only supports HTTP sites not HTTPS sites)

- [ ]  WordPress Post type popup workin

- [ ] h5p type popup - Same problem as with youtube video upload, there is no continue button visible and thus cannot skip out of the popup
- [ ] Multi-content type popup (disabled as per Eldad stated above)

EldadZZipori:
TODO:

Urgent - 
- [ ] Add regular H5P content component 
- [ ] Get an event when H5P content is completed

After - 
- [ ] Unable to see media-bar when using H5P video as popup for a regular video file
- [ ] Interacting with H5P content pauses the video

Frozemint:
Tested, and many issues I have noticed has been pointed out and fixed by @Enid. This looks good.

wynnset:
Found an issue that doesn't exist on master: If the screen aspect ratio is wider than the video aspect ratio, the video gets cut off:
image

.gitignore
```diff
@@ -7,3 +7,5 @@ cypress.json
 
 data/
 bin/config.sh
+
+vendor/
```
wynnset, line 11: This shouldn't be in this PR.

templates/vue/src/components/Lightbox/media/VideoMedia/H5PVideoMedia/H5PIframe.vue
```diff
@@ -6,9 +6,8 @@
       'context-multi-content': hasMultiContentContext,
     }"
     :style="{
-      height: frameHeight ? frameHeight + 'px' : 'auto',
-      width: frameWidth ? frameWidth + 'px' : '100%',
-      opacity: loading ? 0 : 1,
+      height: `${dimensions.height}px`,
+      width: '100%',
```
wynnset, line 10: This is probably what's causing the videos to be full width, even when they shouldn't be. I found that even if I set this to `${dimensions.width}px`, the iframe inside will still be cut off, so you might need to figure out when the width is off and refresh the iframe so it fits in properly.

templates/vue/src/components/Lightbox/media/VideoMedia/H5PVideoMedia/H5PIframe.vue
```diff
@@ -93,6 +99,39 @@ export default {
...
+      }
+    },
+    updateVideoProgress() {
+      const video = this.getInstance()
+
+      /**
+       * An H5P video is just one of the many types H5Ps can be, so we only want to
+       * trigger this if the H5P is in fact a video.
+       */
+      if (video) {
```
wynnset, line 118: Is this true anymore? Should we still have this check in place if we know it's a video?

templates/vue/src/components/Lightbox/media/VideoMedia/H5PVideoMedia/H5PIframe.vue
```diff
@@ -232,21 +253,51 @@ export default {
...
+    handleVideoLoad(currentTime) {
+      this.$emit("load", {
+        type: "h5p-video",
+        currentTime,
+      })
+      this.$nextTick(() => {
+        this.setFrameDimensions()
       })
     },
     handleLoad() {
```
wynnset, line 303: There is some code here that assumes h5p might not be an interactive video (E.g. `if (this.library !== "H5P.InteractiveVideo")`, `h5pInstance.recorder`, `switch (this.library)`) these all should be removed since this component is only called for H5P interactive videos.

templates/vue/src/components/Lightbox/media/VideoMedia/H5PVideoMedia/index.vue
```diff
@@ -0,0 +1,81 @@
+<template>
+  <div class="container">
+    <h5p-iframe
```
wynnset, line 3: Can we merge H5PIframe component into this component? I don't see a reason to have 2 separate components when this component is so minimal and the H5PIframe component is not used by any other components.

wynnset (changes requested):

templates/vue/src/components/Lightbox/media/VideoMedia/H5PVideoMedia/H5PIframe.vue
EldadZZipori: This one seems to be the only check that is necessary throughout the component. When clicking an interactive content on the video and the closing it the interval keeps running. I added an else clauses to stop the interval if the video does not exists. 

To sum it up, the check is necessary for something else and this comment is inaccurate.

templates/vue/src/components/Lightbox/media/VideoMedia/H5PVideoMedia/index.vue
odotadot: Merged the two components together. It seems that the H5PVideo media never $emits("complete") to be caught by VideoMedia.vue.

templates/vue/src/components/Lightbox/media/VideoMedia/H5PVideoMedia/H5PIframe.vue
EldadZZipori: This is now open as a separate PR - 1059

wynnset (changes requested):

classes/class.tapestry.php
```diff
@@ -173,6 +174,11 @@ public function addNode($node)
             $this->rootId = $node->id;
         }
 
+        $nodeArray = [$node];
+        $nodeArray = $this->_addH5PMeta($nodeArray);
+        $node = array_values($nodeArray)[0];
+        $node = $tapestryNode->save($node);
+
```
wynnset, line 181: @odotadot we should not be saving h5p meta to the database because h5p meta might change outside of the tapestry context. This information needs to get passed to the frontend dynamically instead. I know this fix was put in to ensure that h5p meta is updated when a node is edited, but this is not the right way to do it. When a node is edited, we should return the new node information and dynamically retrieve and return the h5p meta for that node as well. The frontend should then take that information and update vuex with that.

wynnset (changes requested):
Found a bug: I am seeing the play screen on top of a YouTube video (not H5P). It autoplayed the video, but the play screen (the play button) remained on the video: 
image

EldadZZipori (commented):
Progress call reduction works

classes/class.tapestry.php
odotadot: Removed those 4 lines of code. Everything still working fine (the bug that this code fixed is covered by other code in h5pForm.)

templates/vue/src/components/modals/NodeModal/forms/ContentForm/index.vue
```diff
@@ -53,8 +53,8 @@
     <b-form-group v-if="node.hasSubAccordion" label="Subaccordion Text">
       <b-form-input v-model="node.typeData.subAccordionText"></b-form-input>
     </b-form-group>
-    <b-form-group v-if="isPopupCandidate" label="Popup">
-      <popup-form :node="node" />
+    <b-form-group v-show="isPopupCandidate" label="Popup">
+      <popup-form :node="node" :isCandidate="isPopupCandidate"/>
```
wynnset, line 57: Props should be in kebob-case so `:is-candidate`

templates/vue/src/components/Lightbox/media/VideoMedia/H5PVideoMedia.vue
```diff
@@ -210,7 +210,11 @@ export default {
          */
         newSettings.caption = {}
       }
-      if (Helpers.isDifferent(newSettings, this.h5pSettings)) {
+      if (
+        this.h5pSettings &&
+        this.newSettings &&
```
wynnset, line 215: What was the error here? Why not pass this check into `Helpers.isDifferent` instead? Also, if two passed variables are both undefined or false, they are technically equal.

odotadot: `this.h5pSettings` and` this.newSettings` are sometimes both null, generating an error in the console when Helper.isDifferent hits `const srcKeys = Object.keys(src)`. I could make a statement in Helper.isDifferent like `if(src == null && other == null) {return false}` instead.

wynnset: @odotadot ok I see that the `isDifferent` function currently only checks objects (which is a bad name for that function). We could extend this function to also check more basic variables (and maybe even arrays) as well and check to see if it's an object instead of assuming that it is one. You do not need to check against `null` or anything like that because if you are doing a check with `===` it will automatically check the type and value for you.

odotadot: I think it makes more sense to keep it the way it currently is. If we move the logic into Helpers.isDifferentObject, it will have to return false (to prevent updating the h5psettings), which seems weird (two null objects should be considered the same).

wynnset (approved):

PR #990, opened 2021-04-16
989 - Upload video content to Kaltura instead of Wordpress

### LINKED ISSUE #989: Upload video content to Kaltura instead of Wordpress

shirleywbi:
## Changes
* Add upload Kaltura video endpoint
* Add thumbnail popup for non-Youtube videos
* Add composer for PHP KalturaClient
* Add cancel behaviour for upload request from client to server. Once the request reaches the server, it cannot be cancelled and the video will be uploaded.
    * This has been attempted with `media->cancelReplace`, `uploadToken->delete`, `session->end`. None of these endpoints were able to stop the upload.
* **Update README with Kaltura setup instructions**
* [ ] Update CI to work with Kaltura update
## Issue Linkage
Closes #989 
## PR Dependency
Depends on: N/A
## Automated Testing
* N/A


wynnset:
@rlnsy assigning this to you as it requires fixing CI (per the conversation with @shirleywbi on Slack).

rlnsy:
Investigation into failing tests revealed that similar API errors are produced when running locally. Please note that on my testing setup (Mac 10.15, chrome, headless) many of the API status codes were 404 as opposed to 500 (which seem to pop up in CI). This hints at a breaking change to the API in this PR. @wynnset Maybe et me know if you can reproduce the results.

rlnsy:
Example error trace:
```
CypressError: `cy.request()` failed on:

http://localhost:8000/wp-json/tapestry-tool/v1/tapestries

The response we received from your web server was:

  > 500: Server Error

This was considered a failure because the status code was not `2xx` or `3xx`.

If you do not want status codes to cause failures pass the option: `failOnStatusCode: false`

-----------------------------------------------------------

The request we sent was:

Method: DELETE
URL: http://localhost:8000/wp-json/tapestry-tool/v1/tapestries
Headers: {
  "Connection": "keep-alive",
  "user-agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/90.0.4430.212 Safari/537.36",
  "accept": "*/*",
  "accept-encoding": "gzip, deflate",
  "content-type": "application/json",
  "content-length": 19
}
Body: {"title":"cypress"}

-----------------------------------------------------------

The response we got was:

Status: 500 - Server Error
Headers: {
  "date": "Tue, 18 May 2021 05:24:11 GMT",
  "server": "Apache/2.4.38 (Debian)",
  "x-powered-by": "PHP/7.4.19",
  "expires": "Wed, 11 Jan 1984 05:00:00 GMT",
  "cache-control": "no-cache, must-revalidate, max-age=0",
  "content-length": "279",
  "connection": "close",
  "content-type": "application/json; charset=UTF-8"
}
Body: {
  "code": "internal_server_error",
  "message": "<p>There has been a critical error on this website.</p><p><a href=\"https://wordpress.org/support/article/faq-troubleshooting/\">Learn more about troubleshooting WordPress.</a></p>",
  "data": {
    "status": 500
  },
  "additional_errors": []
}


https://on.cypress.io/request

Because this error occurred during a `before each` hook we are skipping all of the remaining tests.
      at http://localhost:8000/__cypress/runner/cypress_runner.js:158829:21
      at tryCatcher (http://localhost:8000/__cypress/runner/cypress_runner.js:10325:23)
      at Promise._settlePromiseFromHandler (http://localhost:8000/__cypress/runner/cypress_runner.js:8260:31)
      at Promise._settlePromise (http://localhost:8000/__cypress/runner/cypress_runner.js:8317:18)
      at Promise._settlePromise0 (http://localhost:8000/__cypress/runner/cypress_runner.js:8362:10)
      at Promise._settlePromises (http://localhost:8000/__cypress/runner/cypress_runner.js:8442:18)
      at _drainQueueStep (http://localhost:8000/__cypress/runner/cypress_runner.js:5032:12)
      at _drainQueue (http://localhost:8000/__cypress/runner/cypress_runner.js:5025:9)
      at Async.../../node_modules/bluebird/js/release/async.js.Async._drainQueues (http://localhost:8000/__cypress/runner/cypress_runner.js:5041:5)
      at Async.drainQueues (http://localhost:8000/__cypress/runner/cypress_runner.js:4911:14)
  From Your Spec Code:
      at Context.eval (http://localhost:8000/__cypress/tests?p=cypress/support/index.js:2582:13)
```

wynnset:
@EldadZZipori I've updated some parts of this and tested locally. IT WORKS!
There are 2 things needed to finish this PR:
1. [ ] **Make Kaltura uploading optional**. Videos should upload to Kaltura ONLY if the Kaltura config variables are defined (see the updated README). Otherwise, we should do what we are currently doing (i.e. upload to localhost).
2. [ ] **Make this work with Docker**. Currently I am not sure if it works with our Docker setup. I believe you have the Docker setup on your end, so please work to make it compatible with Docker.

We don't need to add automated test for Kaltura. If we make Kaltura upload optional, the automated tests should work since they do not use Kaltura. We also therefore do not need to run the composer commands for automated testing.

odotadot:
@EldadZZipori The current commit does not work on my end unless these 3 changes are made:

1. In wp.config: `define('KALTURA_LOAD', true);`
2. In wp.config: changing the KALTURA_SERVICE_URL to use 'http' instead of 'https'
3. In "services\class.kaltura-api.php": Adding `set_time_limit(300);` to line 109 right above` sleep(5)`

Once these 3 changes are added, everything works great, and all local tests pass. I think (2) might have to be looked into. If I change it back to 'https', it will throw a 500 error.

EldadZZipori:
@odotadot 
I fixed No.1, I accidentally wrote the wrong variable name in the read.me

Regarding point number two I am able to upload with http and https to Kaltura. We should probably troubleshoot trough it to see why is it being blocked. @harry02031 @Frozemint  Could one of you check this as well so we will be able to narrow it to a setup issue or code issue?

Point three is interesting as it does not occur for me as well, why 300 specifically? What error are you getting without that line of code - just http error 500?

odotadot:
For (2) I get this error: ` Kaltura Client Error: Kaltura\Client\ClientException: SSL certificate problem: unable to get local issuer certificate. RC : 0`
For (3) I get this error: 'PHP Fatal error:  Maximum execution time of 30 seconds exceeded'

I thought 300 (5 minutes) was a good max on uploading time. I tried 90 (1min 30 secs) for a 8MB video and it still works, but I thought to be more generous with the time limit in case of a bigger file size.

odotadot:
This commit now works with the below changes in the php.ini file. For solving the http/https issue I used this: https://stackoverflow.com/questions/28858351/php-ssl-certificate-error-unable-to-get-local-issuer-certificate

3. Additional Configurations for the 'php.ini' file if using MAMP:
    
```
    post_max_size = 64M
    upload_max_filesize = 64M
    max_execution_time = 90  
    
    // Download the cacert.pem file from here: https://curl.se/docs/caextract.html and move it into C:/MAMP
    // make sure this is uncommented
    extension=php_openssl.dll

    // Add these configuration lines
    [SSL]
    curl.cainfo="C:/MAMP/cacert.pem"
    openssl.cafile="C:/MAMP/cacert.pem"
```

The video uploading is fine, but the thumbnail behaviour is a bit weird. You can only remove the thumbnails if you click 'Cancel' when the thumbnail dialogue popup happens.


In addition, whenever you click 'Publish' after editing  the thumbnail dialogue popups always, even when you are not doing anything with the thumbnails.
image

EldadZZipori:
I believe this is the wanted functionality as we want the user to be able to use the picture of the video as a thumbnail in the future as well. This gives him that options even after he cancels the thumbnail. @wynnset can you please comment on this? 

Thanks for the update on the php.ini I will make a README

wynnset:
> For solving the http/https issue I used this: https://stackoverflow.com/questions/28858351/php-ssl-certificate-error-unable-to-get-local-issuer-certificate
> 
> Additional Configurations for the 'php.ini' file if using MAMP:

@odotadot can you please document this on Notion in case someone has problems with this in the future?
> The video uploading is fine, but the thumbnail behaviour is a bit weird. You can only remove the thumbnails if you click 'Cancel' when the thumbnail dialogue popup happens.

This is expected behavior.
> In addition, whenever you click 'Publish' after editing the thumbnail dialogue popups always, even when you are not doing anything with the thumbnails.

This should **not** happen all the time @EldadZZipori . That should only show up if someone is uploading a new video or otherwise changing the content type in a way that causes a new thumbnail to be available. If a user changes the thumbnail to something else, then they are stuck with that thumbnail (for now) and cannot "revert". Maybe we can create a new task in Asana backlog to add a new feature for auto-retrieving the thumbnail using a button. But for now, let's not bother the user with this confirmation in this case.

EldadZZipori:
The core problem seems to come from the Kaltura API. 

Line 1494 in endpoints.php - 
**$result = $kalturaApi->uploadKalturaVideo($file, $category);**

This is the line in Kaltura that uploads the file before Kaltura starts converting the file to different formats. 

I believe we should test the timing on the real server and see how long it would take, After that it would be easy to asses what we should do next. (i.e. not waiting for the file to be converted)

odotadot:
Tested video uploading for small videos (~30 MB) and thumbnail behaviour, everything works as expected. All local tests pass.

EldadZZipori:
Outline of how uploading now works - 
image(3)

EldadZZipori:
You are completely right, I moved it to TapestryNode

EldadZZipori:
Updated Outline
image

templates/vue/src/components/modals/NodeModal/index.vue
```diff
@@ -348,7 +359,7 @@ export default {
     linkHasThumbnailData() {
       return (
         (this.node.mediaType === "url-embed" && this.node.behaviour !== "embed") ||
-        this.node.mediaFormat === "youtube"
+        this.node.mediaType === "video"
```
wynnset, line 362: Should this not be:
```suggestion
        this.node.mediaFormat === "youtube" ||
        (this.node.mediaType === "video" && this.useKaltura)
```

templates/vue/src/components/modals/NodeModal/index.vue
```diff
@@ -690,10 +708,17 @@ export default {
...
+            (this.type === "add" || this.originalUrl !== this.node.typeData.mediaURL)
+          ) {
+            this.confirmThumbnailsPopup(this.node.typeData.thumbnailURL)
+          } else {
+            await this.setLinkData()
+          }
         }
 
-        if (this.shouldReloadDuration()) {
+        if (this.shouldReloadDuration() && this.kaltura_info === null) {
```
wynnset, line 721: Can we move `this.kaltura_info === null` inside of `shouldReloadDuration()` please?

templates/vue/src/components/modals/NodeModal/index.vue
```diff
@@ -892,29 +927,38 @@ export default {
       if (evt === "multi-content") this.node.presentationStyle = "accordion"
     },
     async setLinkData() {
-      if (shouldFetch(this.node.typeData.mediaURL, this.node)) {
+      if (
+        this.kaltura_info === null &&
```
wynnset, line 931: Can we move this inside of `shouldFetch()` please?

templates/vue/src/components/modals/NodeModal/index.vue
```diff
@@ -892,29 +927,38 @@ export default {
...
-            this.node.imageURL = data.image
-          }
-          if (
-            confirm(
-              "Would you like to use the link preview image as the locked thumbnail image?"
-            )
-          ) {
-            this.node.lockedImageURL = data.image
+
+          if (this.type === "add") {
```
wynnset, line 940: Hmm.. What if the user is changing the media? Should we not ask them in that case as well?

templates/vue/src/components/modals/common/FileUpload.vue
```diff
@@ -83,8 +83,8 @@
               required
               data-qa="import-file-input"
               @dragover.prevent
-              @drop.prevent="uploadFile"
-              @change="uploadFile"
+              @drop.prevent="handleFileUpload($event)"
+              @change="handleFileUpload($event)"
```
wynnset, line 87: I think this can be changed to simply
```suggestion
              @drop.prevent="handleFileUpload"
              @change="handleFileUpload"
```

templates/vue/src/components/modals/common/FileUpload.vue
```diff
@@ -110,14 +110,27 @@
...
+            OK
+          </b-button>
+        </b-col>
+      </b-row>
+    </b-alert>
+    <b-alert v-else-if="!confirmedUploadToKaltura" show variant="success">
+      <b-row>
+        <b-col cols="auto" class="upload-label mr-auto text-muted">
+          File will now upload to Kaltura in the background! You can now publish the
+          node and it will become available when the file is finish uploading.
```
wynnset, line 130: I don't think we need to include the user in this level of detail. We can just let them know the video they uploaded is being processed in the background. No need to mention Kaltura.

wynnset (changes requested):
Great job with this. It's really coming along. I think we still a bit more refactoring to be amazing. It's normal for a PR of this size to have a few back and forths like this. Please see my review comments. Now that I am almost done with the review, I have some insight that I'd like to share that will mean some of my review comments might not apply:

I think we should abstract the fact that it's Kaltura that's processing the information from the frontend. Basically, I think the frontend should only be concerned with 2 things:
1. Whether the video is uploaded to the backend (i.e. NodeModal can be closed)
2. Whether the video is ready to be viewed

The "ready" state of the video would depend on the backend, so it doesn't matter to the frontend _why_ the video is not ready. If the video is not ready, it means it's "being processed" and that's it. If we move all of that to the backend, the frontend will not need to keep track of much, except for the video status which is either "ready" or "processing". So we should remove all references to Kaltura from the frontend. This will make the code much simpler and also allow us to use other video platforms in the future, should we choose to do so.

.github/workflows/ci.yml
```diff
@@ -8,6 +8,12 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Cache Composer dependencies
+        uses: actions/cache@v2
+        with:
+          path: /tmp/composer-cache
+          key: ${{ runner.os }}-${{ hashFiles('**/composer.lock') }}
+      - uses: php-actions/composer@v5
```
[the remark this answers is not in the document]
EldadZZipori: Correct, these are not needed.

templates/vue/src/services/KalturaAPI.js
```diff
@@ -0,0 +1,72 @@
+import axios from "axios"
```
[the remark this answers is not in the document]
EldadZZipori: Are you taking about the KalturaAPI as a whole?

EldadZZipori:
This is superseded by #1052

PR #993, opened 2021-04-21
992 - TYDE CoS #7 - Community view onboarding

### LINKED ISSUE #992: TYDE CoS #7 - Community view onboarding

narendrasss:
## Changes
This PR introduces onboarding for the community view of the TYDE Circle of Support.

## Screenshot
TODO

### State Machine

This PR implements the onboarding flow using the following state machine:

CoS_OnboardingFlow

This state machine is defined and managed using the APIs provided by the [xstate](https://xstate.js.org/docs/) library. The entire machine is divided into two sub-machines, `Communities` and `Connections`, that represent the onboarding flow for the add community and add connection flows respectively. This machine is defined in the [CommunityView/onboardingMachine.js](https://github.com/wynnset/tapestry-wp/blob/d82117f111a0e05101a3ee1c5f99046b00fd8a63/templates/vue/src/components/tyde/activities/CircleOfSupport/CommunityView/onboardingMachine.js) file, separate from the Vue component.

By decoupling the state management from the view layer, we can drastically simplify the logic in the Vue component itself. The only thing the Vue component has to do is display the screen that matches the current state and send events to the state machine. Both of these methods are defined in the `CommunityView.vue` component:

https://github.com/wynnset/tapestry-wp/blob/d82117f111a0e05101a3ee1c5f99046b00fd8a63/templates/vue/src/components/tyde/activities/CircleOfSupport/CommunityView/index.vue#L134-L139

The onboarding state itself is kept inside `CommunityView`'s data option, where `service` represents the running state machine and `current` represents the current machine state:

https://github.com/wynnset/tapestry-wp/blob/d82117f111a0e05101a3ee1c5f99046b00fd8a63/templates/vue/src/components/tyde/activities/CircleOfSupport/CommunityView/index.vue#L100-L103

### Views

As mentioned, decoupling the machine from the view lets us keep the view as simple as possible. As an example, showing the `AddConfirmation` screen is just a matter of checking if the state matches `AddConfirmation`:

https://github.com/wynnset/tapestry-wp/blob/d82117f111a0e05101a3ee1c5f99046b00fd8a63/templates/vue/src/components/tyde/activities/CircleOfSupport/CommunityView/index.vue#L32-L36

Capturing and handling events is simple as well. As you can see from the snippet above, all we have to do is call `send` with the right event name and the machine handles the rest.

## Todos

Currently, only the state management and some basic views are implemented. The remaining work includes:

- [x] Implementing and styling the onboarding views (both communities and connections)
- [x] Ensure the state flow matches the figma wireframes
- [x] Implement the starting transitions
  - Currently, the onboarding flow starts at `Communities.Welcome` regardless of how much onboarding the user has completed. Instead, we want the starting state to depend on the current CoS state. These transitions are represented as the three lines that go out of "start" in the state diagram.

## Issue Linkage
Closes #992 

## PR Dependency
Depends on: #970 

## Automated Testing
TODO


EldadZZipori:
Failing tests are not related to TYPE.
From what I saw they are failing because cypress is unable to find the close button when there are changes to the node or that it prompts for discarding changes although no changes have been done.

harry02031:
Screen Shot 2021-05-26 at 3 45 17 PM

Everything else works according to this state diagram but I noticed that I didn't get any (add more confirmation) message. Not sure if this stage is missing or should include an instruction or some sort to be displayed on the screen(and there is a maybe later option to exit out of this stage as displayed on state diagram). 
This is what is displayed on my end.

https://user-images.githubusercontent.com/30706624/119741207-16830d00-be3a-11eb-84fb-738ab8320238.mov

EldadZZipori:
So there is a simulation of how this should work - https://www.figma.com/proto/l7Zq0bLJD9AU08PlLZ1559/TYDE-Community-Map-Mockup?node-id=64%3A142&scaling=scale-down it was on Asana not here for some reason. The more configuration for the connection is not presented there so I though it should not (also felt a bit unnecessary to me).
It should show another view when you close the connection tab (and should present different amount of connection dependent on how many connection you have created). 
It is also important to check that the on-boarding starts from the connection part if the user already has communities.

harry02031:
Okay got it, everything looks good according to https://www.figma.com/proto/l7Zq0bLJD9AU08PlLZ1559/TYDE-Community-Map-Mockup?node-id=64%3A142&scaling=scale-down, all the views are present and follows logically exactly as the prototype.

harry02031 (approved):
Code looks good and functionality works as expected according to figma prototype. All the state views follow logically from each other

templates/vue/src/components/tyde/activities/CircleOfSupport/CommunityView/index.vue
```diff
@@ -15,36 +15,99 @@
...
+    <tooltip v-if="isState('Connections.AddAnotherTooltip')" position="left">
+      <h3>
+        Click here to add some
+        <br />
+        of your connections!
+      </h3>
+      <b-button pill variant="secondary" @click="send(OnboardingEvents.Add)">
+        Got it &#8594;
+      </b-button>
+    </tooltip>
```
wynnset, line 75: Since these are all related to the onboarding process, I suggest we move all of this and whatever onboarding logic or code we can to a separate component to keep this file clean and minimal.

templates/vue/src/components/tyde/activities/CircleOfSupport/CommunityView/onboarding/AddConfirmation.vue
```diff
@@ -0,0 +1,54 @@
...
+  components: {
+    CosModal,
+  },
+}
+</script>
+
+<style scoped>
+.row {
+  padding: 5px 0;
+}
```
wynnset, line 53: Just FYI, Bootstrap has really useful spacing classes that can be used to avoid doing custom CSS like this. For example, you can just give the class `py-2` to have a similar padding added. See https://getbootstrap.com/docs/4.0/utilities/spacing/

templates/vue/src/components/tyde/activities/CircleOfSupport/CommunityView/onboarding/ObFinishView.vue
```diff
@@ -0,0 +1,98 @@
...
+        >
+          <p class="ob-connection">{{ connection.name }}</p>
+          <h1>{{ connection.avatar }}</h1>
+        </b-col>
+      </b-row>
+      <b-row align-h="center">
+        <h3>
+          Great job! Here are some of
+          <br />
+          your connections.
```
wynnset, line 19: `<br>` is generally not a very responsive way of coding since smaller screen sizes will have their own line breaks in addition as well. I recommend adding a `max-width` to the element here to achieve the results you want instead.

templates/vue/src/components/tyde/activities/CircleOfSupport/CommunityView/onboarding/ObFinishView.vue
```diff
@@ -0,0 +1,98 @@
...
+      </b-row>
+      <b-row align-h="center">
+        <h5 class="ob-secondary">
+          You'll be able to edit and add more at any time.
+        </h5>
+      </b-row>
+      <b-row align-h="center">
+        <b-col v-show="visibleConnections.length >= 4" class="ob-connection-box">
+          <p class="ob-connection">
+            {{ (visibleConnections[3] && visibleConnections[3].name) || "" }}
```
wynnset, line 30: This is a very "reacty" way of doing things :) It'd be better to use v-if or v-show on the element instead.

templates/vue/src/components/tyde/activities/CircleOfSupport/CommunityView/onboarding/ObFinishView.vue
```diff
@@ -0,0 +1,98 @@
...
+          </p>
+          <h1>
```
wynnset, line 32: Generally, it's not a good idea to use heading elements if they are not really headings. If you are trying to achieve the look and feel of a heading, it's a better idea to use CSS for that.

templates/vue/src/components/tyde/activities/CircleOfSupport/CommunityView/onboarding/ObFinishView.vue
```diff
@@ -0,0 +1,98 @@
+<template>
+  <cos-modal show>
+    <b-container class="modal-cont">
```
wynnset, line 3: Please avoid shortening names where possible. `modal-container` would be a perfectly fine and more clear class name here.

templates/vue/src/components/tyde/activities/CircleOfSupport/CommunityView/onboarding/Tooltip.vue
```diff
@@ -0,0 +1,102 @@
...
+      tooltip.style.bottom = "95px"
+      tooltip.classList.add("tooltip-container-left")
+    }
+
+    // const { x, y } = Helpers.positionTooltip(target, tooltip, container, {
+    //   dryRun: true,
+    // })
+    // tooltip.style.transform = `translate(${x}px, ${y -
+    //   tooltip.clientHeight -
+    //   target.clientHeight}px)`
```
wynnset, line 53: Not sure why this is commented out? If you want to leave commented out code, please also include a comment to explain why it's currently commented out.

templates/vue/src/components/tyde/activities/CircleOfSupport/CommunityView/onboarding/Tooltip.vue
```diff
@@ -0,0 +1,102 @@
...
+    // Temporary fix of positioning
+    if (this.position === "right") {
+      tooltip.style.right = "22px"
+      tooltip.style.bottom = "95px"
+      tooltip.classList.add("tooltip-container-right")
+    } else if (this.position === "left") {
+      tooltip.style.left = "22px"
+      tooltip.style.bottom = "95px"
+      tooltip.classList.add("tooltip-container-left")
+    }
```
wynnset, line 46: Is it not possible to do this with CSS instead?

templates/vue/src/components/tyde/activities/CircleOfSupport/CommunityView/onboarding/WelcomeCommunities.vue
```diff
@@ -0,0 +1,122 @@
...
+
+      <b-row align-h="center">
+        <h4 class="ob-secondary">
+          Let's start with you — what communities are you a part of?
+        </h4>
+      </b-row>
+
+      <b-row align-h="center">
+        <h5 class="ob-secondary">Here are a few ideas to get you started.</h5>
+      </b-row>
```
wynnset, line 18: Again, it's better to avoid using heading elements just for styling purposes. Use (S)CSS instead

templates/vue/src/utils/Helpers.js
```diff
@@ -432,7 +432,7 @@ export default class Helpers {
    * 3. The tooltip is clipped on BOTH the bottom and the right (when the
    *    connection is on the bottom-right of the CoS)
    */
-  static positionTooltip(target, tooltip, container) {
+  static positionTooltip(target, tooltip, container, { dryRun = false } = {}) {
```
wynnset, line 435: What is `dryRun`? Could we include something in the comment block above about this?

wynnset (changes requested):
This is looking really good! Love the structure behind it and very neat organization of everything in general. Please see my review comments for some feedback. Some of my feedback would apply to more than the spots that I left them - I just didn't want to repeat my feedback. I also noticed a few bugs and things that could use UI improvements:
- [x] If I follow through with the instructed action, it should automatically do what the "Got it" button does as well (i.e. hide the instructions there). Otherwise, we should not allow the user to click the (+) button in this case. This also applies to the connection instruction and anywhere else that this same concept would apply. 
image
- [x] Not sure if this is intended, but currently when I partially go through the onboarding, if I refresh, I have to start the onboarding all over again and lose my progress or anything I added in. This should get fixed for the final version of this PR and I cannot test this currently since my data is not getting saved.
- [x] Could we make disabled buttons a bit more pale? It still looks quite clickable here: 
image
- [x] In the same screenshot as above, if I click on the text "FAMILY" or the icon below it, the checkbox should also check. The checkbox itself is quite small. You can fix this by adding a label element as the wrapper to the input that includes the input and the label ("FAMILY" and the icon). 
- [x] When submitting, can we add a loading icon? Currently it takes a couple of seconds for me and I'm not sure if I submitted or not. We should also disable the submit button to prevent double-clicking.

harry02031:
@EldadZZipori 

https://user-images.githubusercontent.com/30706624/124331971-37c7cf00-db45-11eb-8b16-052ac5b81110.mov

Screen Shot 2021-07-02 at 2 52 18 PM

I found out that after I choose to add another community and added it, it does not take me to the "Add More confirmation stage". Which makes it unable to reach the connections stage.

harry02031:
Screen Shot 2021-07-02 at 3 05 57 PM

In the connections stage, after I added a connection, it takes me straight to the community home stage. It skips the stage of the stages(and the transitions) outlined above and does not show the message "Great Job! Here are some of your connections", is this expected behaviour?

EldadZZipori:
Fixed both problems + some linting and conditional component initialization. 
Please node to test this with a new user as the saving of the progress was changed in some point to use and Object and not Array and saving with a user that tested this PR in the past causes some issues.

harry02031:
In the add connection form, there is an option to add more communities but this does not work right now. Is this functionality supposed to be supported here? If not I suggest, just removing this add community option
https://user-images.githubusercontent.com/30706624/124858577-2fb3c900-df63-11eb-924c-b268f89c97e7.mov

harry02031:
Screen Shot 2021-07-07 at 8 40 32 PM
Screen Shot 2021-07-07 at 8 40 52 PM

Should the button say "continue" instead? Currently it says connections, but on the diagram, it says continue

EldadZZipori:
Yes you are right about both. I deleted the funcctionality on the first one because I did not notice it. I will revert it and fix the texts.

templates/vue/src/components/tyde/activities/CircleOfSupport/CommunityView/onboarding/Tooltip.vue
```diff
@@ -0,0 +1,89 @@
...
+  },
+  data() {
+    return {
+      positioned: false,
+    }
+  },
+  mounted() {
+    // const container = document.getElementById("cos")
+    // const target = container.querySelector(`#${this.refId}`)
+    this.positioned = true
```
harry02031, line 31: Get rid of comments in this file

harry02031:
Very nicely done on this PR, I really like how the community view onboarding looks. All the functionality is correct and the code looks very nice as well.

EldadZZipori:
This PR is superseded by #1072

PR #1016, opened 2021-05-12
1009 - Allow choosing color for node and text on node

harry02031:
## Changes
Added Color Picking functionality to the appearance tab when editing a node. The user is able to change the background color of the node and the content text color of the node using the ColorPicker Component
## Screenshot/Video of Feature
image
image
## Issue Linkage
Closes (issue #1009 )
## PR Dependency
(N/A)
## Automated Testing
* Please enter what automated testing you have added so that testers will know what they should NOT test

Added new unit test to cypress/integration/node-appearance.spec.js


EldadZZipori:
I tested this PR and found something that did not work as expected on my end.

1. Background color and text color disappear when page is refreshed both on Firefox and Chrome.

- This occur without any console error, both on parent and child node, with no dependency on how many nodes there are. 

- If there was an image previous to a background color the node will revert to the picture on refresh even though thumbnail is unchecked

Refresh
Refresh-Pic


2. This one is not a bug, just a suggestion - It is really hard to see/notice the box when it is white, maybe add border when the selected color is really light, or just a border for all colors.

harry02031:
@EldadZZipori 
Screen Shot 2021-05-17 at 12 02 45 PM
added border as such

I didn't notice this issue on my machine. After setting text color and background color, even after refreshing, the node still preserved all its colors. Did you try npm install && npm start just in case?




https://user-images.githubusercontent.com/30706624/118543305-f4d2a900-b708-11eb-8df2-8392549330e0.mov


For the thumbnail issue, it is a separate issue that was found before and is unrelated to this component(another bug ticket).

EldadZZipori:
I rechecked and debugged this and it still seems that the color is not being saved. I double checked with @wynnset and it didn't save on his end either. I saw that there are no commits for back end php files, maybe they did not upload and that is why the fix doesn't work if we pull this branch,

harry02031:
The php file have been committed properly now and the node should now able to save its color after changing them.

EldadZZipori (approved):
Everything works and is saved as expected. No console errors. Code looks really good. I really like this color picker. All tests pass locally as well.

wynnset (approved):

PR #1025, opened 2021-05-19
1024 - Hyperlinks don't update when editing multiple nodes

EldadZZipori:
## Changes
- Previous code didn't set the URL if it already existed.
- Setting the URL from the prompt to the current URL
## Screenshot

To see this problem at first you need to -
Hyperlink-Update

1. Open a text form in a node (description or anything)
2. Set a URL for any text
3. Try to update the URL (Note that the URL prompt is different with "update" instead of "add")
4. Note the URL disappears completely and needs to be set again

## Issue Linkage
Closes (issue #1024 )


Frozemint:
Yep, can confirm I can no longer replicate this issue in your new branch. All tests passes locally, and the code makes sense.

Frozemint (approved):

wynnset (approved):

PR #1027, opened 2021-05-19
1022 - Bug fix: Thumbnail didn't remove upon refresh even after removing, deselecting add a thumbnail

### LINKED ISSUE #1022: Thumbnail didn't remove upon refresh even after removing, deselecting add a thumbnail

odotadot:
## Changes
- Fixed issue with thumbnail image persisting even when user removes it after refreshing page
- Fixed issue with locked thumbnail image persisting even when user removes it after refreshing page
## Issue Linkage
Closes #1022
## PR Dependency
N/A
## Automated Testing
- Create new node with thumbnail. Upload picture. Refresh the page and the image should still persist. Edit node again, remove the thumbnail. The image should be gone. Refresh the page, the image should be gone. (same for locked thumbnails)
- Create new node with thumbnail. Upload picture. Uncheck the "add a thumbnail checkbox". Publish node. The thumbnail should be gone. Refresh page, and the thumbnail should still be gone.


EldadZZipori:
The problem of removing an Image is fixed, but I noticed that when unchecking the "Add a thumbnail" checkbox the same problem occurs and the picture is displayed after a refresh. Not sure this is the same as the reported bug. @odotadot

odotadot:
Great catch! I will look into it.

EldadZZipori (approved):
All issues fixed, all tests pass on local machine as well, and code looks good. If this works with php's empty function that might be better instead of checking for an empty string in case we will call this functionality somehow in the future.

wynnset (approved):

PR #1030, opened 2021-05-25
1029 - TYDE - Allow creating avatar and use it in app

### LINKED ISSUE #1029: TYDE - Allow creating avatar and use it in app

wynnset:
## Changes
* Show an interface to the user asking them to customize their avatar the first time they use a tapestry
* Allow users to go in and edit their avatar at any point in the settings
* Update TYDE to use avatars where needed and create a simple way of viewing avatars where needed in the code
* Avatars are shared across all tapestries on a given site
## Screenshot
https://user-images.githubusercontent.com/34692466/123856151-9850d500-d8d5-11eb-8076-24c8d4a6b8bf.mp4
## Issue Linkage
Closes #1029
## PR Dependency
N/A
## Automated Testing
- Changing .env variable APP_VUE_AVATARS=TRUE will show the avatar customizer prompt upon entering tapestry if no avatar has been saved for user yet
- Changing .env variable APP_VUE_AVATARS=FALSE will disable ALL avatar-related components (popup, user-settings button)
- If not logged in, no Avatar related components will show up
- Can go to the User Settings (new toolbar button) to change Avatar and save it
- Saved Avatar look should reflect in the Avatar Customizer every time you open it
- Avatar is saved for user across Tapestries


odotadot:
Notes: 
- Have not yet updated TYDE to use these Avatars yet.
- If modifying the .env file in tapestry-wp folder doesn't work, try modifying the .env file located in templates/vue

EldadZZipori:
1. For me both on Chrome and Firefox, the avatar appears to be small and sticks to the left. (left is Firefox, right is Chrome)
Screenshot from 2021-06-17 09-15-42

2. I am not getting prompt to create my avatar. I tried the following - 
- Create new user and open old tapestry
- Create new tapestry and open it with new user
- Access old tapestry with admin and subscriber
- Setting the .env to true and accessing a tapestry with a new user

I am a bit confused about the APP_VUE_AVATAR_POPUP settings. Is it just for automated testing or is it also for disabling the avatar? And if it is for disabling the avatar does it disable the whole feature or just the pop up for the avatar?

3. I am unable to access the avatars with non-admin user as they don't have access to the settings tool tip. 
Screenshot from 2021-06-17 09-22-03

Other than that saving across the site seems to work. Changing the avatars settings works smoothly. All the tests pass locally with both settings.

templates/vue/src/components/TapestryApp.vue
```diff
@@ -32,13 +47,26 @@ export default {
...
     },
+    hasAvatar() {
+      return !(!this.avatar || Object.keys(this.avatar).length === 0)
+    },
+    avatarPopup() {
+      if (
+        process.env.VUE_APP_AVATAR_POPUP === "true" ||
+        process.env.VUE_APP_AVATAR_POPUP === "TRUE" ||
+        process.env.VUE_APP_AVATAR_POPUP === 1
+      ) {
```
EldadZZipori, line 65: I believe there is no need for three "or" as JS has a spacial definition for "truly" expressions - https://developer.mozilla.org/en-US/docs/Glossary/Truthy

templates/vue/src/App.vue
```diff
@@ -70,7 +70,9 @@ export default {
         })
       })
     }
-
+    Promise.resolve(client.getAvatar()).then(savedAvatar => {
+      this.addAvatar(savedAvatar.data)
+    })
```
EldadZZipori, line 75: Would this still occurs with no user is logged in?

odotadot: That code still runs even if the user is not logged in. I think we should prevent it from running if that is the case, but I think Aidin mentioned something about a mechanism forcing users to log in on the TYDE app.

templates/vue/src/components/TapestryApp.vue
odotadot: Good catch, thanks!

odotadot:
> 1. For me both on Chrome and Firefox, the avatar appears to be small and sticks to the left. (left is Firefox, right is Chrome)
>    Screenshot from 2021-06-17 09-15-42

It doesn't happen on my end, but I pushed a commit with some modifications to the CSS. Let me know if it still looks like that.

> 2. I am not getting prompt to create my avatar. 

I am unable to re-create this issue as well. Did you modify the .env file in your templates/vue folder? Then re-run `npm start` for the changes to occur. The pop-up will only show up if you don't have a saved avatar (so it should **always** pop-up if you are not logged-in, and the for the **first time** if you open any tapestry before saving an avatar.)

> I am a bit confused about the APP_VUE_AVATAR_POPUP settings. Is it just for automated testing or is it also for disabling the avatar? And if it is for disabling the avatar does it disable the whole feature or just the pop up for the avatar?

We can set APP_VUE_AVATAR_POPUP=false to let the tests pass, and it is used to enable/disable whether the Avatar pop-up prompt occurs if a user without an avatar first goes into a tapestry. It shouldn't affect the avatar form in the Settings modal though.

> 3. I am unable to access the avatars with non-admin user as they don't have access to the settings tool tip.
>    Screenshot from 2021-06-17 09-22-03

You're right...that might be an issue. As of currently non-admin people can save an avatar from the first pop-up prompt, but can't modify it.

@wynnset What should I do about (3)? And let me know if I'm wrong about any the assumptions above.

wynnset:
> What should I do about (3)? And should I also prevent the pop-up prompt if the user is not logged in? Let me know if I'm wrong about any the assumptions above.

@odotadot let's create a new item in the toolbar for user settings (which currently would only be for avatar modifications). This should show up only if the user is logged in and the flag for using avatars is on in the .env file.

Let's prevent the pop-up if the user is not logged in. Thanks!

odotadot:
> @odotadot let's create a new item in the toolbar for user settings (which currently would only be for avatar modifications). This should show up only if the user is logged in and the flag for using avatars is on in the .env file.

@wynnset All avatar related components now only show if the user is logged in and the avatar .env variable is set to TRUE. However, how can a non-admin user (subscriber) edit their avatar? They will see the avatar customizer pop-up when they enter the tapestry, but they cannot access the user-settings in the toolbar.

wynnset:
> they cannot access the user-settings in the toolbar.

Why not? All authenticated users should be able to access this new menu item. To reiterate, this is a new item and not the same as the tapestry settings menu item.

odotadot:
> Why not? All authenticated users should be able to access this new menu item. To reiterate, this is a new item and not the same as the tapestry settings menu item.

You're totally right. Didn't realize I could just use some variables to do this.

EldadZZipori:
Hi, can we add more documentation to this PR. At the description it says the env variable is APP_VUE_AVATARS, and then in one of the comments APP_VUE_AVATAR_POPUP. 

Also in which of the .env file should this variable be, we have two in the project.

UI fixes: 
1. Capitalize first letter in Background option inside the avatar popup
2. Add spaces between words in Face option - Eyebrow Type, Hair - Hair Type, Hair - Hair Colour, Hair - Hat Colour, Hair - Facial Hair Type, Hair - Facial Hair Colour, Clothing - Clothing Type, and Clothing - Clothing Colour.
3. My avatar is still really small. 
image
4. The settings button is a bit off center
image

templates/vue/src/App.vue
```diff
@@ -69,8 +69,10 @@ export default {
           that.loggedIn = data["wp-auth-check"]
         })
       })
+      Promise.resolve(client.getAvatar()).then(savedAvatar => {
+        this.addAvatar(savedAvatar.data)
+      })
```
EldadZZipori, line 74: I"m not sure if this check should be under app.vue, maybe under tapestryApp.vue?

templates/vue/src/components/modals/UserSettingsModal/AvatarForm.vue
```diff
@@ -0,0 +1,453 @@
...
+        { value: "Gray02", text: "Gray02" },
+        { value: "Heather", text: "Heather" },
+        { value: "PastelBlue", text: "PastelBlue" },
+        { value: "PastelGreen", text: "PastelGreen" },
+        { value: "PastelOrange", text: "PastelOrange" },
+        { value: "PastelRed", text: "PastelRed" },
+        { value: "PastelYellow", text: "PastelYellow" },
+        { value: "Pink", text: "Pink" },
+        { value: "Red", text: "Red" },
+        { value: "White", text: "White" },
```
EldadZZipori, line 385: Since there is so much data here maybe putting it into a JS file that exports this static data will make it look cleaner.

templates/vue/src/components/modals/UserSettingsModal/AvatarForm.vue
```diff
@@ -0,0 +1,453 @@
...
+      newAvatar.eyebrowType = this.eyebrowType
+      newAvatar.eyeType = this.eyeType
+      newAvatar.facialHairColor = this.facialHairColor
+      newAvatar.facialHairType = this.facialHairType
+      newAvatar.graphicType = this.graphicType
+      newAvatar.hairColor = this.hairColor
+      newAvatar.mouthType = this.mouthType
+      newAvatar.skinColor = this.skinColor
+      newAvatar.topType = this.topType
+      newAvatar.topColor = this.topColor
```
EldadZZipori, line 431: Allocating all these props inside the {} would make the code look cleaner as you will not have to access newAvatar all the time.

templates/vue/src/components/modals/UserSettingsModal/AvatarForm.vue
```diff
@@ -0,0 +1,453 @@
...
+      this.eyeType = this.avatar.eyeType
+      this.facialHairColor = this.avatar.facialHairColor
+      this.facialHairType = this.avatar.facialHairType
+      this.graphicType = this.avatar.graphicType
+      this.hairColor = this.avatar.hairColor
+      this.mouthType = this.avatar.mouthType
+      this.skinColor = this.avatar.skinColor
+      this.topType = this.avatar.topType
+      this.topColor = this.avatar.topColor
+    },
```
EldadZZipori, line 414: I think making these computed properties will make everything cleaner and that way you won't need to set them and then use a function to check them.
something like 
```
isCircle(){
 return this?.avatar.isCircle ? this.avatar.isCircle : true
}
```

EldadZZipori (commented):
@wynnset I am not 100% sure about my comments, if you can give your input on them as well.

classes/class.tapestry-user-progress.php
```diff
@@ -120,6 +120,50 @@ public function getH5PSettings()
...
+        }
+        $search_criteria['field_filters'][] = [
+            'key' => 'created_by',
+            'value' => $userId,
+        ];
+        $entries = GFAPI::get_entries($formId, $search_criteria);
+
+        return $this->_formatEntries($entries);
+    }
+
```
wynnset, line 166: This should not be here.

endpoints.php
```diff
@@ -206,6 +206,27 @@
...
+            'callback' => 'updateUserAvatar',
+        ],
+    ],
+    'GET_USER_AUDIO' => (object) [
+        'ROUTE' => '/tapestries/(?P<tapestryPostId>[\d]+)/nodes/(?P<nodeMetaId>[\d]+)/audio/(?P<questionId>[\d]+)',
+        'ARGUMENTS' => [
+            'methods' => $REST_API_GET_METHOD,
+            'callback' => 'getUserAudio',
+        ],
+    ],
```
wynnset, line 229: This should not be here

classes/class.tapestry-user-progress.php
```diff
@@ -216,6 +260,18 @@ private function _getUserH5PSettings()
...
+    {
+        update_user_meta($this->_userId, 'avatar_'.$this->_userId, $avatarData);
+    }
+
+    private function _getAvatar()
+    {
+        $storedAvatar = get_user_meta($this->_userId, 'avatar_'.$this->_userId, true);
+
+        return $storedAvatar ? json_decode($storedAvatar) : (object) [];
+    }
```
wynnset, line 273: Since we are already saving this for this user, it's not needed to have their user ID in the key for this user meta. I also suggest adding something that makes it clear this is a tapestry user meta, e.g. `tapestry_avatar`

interfaces/interface.tapestry-user-progress.php
```diff
@@ -56,6 +56,34 @@ public function updateH5PSettings($h5pSettingsData);
...
+    /**
+     * Get all gravity form entries submitted by this user.
+     * If $formId is passed, returns entries for only that form.
+     *
+     * @param int $formId
+     *
+     * @return string user entries in json format
+     */
+    public function getUserEntries($formId);
+
```
wynnset, line 86: This shouldn't be here.

templates/vue/src/App.vue
wynnset: Why can't we just get this along with all the other user progress stuff so we don't have to do a separate request just for this. Is there a reason for this?

templates/vue/src/assets/icons/user.svg
```diff
@@ -0,0 +1 @@
+<svg aria-hidden="true" focusable="false" data-prefix="fas" data-icon="user" class="svg-inline--fa fa-user fa-w-14" role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 448 512"><path fill="currentColor" d="M224 256c70.7 0 128-57.3 128-128S294.7 0 224 0 96 57.3 96 128s57.3 128 128 128zm89.6 32h-16.7c-22.2 10.2-46.9 16-72.9 16s-50.6-5.8-72.9-16h-16.7C60.2 288 0 348.2 0 422.4V464c0 26.5 21.5 48 48 48h352c26.5 0 48-21.5 48-48v-41.6c0-74.2-60.2-134.4-134.4-134.4z"></path></svg>
```
wynnset, line 1: Why not just use font awesome? Prefer not to increase the bundle size where possible.

templates/vue/src/components/TapestryApp.vue
```diff
@@ -1,6 +1,19 @@
...
+      </b-container>
+      <template slot="modal-footer">
+        <b-button size="sm" variant="secondary" @click="cancel">
+          Cancel
+        </b-button>
+        <b-button size="sm" variant="primary" @click="saveAvatar">
+          Save Avatar
+        </b-button>
+      </template>
+    </b-modal>
```
wynnset, line 16: Is this needed here? I thought we have the modal inside of UserSettingModal which is inside the Toolbar component already.

templates/vue/src/components/Toolbar/index.vue
```diff
@@ -1,17 +1,23 @@
...
-      <settings-modal-button
-        v-if="canEdit"
-        :max-depth="maxDepth"
-      ></settings-modal-button>
-      <tapestry-depth-slider
-        v-show="!showMap && hasDepth"
-        @change="updateViewBox"
-        @change:max-depth="maxDepth = $event"
-      ></tapestry-depth-slider>
+    <div v-show="isLoggedIn" class="slider-wrapper">
```
wynnset, line 4: This is not sufficient because it will show this even if all elements inside are not showing. The condition here should ensure that this toolbar is only displayed if at least one of the elements inside will display.

templates/vue/src/components/Toolbar/index.vue
```diff
@@ -1,17 +1,23 @@
...
+      <user-settings-button
+        v-if="avatarsEnabled"
+        data-qa="user-settings-button"
+      ></user-settings-button>
+      <div v-show="canEdit || (!showMap && hasDepth)" class="canEdit">
```
wynnset, line 9: Classes must be in kebab-case

templates/vue/src/components/modals/UserSettingsModal/AvatarForm.vue
wynnset: Can't we just merge in the whole avatar object with `this` using `...` notation so that we don't have to repeat this line by line?

templates/vue/src/components/modals/UserSettingsModal/AvatarForm.vue
wynnset: Again, I suggest saving the whole thing rather than line by line. Maybe you can just create a `preferences` prop that saves all of these into an object so you can just pass that in rather than set it line by line.

templates/vue/src/store/actions.js
```diff
@@ -303,6 +303,17 @@ export async function updateUserFavourites({ commit, dispatch }, favourites) {
   }
 }
 
+// avatars
+
+export async function updateAvatar({ commit, dispatch }, newAvatar) {
```
wynnset, line 308: Maybe it's best this is called `updateUserSettings` and avatar is just a property of that. This will make it flexible enough to add future user settings easily using the same structure.

wynnset (commented):
This is looking good, but still needs some work as per @EldadZZipori and the comments I've made now. In addition, the README file should explain how to add the .env variable for enabling avatars.

templates/vue/src/App.vue
odotadot: Leaving as is as per the Zoom chat.

templates/vue/src/store/actions.js
odotadot: The retrieving of the avatar has remained the same, but the saving of the avatar has been modified to use the userSettings object which has the avatar property in it.

odotadot:
Addressed all the comments left since the last review. However, I am still unable to re-create the issue with the Avatar size being small inside the user settings modal.

EldadZZipori (approved):
Everything looks good.

- There is one change I believe can by done to trim the code down and should work.
- We need to resolve all comments on this PR (some are still open)
- We should open a new task to utilize the avatars on the tyde2 branch wherever relevant. 

Merge conflicts are resolved so I believe that after these, this PR should be ready to merge.

templates/vue/src/components/modals/UserSettingsModal/AvatarForm.vue
```diff
@@ -0,0 +1,249 @@
...
+      eyebrowTypeOptions: avatarOptions.eyebrowTypeOptions,
+      eyeTypeOptions: avatarOptions.eyeTypeOptions,
+      facialHairColorOptions: avatarOptions.facialHairColorOptions,
+      facialHairTypeOptions: avatarOptions.facialHairTypeOptions,
+      graphicTypeOptions: avatarOptions.graphicTypeOptions,
+      hairColorOptions: avatarOptions.hairColorOptions,
+      mouthTypeOptions: avatarOptions.mouthTypeOptions,
+      skinColorOptions: avatarOptions.skinColorOptions,
+      topTypeOptions: avatarOptions.topTypeOptions,
+      topColorOptions: avatarOptions.topColorOptions,
```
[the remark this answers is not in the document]
odotadot: great catch

wynnset:
@odotadot there are a lot of unrelated changes in this PR, which I think is a mistake... Please fix and assign back to me.

odotadot:
This PR has been superseded by the other avataaars branch #1082.

Issue #991, opened 2021-04-20
Create "Question" node type with audio and text answer types


shirleywbi:

wynnset:
Replaced by #1043

PR #1034, opened 2021-05-28
1033 - Add list answer type for Question node type

### LINKED ISSUE #1033: Add list answer type for Question node type

odotadot:
## Changes
- Adding a list answer type for Question Node
## Issue Linkage
Closes #1033
## PR Dependency
Depends on: #991 
## Automated Testing
- Can create a Question Node with a List-type answer
- Can add and delete answers in the List form
- Can add answers by pressing Enter key
- List Form state will persist between opening and closing the lightbox


EldadZZipori (approved):
New feature works great both in normal node and multi-content node. Code for the list type and the cypress test looks really good. All tests pass locally. One time Mic test failed on my machine because cypress did not find the submitting  text, but this has happen in PR that are not branched from master directly, is not related to this code or PR, and only happens on occasion.

PR #1036, opened 2021-05-31
1026 - Add new content type "Answer"

### LINKED ISSUE #1026: Add new content type "Answer"

odotadot:
## Changes
- Adds the AnswerMedia and AnswerForm Vue components, creating the Answer Node type
- Adds Answer field to the Tapestry Node
- Allows users to select a question in the tapestry and displays the answer to it
- Allows users to change the follow-up text in the Answer node. If blank, will use the default follow-up text from the question node
- Added Cypress test to ensure basic functionality
## Issue Linkage
Closes #1026
## PR Dependency
Depends on #991 
## Automated Testing
- Can create a Question Node with one question and make an Answer Node displaying the answer to that question
- Can multiple Question Nodes with multiple questions and edit the Answer node form to display the answer to the selected question
- Will display a message stating that the question is not completed if it has not been answered yet
- The answer form will retain the previous entered information upon edit

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ cypress.json
 
 data/
 bin/config.sh
+
+vendor/
--- a/classes/class.tapestry-node.php
+++ b/classes/class.tapestry-node.php
@@ -38,6 +38,7 @@ class TapestryNode implements ITapestryNode
     private $hideMedia;
     private $skippable;
     private $quiz;
+    private $answer;
     private $fullscreen;
     private $childOrdering;
     private $fitWindow;
@@ -85,6 +86,7 @@ class TapestryNode implements ITapestryNode
         $this->hideMedia = false;
         $this->skippable = true;
         $this->quiz = [];
+        $this->answer = (object) [];
         $this->fullscreen = false;
         $this->childOrdering = [];
         $this->fitWindow = true;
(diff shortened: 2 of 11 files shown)


odotadot:
Note: This is the correct branch "1026-add-new-content-type-answer-2" rather than "1026-add-new-content-type-answer" for this ticket. This branch implements the Answer Node for the Question node while the other branch mistakenly implements the Answer node for the Activity node.

EldadZZipori:
1. If we set a node with both audio and text answer and answer with both one after the other, we are able to see both of them in the answer node. Not the last answer. Is this the expected behavior?
ezgif com-gif-maker(2)

2. This is not really argent - when there is no answer the text states "Previously, you said:"/follow up text and "you have not completed this question yet" which feels a bit weird, maybe we can only display "you have not completed...." if there is no answer.
Screenshot from 2021-06-03 08-58-48

Other then these two, everything works. With multi-content, and both available question types.

odotadot:
> 1. If we set a node with both audio and text answer and answer with both one after the other, we are able to see both of them in the answer node. Not the last answer. Is this the expected behavior?
@wynnset Should I change it to only show the most recent answer?

wynnset:
Showing both is fine. I think maybe we would want to present them in a way that you only see one at a time (maybe with some minimal tabs at the top or on the side), defaulting to the latest answer.

EldadZZipori (approved):
Everything works great. I really liked the small tabs for the questions with more then one answer type. 
Code for tests and feature look good.

Some test related to Everything works great. I really liked the small tabs for the questions with more then one answer type. 
Code for tests and feature look good.

I Sometimes get some tests failing locally, related again to the "submitting..." of audio question, but this is not related to this PR. - This does not always occurs.

Just a note that the problem from @991 that one users sees the other user's answer persists. But is not really related to this PR.

PR #994, opened 2021-04-22
991- Create "Question" node type with audio and text answer types

shirleywbi:
## Changes
* Add question node content type (similar to "Activity" except it handles a single question)
    * *QuestionForm*         
        * Add ability to toggle text and audio response, show a previous answer before an activity, and customize confirmation page
            * Previous answers are cross-compatible between activity and question nodes.
            * Options are stored in `node.typeData` as 
image
        * For a text response, allow selecting multi-line vs. single line. Single line allows setting an optional placeholder.
        * Add error handling for empty question text and unselected response type
    * *Lightbox*
        * Shows activity screen without navigation
        * Update activity screen to use textarea and input for question nodes, based on form selection. Responses are saved under `quiz[0].entries` as `textId: { [formId]: "response" }` to make it similar to how audio and gravity forms are structured.
* If the current node is not an "activity", hide "Activity" from Content Type dropdown
## Screenshot
ezgif com-gif-maker (12)
## Issue Linkage
Closes #991 
## PR Dependency
Depends on: N/A
## Automated Testing
* Able to complete with an text-based answer
* Able to complete with an audio-based answer

PR #1039, opened 2021-06-02
1032 - Add drag and drop answer type for Question node type

harry02031:
## Changes
Author is now able to select drag and drop answer type for question node. Author can add "from buckets" and "to buckets" and also can add items in "from buckets" in authoring. Each bucket item is customizable to have a background color, in addition, if the author chooses to add background image, they can also add a text label for the item. The user can drag and drop items between "from buckets" and "to buckets", between different "to buckets", between different "from buckets", and from a "to bucket" back to a "from bucket"

Author input for this feature is saved under node.typeData.options.dragDrop
Screen Shot 2021-06-02 at 9 23 21 PM

User answer for this feature is saved under quiz[0].entries as dragdropId: { [dragdropId]: "response" } to make it similar to how text and audio responses are stored in the depended PR 991
Screen Shot 2021-06-02 at 9 23 50 PM

## Screenshot/ Video of feature (

https://user-images.githubusercontent.com/30706624/120587725-c0d6d380-c3ea-11eb-89e6-2eb2393de5ae.mov

## Issue Linkage
Closes (issue #1032 )
## PR Dependency
Depends on: (PR #994 )
## Automated Testing (coming soon)
* Should be able to complete drag and drop answer with items with no background image
* Should be able to move items to different buckets in all directions in drag and drop answer mode
* Should be able to complete drag and drop answer with items with background image and text


EldadZZipori:
1. When submitting the page again (Hard Refresh or Logout-Login), the previous answer is not retrieved .
ezgif com-gif-maker

2. The problem that a previously entered answer by one user is visible by another user does not persist, but this might be related to (1)

3. You can check the Background image, and then text, without providing an image/text and submit without any alert or warning. They will stay checked and empty when re-editing the node.

These are more for discussion and general thought for User Experience - 

- Why not allow text without when there is no image added to the bubble, it feels a bit empty just with colors
- Maybe remove the "Remove Bucket" when there is only one, it is a bit confusion even when it is disabled

harry02031:
@EldadZZipori 
Hey I made the circle smaller, and the remove button will no longer show up if there is only one item. I also made sure that the background image checkbox and the text checkbox will not stay checked if the user does not upload any image or input any text into the field. Regards to your comment "Why not allow text without when there is no image added to the bubble, it feels a bit empty just with colors", it states on Asana that the following "For each "from" bucket, the author should be able to add items in it. Each item should be configurable to have: 
A background color (default grey) 
A background image (optional)
Text (optional if a background image was added)"
which means that the text field only appears when the user selects background image.

Regards to your number 1 bug, I couldn't reproduce that on my end, when I submitted the answer for the drag and drop and refresh it again, the answer will always be the same as the previous answer. 


https://user-images.githubusercontent.com/30706624/120959840-841f1b00-c70f-11eb-906b-b9c9152f7a01.mov

EldadZZipori:
Everything seems to work fine. All local test pass. No console errors. 
There is just one thing which I think is related to the base branch @991 and not this one.
When saving an answer on one user it is also saved for another, it seems like answers are being saved only for the admin maybe.

https://user-images.githubusercontent.com/50620330/121058094-d6deee00-c774-11eb-8dfa-d14477c5a737.mp4

harry02031:
Hey @EldadZZipori, that issue is not present on my end. Switching users/refreshing does not remove the submitted answer on the tapestry, and submitted answer are saved for all users. 


https://user-images.githubusercontent.com/30706624/121064847-ee21d980-c77c-11eb-927f-665bd75fee22.mov

EldadZZipori:
I think that I see the same thing happening on your end. The second user sees the answer the admin submitted instead of seeing his own answer. It is similar to what is happening on the base branch that all users see the answer admin submitted.

EldadZZipori:
This was superseded by #1058

EldadZZipori:
This was superseded by #1058

PR #1040, opened 2021-06-02
1037 - Allow navigating tapestry by keyboard

### LINKED ISSUE #1037: Allow navigating tapestry by keyboard

Frozemint:
## Changes
This PR removes many instances of `outline: none` CSS styles such that outlines for keyboard navigation users shows up for better accessibility. Typically [outlines](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) are [removed for aesthetic purposes](https://css-tricks.com/almanac/selectors/f/focus-visible/), but this is [terrible for accessibility](https://www.a11yproject.com/posts/2013-01-25-never-remove-css-outlines/). 

In addition, this PR also adds high visibility CSS `outlines` for keyboard navigation users to make the basic keyboard navigation support explicit by drawing a [VoiceOver-like](https://en.wikipedia.org/wiki/VoiceOver) cursor outline. This is done by using the `:focus-visible` pseudo-class. See example screenshot below.

## Screenshot
image
image

## Issue Linkage
Closes #1037 
## Depends on: (PR # or N/A)
N/A
## Automated Testing (:information_source: : Please read notes)
Due to the limitations of [Cypress](https://docs.cypress.io/api/commands/type), which does not allow us to realistically test browser keyboard focus and navigation functionality, it's best for testers to test this PR using their own browser environment. Ensure your browser do not have anything that overrides the `:focus-visible` CSS pseudo-class, and try to navigate through all the tapestry on-screen elements by using tab. 

You can tab forward by pressing `TAB`, go back to the previous element by pressing `SHIFT+TAB`. To activate and "click" an element, press `ENTER`. Clicking an element should yield identical behaviour as a mouse click. For modifying sliders, use your `LEFT` and `RIGHT` arrow keys.

## Appendix: Using Tapestry with a Screen-reader (VoiceOver, MacOS version 11.4)

Currently tapestry support for screen readers are limited, with the main tapestry screen of nodes being recognised as an image and cannot be interacted with.

https://user-images.githubusercontent.com/11191061/121578928-11988e80-c9e0-11eb-9dcf-78fcc7907b81.mov

### DIFF
--- a/templates/tapestry.css
+++ b/templates/tapestry.css
@@ -111,7 +111,6 @@ li p {
     color: #999;
     font-size: 1.2em;
     transition: all 0.2s ease;
-    outline: none;
 }
 
 .settings-button:hover {
--- a/templates/tapestry.css
+++ b/templates/tapestry.css
@@ -123,7 +123,6 @@ li p {
 
 #tapestry-depth-slider {
     margin-top: 6px;
-    outline: none;
     background: #d3d3d3;
     height: 10px;
     opacity: 0.8;
@@ -131,6 +130,7 @@ li p {
     position: relative;
     margin: 0 32px;
 }
+
 
 #tapestry-depth-slider::before,
 #tapestry-depth-slider::after {
--- a/templates/vue/src/App.vue
+++ b/templates/vue/src/App.vue
@@ -132,10 +132,6 @@ html {
       }
     }
 
-    button:focus {
-      outline: none;
-    }
-
     .btn {
       &:disabled,
       &.disabled {
--- a/templates/vue/src/components/Lightbox/TapestryModal/ModalButton.vue
+++ b/templates/vue/src/components/Lightbox/TapestryModal/ModalButton.vue
@@ -2,6 +2,7 @@
   <button
     class="modal-button"
     :style="buttonStyles"
+    tabindex="0"
     @click="$emit('clicked')"
     @mouseover="hovering = true"
     @mouseout="hovering = false"
--- a/templates/vue/src/components/Lightbox/media/TapestryMedia.vue
+++ b/templates/vue/src/components/Lightbox/media/TapestryMedia.vue
@@ -14,17 +14,21 @@
   >
     <text-media
       v-if="node.mediaType === 'text'"
+      ref="text"
       :node="node"
       :context="context"
+      tabindex="0"
       @complete="complete"
       @load="handleLoad"
     />
     <video-media
       v-if="node.mediaFormat === 'mp4'"
+      ref="mp4"
       :autoplay="autoplay"
       :node="node"
       :dimensions="dimensions"
       :context="context"
+      tabindex="0"
       @load="handleLoad"
       @complete="complete"
       @timeupdate="updateProgress"
@@ -32,10 +36,12 @@
     />
     <youtube-media
       v-if="node.mediaFormat === 'youtube'"
+      ref="youtube"
       :autoplay="autoplay"
       :node="node"
       :dimensions="dimensions"
       :context="context"
+      tabindex="0"
       @load="handleLoad"
       @complete="complete"
       @timeupdate="updateProgress"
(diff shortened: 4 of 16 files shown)


EldadZZipori:
Test using Chrome -
Most of the testing was done from the perspective of an end user and not a user that can edit a tapestry as I believe it is more relevant to this task.

- Currently the first time you hit tab does not change the focus to the root node, but rather to the wordpress button on the top left. Is there a way to control the beginning of the navigation stream?
- Once entering the node the first focus is on the content itself, in order to get to the heart button or exit button I had to click shift-tab and not tab. I believe the user should get to them when they finish the node's content.
- When existing a node's lightbox the focus did not return to the node I closed.
- When opening an upload video (both in lightbox and with full screen option)/ H5P video/  node the focus does not transfer to any element inside the lightbox of that node and it takes a lot of time to get to the play, media bar, and other buttons in the video player.
- The navigation is not coherent throughout the process.  When hitting tab sometimes it jumps to a child node, some times to a sibling and some time to a button outside of the tapestry.

Overall the code looks really good and uses smart/relevant code for accessibility (I learned a lot from it). The current functionality is impressive and allows  navigating tapestry with the keyboard, although with some discomfort. Somethings like multi-content and editing a new node work impressively well. Also documentation of this PR made it really simple to follow along. Really great job.

From my perspective to improve the navigation of tapestry to be perfect for accessibility uses, we will have to implement a "time machine" or a state machine similar to the one in on-boarding in order to track events and situations in order to know how the system should behave.

Frozemint:
> Currently the first time you hit tab does not change the focus to the root node 

It is possible to specify the tabing order by using [`tabindex`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex), but that breaks more things than it fixes, as you now need to specify tab order for every element on screen that could be tabbed to. 
Additionally, a fundamental limitation of the "tabbing" keyboard navigation right now is that it applies to the entire HTML page, and tabbing is supposed to work like that. If we restrict tabbing to only work in the Tapestry App it would break and betray the fundamental functionality of the HTML tabbing feature. I agree that the current method of tabbing is basic and limited, but to implement our own keyboard navigation controls would require substantially more work; this is why I am implementing the basic tabbing functionality support first.

> Once entering the node the first focus is on the content itself

This is done to partially support screen readers. Right now screen readers does not recognise that a lightbox popup has opened when we open a node, and I have to manually override the keyboard focus using `.focus()` to the content such that the screen reader would read the popup content correctly. 

> in order to get to the heart button or exit button I had to shift-tab

This is a problem with the ordering of the HTML elements, and would require us to rearrange the HTML code if we want to fix this. I want to note that the tabbing order depends on how the HTML elements are arranged in the code.

> When existing a node's lightbox the focus did not return to the node I closed.

It should be possible to override the keyboard focus back to the now-closed node. I could look into that.

> When opening an upload video (both in lightbox and with full screen option)/ H5P video/ node the focus does not transfer to any element inside the lightbox

I'll look at changing the focus to the controls if that is possible.

> The navigation is not coherent throughout the process. When hitting tab sometimes it jumps to a child node

This is a consequence of how nodes are arranged in HTML, all tabbing does is navigate the HTML sequentially. If we want to fix this, we'll have to change how nodes are ordered in HTML. 

@EldadZZipori

wynnset:
Thanks for the review. @Frozemint :

> Currently the first time you hit tab does not change the focus to the root node

I think this can be fixed by applying `.focus()` on the root node when tapestry is done loading.

With regards to tab ordering, I think we can easily switch that in the HTML (and we should do that as part of this task).

Frozemint:
Thanks Aidin, the `.focus()` feature is now in. It took a while because `.focus()` by default also brings up the keyboard navigation outline, which may be confusing for the regular user. I found that by calling `focus()` and then `blur()` gives the desired effect. Will be working on rearranging the HTML element order next.

wynnset:
Superseded by #1199

Issue #1049, opened 2021-06-15
TYDE2 - Create TYDE shell


odotadot:

wynnset:
Superseded by #1061

PR #1053, opened 2021-06-16
1049 - TYDE2 - Create TYDE shell

odotadot:
## Changes
- Creating TYDE Shell
## Screenshot
TBD
## Issue Linkage
Closes (issue #1049 )
## PR Dependency
TBD)
## Automated Testing
TBD

### DIFF
--- a/classes/class.constants.php
+++ b/classes/class.constants.php
@@ -2,22 +2,22 @@
 
 class ConditionTypes
 {
-    const NODE_COMPLETED = 'node_completed';
-    const DATE_NOT_PASSED = 'date_not_passed';
-    const DATE_PASSED = 'date_passed';
+    public const NODE_COMPLETED = 'node_completed';
+    public const DATE_NOT_PASSED = 'date_not_passed';
+    public const DATE_PASSED = 'date_passed';
 }
 
 class NodeStatus
 {
-    const PUBLISH = 'publish';
-    const DRAFT = 'draft';
-    const SUBMIT = 'submitted';
-    const REJECT = 'rejected';
-    const ACCEPT = 'accepted';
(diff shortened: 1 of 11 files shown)


odotadot:
[](https://user-images.githubusercontent.com/34692466/122481668-f4396680-cf83-11eb-93f6-3eaf4d8e3998.mp4)
I have a very rough implementation of the shell right now. I take the root node (a multicontent node) and pass it into the router to display the first multicontent tab for the TYDE shell. Should I be setting up new routes to show multicontent and nestedMulticontent views from the tyde shell? (If I click on an accordion item in the shell right now, it will take me back to the lightbox view of that item.) In addition, the tyde shell currently dominates the screen upon loading the App. Should the tapestry be visible at any point? I think you mentioned that a user that's logged in should see the Tapestry App.

wynnset:
If a user has access to edit, they should be able to see the tapestry itself and view the tyde shell by clicking on the media button for the root node. We should put a close button or something to take the admin user back to the tapestry view. 

In terms of implementation, I recommend that the route should maintain the same form when the tyde shell is on. Thinking about this some more, I think the tyde shell can actually be an extension of the lightbox to make things easier and more consistent. Essentially, the tyde shell is the lightbox in fullscreen mode with some added tabs to it. That might be the easiest way to implement it. If a user does not have edit capability, we would:
1. Hide the tapestry from them
2. Automatically open the root node in fullscreen lightbox when the page loads
3. Ensure they cannot exit out of the lightbox

Hope that clears some things up. I'm happy to jump on a call to discuss this further!

EldadZZipori:
Moved to 1061 - add tyde mode

PR #1056, opened 2021-06-21
1026 - Add new content type "Answer"

odotadot:
This PR adds an "Advanced" node type for displaying a single answer from an activity. 
The user will be able to view all answers (i.e. audio, text including multi-choice and future answer types) they submitted for a specific question.

The answer node component uses the completed-activity-media in order to display the answer to a question, similar to the "Show previous answer functionality" in activity nodes.

## Changes

1. Create a AnswerMedia component for displaying the answer node to the end user
2. Create AnswerForm to let editors create the answer node
3. Renaming files and rearranging  folder 

- .../ActivityMedia/AudioRecorder.vue → .../ActivityMedia/Question/AudioRecorder.vue 
- .../ActivityMedia/TextQuestion.vue → .../ActivityMedia/Question/TextQuestion.vue 
-  .../ActivityMedia/AnswerButton.vue → .../ActivityMedia/Question/AnswerButton.vue 

## Screenshot

https://user-images.githubusercontent.com/34692466/124504852-1ea96300-dd7d-11eb-9548-d0336b5a4085.mp4

## Issue Linkage
Closes (issue #1026 )
## PR Dependency
Depends on: (PR #1068)
## Automated Testing
Added answer.spec.js to test the following answer node functionality
1.  Creating an Answer Node and displaying the correct answer for text answers including custom follow up text


odotadot:
TODO:
- get AnswerMedia working
- not sure if the AnswerMedia should display answer to more than 1 question
- fix Cypress tests

templates/vue/src/components/Lightbox/media/AnswerMedia.vue
```diff
@@ -0,0 +1,109 @@
...
+  },
+  props: {
+    node: {
+      type: Object,
+      required: true,
+    },
+  },
+  computed: {
+    ...mapState(["userAnswers"]),
+    ...mapGetters(["getEntry", "getQuestion", "getAnswers"]),
```
harry02031, line 64: Just make sure to delete getEntry

harry02031 (commented):
Just make sure to address my comment on this, but everything else looks good.

templates/vue/src/components/Lightbox/media/AnswerMedia.vue
```diff
@@ -0,0 +1,109 @@
...
+    answer() {
+      return this.node.answers
+    },
+    question() {
+      return this.getQuestion(this.answer.questionID)
+    },
+    followUpText() {
```
harry02031, line 71: There is already a followUp structure under each question. FollowUpText is not relevant anymore and should be referred to as question.followUp.text

odotadot: I think this line of code is correct, as the AnswerForm allows the user to override the question.followUp.text with their own followup text. If left blank, then it will default  to question.followUp.text. I changed the code in the template to reflect this.

templates/vue/src/components/Lightbox/media/ActivityMedia/TapestryActivity.vue
```diff
@@ -2,12 +2,13 @@
   <div class="tapestry-activity">
     <div class="icon"><tapestry-icon :icon="type" /></div>
     <div v-if="type === 'text'" class="text">{{ answerData }}</div>
-    <audio v-if="type === 'audio'" controls :src="answerData"></audio>
+    <audio v-if="type === 'audio'" controls :src="audioUrl"></audio>
```
harry02031, line 5: This should be answerData as it is. I see that you cannot get the audio now if you show previous answer to an audio activity.

templates/vue/src/components/Lightbox/media/ActivityMedia/TapestryActivity.vue
```diff
@@ -25,6 +26,11 @@ export default {
       required: true,
     },
   },
+  computed: {
```
harry02031, line 29: This is not needed anymore, as the 1043 handles this logic.

templates/vue/src/components/Lightbox/media/ActivityMedia/TapestryActivity.vue
```diff
@@ -25,6 +26,11 @@ export default {
...
+    audioUrl() {
+      return wpData.uploadDirArray.baseurl + "/" + this.answerData
+    },
+  },
```
harry02031, line 33: This is not needed anymore, as the 1043 handles this logic.

classes/class.tapestry-node.php
```diff
@@ -99,6 +100,7 @@ public function __construct($tapestryPostId = 0, $nodeMetaId = 0)
             'lng' => '',
         ];
         $this->popup = null;
+        $this->answers = (object) [];
```
EldadZZipori, line 103: I think that we can use new stdClass() here instead of converting an array to an object. 
https://www.php.net/manual/en/language.types.object.php

harry02031 (approved):
Everything looks good now. Functionality are working as expected.

classes/class.tapestry-node.php
```diff
@@ -47,6 +47,7 @@ class TapestryNode implements ITapestryNode
     private $references;
     private $mapCoordinates;
     private $popup;
+    private $answers;
```
wynnset, line 50: Content-type-specific properties should be stored under typeData.

templates/vue/src/components/modals/NodeModal/forms/ContentForm/AnswerForm.vue
```diff
@@ -0,0 +1,124 @@
...
+        <template v-slot="slotProps">
+          <p>
+            {{ slotProps.option.text }}
+          </p>
+        </template>
+      </combobox>
+    </b-form-group>
+    <b-form-group
+      label="Show this text first: "
+      description="If empty, will default to the follow-up text of the activity form."
```
wynnset, line 40: Instead of this, can we show the actual question text in the placeholder for this input?

templates/vue/src/components/modals/NodeModal/forms/ContentForm/AnswerForm.vue
```diff
@@ -0,0 +1,124 @@
+<template>
+  <div>
+    <b-form-group data-qa="activity-combobox" label="Select an activity:">
```
wynnset, line 3: Although I like the directive language here, to keep consistent with other contents, let's just use "Activity" as the label here (without a colon).

templates/vue/src/components/modals/NodeModal/forms/ContentForm/AnswerForm.vue
```diff
@@ -0,0 +1,124 @@
...
+        <template v-slot="slotProps">
+          <p>
+            {{ slotProps.option.title }}
+          </p>
+        </template>
+      </combobox>
+    </b-form-group>
+    <b-form-group
+      data-qa="question-select"
+      label="Select a question from that activity:"
```
wynnset, line 21: Same as above ... "Question"

templates/vue/src/components/modals/NodeModal/forms/ContentForm/AnswerForm.vue
```diff
@@ -0,0 +1,124 @@
...
+    >
+      <combobox
+        v-model="currentQuestion"
+        :options="getCurrentQuestions"
+        data-qa="choose-question"
+        item-text="text"
+        item-value="id"
+        empty-message="There is no activity selected yet."
```
wynnset, line 29: "Please select an activity first."

templates/vue/src/components/modals/NodeModal/forms/ContentForm/AnswerForm.vue
```diff
@@ -0,0 +1,124 @@
...
+  data() {
+    return {
+      activityNodeId: "",
+      currentQuestion: "",
+      followUpText: "",
+    }
+  },
+  computed: {
+    ...mapState(["nodes"]),
+    currentActivityNodes() {
```
wynnset, line 69: Since this is a computed property, it should be named like it's a property... i.e. `activityNodes`

templates/vue/src/components/modals/NodeModal/forms/ContentForm/AnswerForm.vue
```diff
@@ -0,0 +1,124 @@
...
+      const activityNodes = Object.values(this.nodes).filter(
+        node => node.mediaType == "activity"
+      )
+      return activityNodes
+    },
+    getCurrentQuestions() {
```
wynnset, line 75: Ditto here.. probably `currentQuestions`

templates/vue/src/components/modals/NodeModal/forms/ContentForm/AnswerForm.vue
```diff
@@ -0,0 +1,124 @@
...
+      this.followUpText = prevAnswers.followUpText
+    }
+  },
+  updated() {
+    const answerObject = {
+      activityID: this.activityNodeId,
+      questionID: this.currentQuestion,
+      followUpText: this.followUpText,
+    }
+    this.node.answers = answerObject
```
wynnset, line 119: As per above, this should be stored under `typeData`, not `answers`. You can either put each property directly under `typeData` or otherwise put the whole object into `typeData`. Note that `typeData` may have other properties in it that should be kept. Putting the properties directly under `typeData` has the benefit that you do not need these `mounted` and `updated` functions as you can just set the `v-model` to `this.node.typeData.questionId`, for example.

I also don't think `answers` is a good name for the property if you want to group these in `typeData` because these are not the answers rather just the needed information to display a single answer. Maybe `answer` would be better.

templates/vue/src/components/modals/NodeModal/forms/ContentForm/AnswerForm.vue
```diff
@@ -0,0 +1,124 @@
...
+        let selectedQuestion = []
+        selectedQuestion = this.getCurrentQuestions.filter(
+          question => question.id == id
+        )
+        this.currentQuestion = selectedQuestion[0].id
+      }
+    },
+    followUpText(text) {
+      this.followUpText = text
+    },
```
wynnset, line 103: This is strange... You are watching the property for change and when it changes you are setting it to the thing it just changed to?

templates/vue/src/components/modals/NodeModal/forms/ContentForm/AnswerForm.vue
```diff
@@ -0,0 +1,124 @@
...
+  props: {
+    node: {
+      type: Object,
+      required: true,
+    },
+  },
+  data() {
+    return {
+      activityNodeId: "",
+      currentQuestion: "",
```
wynnset, line 63: Can we keep these consistent? Also, "current" is a bit ambiguous.. can we use `selected` instead? Also, if something is an ID it should say so... I recommend renaming these to `selectedNodeId` and `selectedQuestionId`

templates/vue/src/components/modals/NodeModal/forms/ContentForm/AnswerForm.vue
```diff
@@ -0,0 +1,124 @@
...
+      followUpText: "",
```
wynnset, line 64: Can we avoid the use of `followUp` here? It doesn't make as much sense in the context of a singular answer. How about just `text`?

wynnset (changes requested):
Some UI changes needed in addition to the code changes:
- [x] The "Advanced" optgroup should be at the very end. Also, let's throw all the other items under a "Basic" optgroup at the beginning. image
- [x] We do not need the answer type icon on the body since we already have it on the tab. image
- [x] The text on the inactive tab should be white for better visibility
- [ ] Can we make the tab background, tab hover border, and the top border on the tab content a darker color? Currently it's too bright for everything else.
- [x] Please do not center the text answer
- [ ] Please add a minimum height to the tab content (maybe around 200px)

templates/vue/src/components/modals/NodeModal/forms/ContentForm/AnswerForm.vue
EldadZZipori: I think we should also flip their order. First select a question, then we will show they questions text as a placeholder. What do you think?
Screenshot from 2021-06-24 17-17-31

templates/vue/src/components/Lightbox/media/AnswerMedia.vue
```diff
@@ -0,0 +1,113 @@
...
+    node: {
+      type: Object,
+      required: true,
+    },
+  },
+  computed: {
+    ...mapState(["userAnswers"]),
+    ...mapGetters(["getQuestion", "getAnswers"]),
+    answer() {
+      return this.node.typeData
```
wynnset, line 61: This is a bit misleading because answer != typeData really. Rather than use this computed property, I think it would make more sense to use the actual thing you are looking for, e.g. `this.node.typeDada.questionId`, etc. Remember that typeData could potentially contain other fields that do not necessarily mean this node is an answer node.

templates/vue/src/components/Lightbox/media/AnswerMedia.vue
```diff
@@ -0,0 +1,113 @@
...
+    },
+  },
+  mounted() {
+    this.$emit("complete")
+    this.$emit("load")
+  },
+  methods: {
+    getFullUrl(url) {
+      return wpData.uploadDirArray.baseurl + "/" + url
+    },
```
wynnset, line 81: As mentioned above, better to move this to the TapestryActivity component.

templates/vue/src/components/modals/NodeModal/forms/ContentForm/AnswerForm.vue
```diff
@@ -0,0 +1,84 @@
...
+      const questions = Object.values(this.activityNodes)
+        .filter(node => node.id == this.node.typeData.activityId)
+        .flatMap(node => node.typeData.activity.questions)
+
+      return questions
+    },
+    originalQuestionText() {
+      return this.availableQuestions.find(
+        question => question.id === this.node.typeData.questionId
+      ).text
```
wynnset, line 78: If you are not absolutely guaranteed to find a result here, you might consider using [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).

templates/vue/src/components/modals/NodeModal/forms/ContentForm/AnswerForm.vue
```diff
@@ -0,0 +1,84 @@
...
+    },
+  },
+}
+</script>
+
+<style scoped></style>
```
wynnset, line 84: This is not necessary

wynnset (changes requested):

templates/vue/src/components/Lightbox/media/AnswerMedia.vue
```diff
@@ -0,0 +1,113 @@
...
+            :answerData="questionAnswer[1]"
+          ></tapestry-activity>
+        </b-tab>
+      </b-tabs>
+      <div v-show="!hasAnswer" class="media-wrapper">
+        You have not completed this question yet.
+      </div>
+    </div>
+    <div v-else>
+      <p>Please fill in the answer form.</p>
```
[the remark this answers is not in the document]
EldadZZipori: Seems like a preventive mechanism in case the form was submitted without filling it. This cannot happen anymore as we have validation. Removing this.

templates/vue/src/components/Lightbox/media/AnswerMedia.vue
EldadZZipori: This seems like a computed component used in order to not access this.node.typdeData all the time. I changed the name for this property for clarity.

templates/vue/src/components/modals/NodeModal/forms/ContentForm/AnswerForm.vue
EldadZZipori: @wynnset This should be guaranteed to find something as you cannot make question without a text .

wynnset: @EldadZZipori I was talking about finding a match in `this.availableQuestions`. If it returns no results, it might cause a console error.

wynnset (approved):

wynnset (approved):

wynnset:
@odotadot / @EldadZZipori this PR is missing proper description. Please update that as I need the information for merging this in.

PR #1064, opened 2021-06-29
1061 - TYDE2 - Add TYDE Mode

EldadZZipori:
## Task Description
Add a new toggle switch under Tapestry Settings > Advanced under Geography map called "TYDE mode". When turned on, the author can identify which nodes to use as the default node for Dyads and for Youth.

(Note: Make sure to pull the latest tyde2 branch. Once you pull it, you will need to deactivate and again activate the Tapestry plugin and this will cause you to see the new Youth and Dyad roles added in which you will need for this PR).

If this is turned on for the tapestry and the user doesn't have ability to edit the root node, the following should apply:

Ensure they cannot exit out of the lightbox.

When the page loads:
Hide the tapestry from them.
Automatically open the default node in fullscreen lightbox when the page loads. (Use the router for this)

Add tabs to the top of the lightbox:
These tabs will allow the user to switch between the lightbox content and the TYDE activities. The first tab should be named "Content" and just be the regular lightbox content that is there already. There should be 3 other tabs that, when pressed, each take the user to the given tyde activity (for now use placeholder content):
My profile
My goals
My circle of support
Each of these tabs may have a sidebar to the left to navigate sub-content.
## Changes
Additional settings properties available for change by the tapestries admin/editors
The user can enable TYDE Mode under Settings->Advanced->Enable TYDE mode
Here the user needs to select a default node to automatically open for every role (If not selected will default to root node)
```
tydeModeEnabled: boolean,
tydeModeDefualtNodes: {
 role: nodeId,
}
```

canEditTapestry is used to determine whether to show apestry-main or navbar components.
```
import { canEditTapestry } from "@/services/wp"

```

## Conceptual Design
image
## Issue Linkage
Closes (issue #1061 )


templates/vue/src/components/tyde/Navbar.vue
```diff
@@ -0,0 +1,98 @@
...
+        <b-nav-item
+          v-for="tab in tabs"
+          :key="tab.name"
+          :ref="tab.name + '-tab'"
+          :to="tab.link"
+          active-class="active"
+          link-classes="link"
+        >
+          <tyde-icon class="selected" :icon="tab.name" />
+          <tyde-icon class="unselected" :icon="tab.name + '-unselected'" />
```
EldadZZipori, line 14: The reason this is being done with css is because active is only set as a class and is not accessible by JS code. 
I am looking for an idea to improve this to use v-if instead, open for suggestions.

templates/vue/src/App.vue
```diff
@@ -47,7 +47,7 @@ export default {
     }
   },
   computed: {
-    ...mapState(["nodes"]),
+    ...mapState(["nodes", "rootId", "settings"]),
```
wynnset, line 50: Maybe we don't need this anymore? I think we should revert this whole file?

templates/vue/src/components/Lightbox/index.vue
```diff
@@ -82,10 +87,11 @@ export default {
       },
       showCompletionScreen: false,
       rowRefs: [],
+      selectedTab: "",
     }
   },
   computed: {
-    ...mapState(["h5pSettings", "rootId"]),
+    ...mapState(["h5pSettings", "rootId", "displayTydeMode"]),
```
wynnset, line 94: Rather than have this in the store, can we just have a computed property in this component? We do not seem to be using this anywhere else for the most part?

templates/vue/src/components/Lightbox/index.vue
```diff
@@ -82,10 +87,11 @@ export default {
       },
       showCompletionScreen: false,
       rowRefs: [],
+      selectedTab: "",
```
wynnset, line 90: We don't seem to be using this in this component anywhere?

templates/vue/src/components/Lightbox/index.vue
```diff
@@ -292,6 +298,9 @@ export default {
         height: this.lightboxDimensions.height,
       }
     },
+    handleTabChange(newTab) {
+      this.selectedTab = newTab
```
wynnset, line 302: Since `selectedTab` is not used anywhere here, this whole method and the event handler on the navbar component can be removed too?

templates/vue/src/components/TapestryApp.vue
```diff
@@ -29,16 +30,29 @@ export default {
     return {
       loading: true,
       viewBox: "2200 2700 1600 1100",
+      currentUser: getCurrentUser(),
```
wynnset, line 33: Prefer we use this function directly in the methods that need it.

templates/vue/src/components/TapestryApp.vue
```diff
@@ -29,16 +30,29 @@ export default {
...
+    isTydeView() {
+      const isAdmin = this.currentUser.roles.some(role => role === "administrator")
+      if (!isAdmin && this.settings.tydeModeEnabled) {
+        const rootNode = this.nodes[this.rootId]
+        const hasEditPermissions = this.currentUser.roles.some(role => {
+          return rootNode.permissions[role].some(premission => premission === "edit")
+        })
+        return !hasEditPermissions
+      }
+      return false
```
wynnset, line 54: We already have a function to determine if the user can edit the tapestry that should be used (you may need to import that function from the `wp.js` service file).
```suggestion
      return this.settings.tydeModeEnabled && !wp.canEditTapestry()
```

wynnset (changes requested):
Unfortunately, this doesn't work for me. I turned on tyde mode and logged in as a subscriber, but the usual tapestry showed up for me. I think simplifying the logic as suggested in the code review might fix this. I also noticed that we have "Authenticated" showing up as a role, but that's not a role:
image

templates/vue/src/App.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <loading v-if="loading" data-qa="tapestry-loading" style="height: 75vh;"></loading>
   <div v-else id="app">
+    <navbar v-if="tydeMode.displayTydeMode"></navbar>
```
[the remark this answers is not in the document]
EldadZZipori: I dont know why you see this code. It is not there anymore. navbar is in the lightbox as the task description requires.

templates/vue/src/components/tyde/Navbar.vue
```diff
@@ -0,0 +1,84 @@
...
+  },
+  data() {
+    return {
+      selectedTab: "tyde",
+    }
+  },
+  computed: {
+    ...mapState(["settings"]),
+    tabs() {
+      return ["tyde", "profile", "goals", "cos"]
```
wynnset, line 38: Can we name the first tab "default"?

templates/vue/src/components/TapestryApp.vue
```diff
@@ -32,7 +32,8 @@ export default {
     }
   },
   computed: {
-    ...mapState(["nodes", "links", "selection", "settings", "rootId"]),
+    ...mapState(["nodes", "links", "selection", "settings"]),
```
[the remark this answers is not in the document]
EldadZZipori: Removed by accident when refactoring. "rootId" is actually needed for this PR.

templates/vue/src/App.vue
EldadZZipori: Correct

templates/vue/src/components/Lightbox/index.vue
EldadZZipori: Currently it is not used but it is relevant for tyde activities because we want to alternate between normal node and tyde views such as cos

templates/vue/src/components/Lightbox/index.vue
EldadZZipori: Ditto as above

EldadZZipori:
Fixed wrong roles. Should no work

wynnset (approved):

PR #1066, opened 2021-06-30
1065 - Distinguish some page child nodes as secondary nodes and show such items under the main menu 

eabean:
## Changes
- If a child of a Page MultiContent node is selected to be a secondary node (in its appearance form), create a new menu for it in the Page Menu
- If secondary nodes exist, the lightbox will only show items in the active menu currently selected
- Lightbox should only show items in the currently active menu
- The route URLs will change when the user clicks on the page menu item, opening it
- Menus will show all direct children of the node in the title of the menu. For further nested content, only children of multi-content nodes will be shown

## Change Details

Here are details of the data structures related to the Page Menu components as well as the changeRow() function in MultiContentMedia.vue. This PR was designed such that all route control would be centralized in changeRows().

`rows()`: A menu.  An array of objects, with each object representing a child of the MultiContent node. Each object looks like { node, children }, where the node value is the child node, and the children value are the children of the child node.

`menuGroups()`: A 2D array of menus. If any of the nodes in rows() is selected to be a secondary node, then a new menu will be created for it.

`filteredMenuGroups`: `menuGroups()`, but with the first menu item (which is the same as the menu title) removed for aesthetic purposes.

`pageMenuData`: Prop used to pass data so that the page menu interactions control the lightbox content. 
•	`menuIndex`: eventually passed to `PageRows.vue` so that only the active menus content is in the lightbox. Represents the selected menu index of menus in `menuGroups()`.
•	`nodeId`: eventually passed to `changeRow()` to control the route. Represents the nodeId of the menu title/item clicked on.
•	`context`: eventually passed to `changeRow()` to control the route. Represents whether the rootId or parentId of the menu title/item clicked on is a multicontent type. 

`changeRow()`: Function in `MultiContentMedia.vue` that controls the routes related to multi-content media. Takes in a parameter `rowInfo` which is { rowId, context }.  Consists of 4 major if/else cases.

Case 1: Pushes routes like these 
`/nodes/<rootId>/view/<parentId>/rows/< nodeId>`
`/nodes/<rootId>/view/<parentId>/rows/< nodeId1>,< nodeId2>…< nodeIdn>`
`rowInfo : { rowId: <nodeId>, context: “multi-content | page | accordion”}`

Case 2: Closes multi-content tabs.
`rowInfo : { rowId: null,  context: “multi-content | page | accordion”}`

Case 3: Pushes routes like these 
`/nodes/<rootId>/view/<nodeId>`
`rowInfo : { rowId: <nodeId>,  context: “” }`

 Case 4: Closes multi-content tabs.
`rowInfo: { rowId: null, context: “”}`

Note: The first menu group is slightly different from the subsequent ones because the titles/items of menus in menuGroup[1]...menuGroup[n] are considered more nested than the ones in menuGroup[0].

## Screenshot

https://user-images.githubusercontent.com/34692466/127394487-c180eb75-92f4-4a3f-ab42-08a30dcd3403.mp4

## Issue Linkage
Closes (issue #1065 )
## PR Dependency
N/A
## Automated Testing
N/A


odotadot:
The styling probably needs work. Let me know any suggestions.

EldadZZipori (commented):
Codes look really good.

EldadZZipori (approved):

wynnset:
This looks great. I forgot to mention that items that are on a secondary menu should not be on the same scrolling screen as the other items, but should rather have their own. So basically, we'd need to only show the items in that secondary menu when that item or one of its children is clicked on. @odotadot is that clear? Let me know if you want us to chat about this further.

odotadot:
> This looks great. I forgot to mention that the content of items that are on a secondary menu should not be on the same scrolling screen as the other items (on the right side), but should rather have their own. So basically, we'd need to only show the items in that secondary menu when that item or one of its children is clicked on. @odotadot is that clear? Let me know if you want us to chat about this further.

@wynnset So only show the "Child" menu when the Child content is clicked on in the lightbox of the page-multicontent Parent? And only show 1 menu at a time in the sidebar?
image

wynnset:
Let's talk about this after our next meeting.

odotadot:
Concerns:
1. Is the current implementation fine? I made a HeadlessMultiContent for each menu as this seemed like the least-intrusive way to switch the lightbox content to the clicked menu item. 
2. The title of the Lightbox is still the main parent node for whichever menu is active.

templates/vue/src/components/modals/NodeModal/forms/AppearanceForm.vue
```diff
@@ -95,6 +95,15 @@
                 Hide media button
               </b-form-checkbox>
             </b-form-group>
+            <b-form-group>
+              <b-form-checkbox
+                v-show="hasPageMultiContentParent"
```
wynnset, line 100: Why not use `v-if` here? `v-show only stylistically hides the element which is generally not what we want.

templates/vue/src/components/modals/NodeModal/forms/AppearanceForm.vue
```diff
@@ -95,6 +95,15 @@
...
+                v-model="node.typeData.isSecondaryNode"
+                data-qa="node-appearance-secondary-node"
+              >
+                Is secondary node
```
wynnset, line 104:
```suggestion
                Show this node and its children in a secondary menu
```

templates/vue/src/components/modals/NodeModal/forms/AppearanceForm.vue
```diff
@@ -181,6 +192,16 @@ export default {
       }
       return "success"
     },
+    hasPageMultiContentParent() {
+      let parentNode = this.getNode(this.getParent(this.node.id))
```
wynnset, line 196: If the node is the root node, it might not have a parent in which case I think you should test to make sure it doesn't throw any errors in the console.

templates/vue/src/components/Lightbox/index.vue
```diff
@@ -82,6 +84,7 @@ export default {
       },
       showCompletionScreen: false,
       rowRefs: [],
+      activeMenuNode: this.nodeId,
```
wynnset, line 87: Since this is an ID, it should more accurately be called `activeMenuNodeId`

templates/vue/src/components/Lightbox/index.vue
```diff
@@ -16,6 +16,7 @@
       :node="node"
       :row-id="rowId"
       :sub-row-id="subRowId"
+      :activeMenuNode="activeMenuNode"
```
wynnset, line 19: Rather than doing this, couldn't you simply change the `node` prop here to be either the original node or the active menu node? (maybe create a computed property that uses the activeMenuNodeId to return which node it should display) That way, you theoretically wouldn't have to change the multiContentMedia  component.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
```diff
@@ -250,5 +298,19 @@ export default {
...
+    margin-top: 8px;
+    margin-bottom: 8px;
+  }
+
+  .menu {
+    z-index: 100;
+    position: relative;
+    float: none;
+    background: white;
+  }
```
wynnset, line 314: I suggest looking into using Bootstrap's vertical nav ([Bootstrap-Vue component](https://bootstrap-vue.org/docs/components/nav#vertical-variation)) instead of having to custom create it like this.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
```diff
@@ -35,16 +35,38 @@
...
+        <div v-for="(menu, index) in menuGroups" :key="index" class="menu-wrapper">
+          <b-card class="menu">
+            <div>
+              <b-card-text v-if="index === 0">
+                {{ getMenuName(index) }}
+              </b-card-text>
+              <b-card-text v-else v-b-toggle="`collapse-${index}`">
+                {{ getMenuName(index) }}
+              </b-card-text>
+            </div>
```
wynnset, line 47: Could this be replaced with something like:
```suggestion
            <b-card-text v-b-toggle="index === 0 ? `` : `collapse-${index}`">
              {{ getMenuName(index) }}
            </b-card-text>
```

wynnset (commented):
@odotadot please see my comments and let me know if you have any questions

templates/vue/src/components/modals/NodeModal/forms/AppearanceForm.vue
odotadot: `parentNode` is undefined if no parent node is found. No errors in the console are generated in this line.

EldadZZipori:
All functionality works great. 

I believe there is an override of some functionality in the router with this PR. the row url parameter is not updated anymore when selecting a different node in the sub-menu. 
ezgif com-gif-maker

There is only one thing that look weird to me with the UI is that there is always a duplicated text and that the user will always have one parent item and all others will be children in the list for secondary nodes.
image

odotadot:
> I believe there is an override of some functionality in the router with this PR. the row url parameter is not updated anymore when selecting a different node in the sub-menu.

I compared with the old PageMenu on master, and it seems the behaviour is still the same. I don't think the URL changes when clicking the page menu, only when you interact with the lightbox contents (hovering, clicking stuff).

https://user-images.githubusercontent.com/34692466/126565976-ceac243f-30db-49eb-9d33-670c0deaface.mp4

> There is only one thing that look weird to me with the UI is that there is always a duplicated text and that the user will always have one parent item and all others will be children in the list for secondary nodes.

For the sake of simplicity I would like to leave it as is. I tried to implement this, and it seems it introduces a lot of conditionals to the code for the sake of styling. For example, if a secondary node has no children, there's no way to click on it to switch to its content in the lightbox. In nested contexts (page menus in page menus, secondary nodes in secondary nodes), even more conditions are added to check if the page menu item should be visible.

@wynnset What do you think?

wynnset:
> I don't think the URL changes when clicking the page menu

This is a bug. It should change AND it should highlight the menu item you just clicked on. Can we please get this fixed with this PR?

> There is only one thing that look weird to me with the UI is that there is always a duplicated text and that the user will always have one parent item and all others will be children in the list for secondary nodes.
> What do you think?

I think we should fix this. It definitely does look weird. It should be pretty simple to fix this - not sure what conditionals you are talking about but it sounds pretty straightforward... if it's a secondary page menu simply pop the first array element no?

EldadZZipori:
Everything looks good and works as intended.

templates/vue/src/components/Lightbox/index.vue
```diff
@@ -15,6 +15,7 @@
       v-if="node.mediaType === 'multi-content'"
       :node="node"
       :row-id="rowId"
+      :pageMenuData="pageMenuData"
```
wynnset, line 18: Please format attributes in kebab-case

templates/vue/src/components/Lightbox/index.vue
```diff
@@ -24,6 +25,7 @@
       :node="node"
       :rowRefs="rowRefs"
       :dimensions="dimensions"
+      @handlePageMenuClick="handlePageMenuClick"
```
wynnset, line 28: Please format attributes in kebab-case

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
```diff
@@ -35,16 +35,39 @@
...
+              :id="`collapse-${index}`"
+              :visible="index === 0"
+              class="mt-2"
+            >
+              <ul class="page-menu-item fa-ul">
+                <page-menu-item
+                  v-for="row in menu"
+                  :key="row.node.id"
+                  :node="row.node"
+                  :lockRows="lockRows"
```
wynnset, line 62: Please format attributes in kebab-case

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
```diff
@@ -35,16 +35,39 @@
...
+                  :disabled="disabledRow(row.node)"
+                  @scroll-to="scrollToRef"
+                  @handleMenuItemClick="handleMenuItemClick"
```
wynnset, line 65: Please format attributes in kebab-case

templates/vue/src/components/Lightbox/media/MultiContentMedia/index.vue
```diff
@@ -36,6 +36,7 @@
       :subRowId="subRowId"
       :context="context"
       :level="level"
+      :pageMenuData="pageMenuData"
```
wynnset, line 39: Please format attributes in kebab-case

wynnset (changes requested):
Great job with the refactor @odotadot ! Just a few minor suggestions to improve the code and this will be ready to go.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
```diff
@@ -35,16 +35,39 @@
...
-            :lockRows="lockRows"
-            :disabled="disabledRow(row.node)"
-            @scroll-to="scrollToRef"
-          />
-        </ul>
+        <div
+          v-for="(menu, index) in filteredMenuGroups"
+          :key="index"
+          style="padding-top: 4px; padding-bottom: 4px;"
+        >
```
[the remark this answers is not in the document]
odotadot: For some reason if I remove the `div`, the styling doesn't work. I've also tried using class styling, but it still doesn't work.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
```diff
@@ -35,16 +35,39 @@
...
+          <b-card>
```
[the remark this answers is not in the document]
odotadot: I'm not sure what this component's template would look like. The majority of this PR's code is located in PageMenu/index.vue, which makes sense to me. It could be possible to create a new component that combines the PageMenu and PageRows components, but I imagine it would take a lot of refactoring, which might be better in a separate PR.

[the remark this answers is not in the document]
wynnset: Sorry I didn't mean to combine components. I rather meant to create a new component for menu groups that you have here so you can just loop over the menu groups and all the logic for those would go into that component (so basically each of these `b-card`s that you have here would be a MenuGroup component).

I also noticed that here you are putting the `v-for` on a div that includes just a `b-card` component inside. I don't see why you need the `div` -- you can just add the `v-for` right on the `b-card`.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
```diff
@@ -158,6 +217,36 @@ export default {
         }
       })
     },
+    getMenuName(index) {
+      return index === 0 ? this.node.title : this.menuGroups[index][0].node.title
+    },
+    getMenuIndexFromNodeId(nodeId) {
```
wynnset, line 223: I'm still pretty confused what the purpose of a menu index is and why we are not just using IDs for these? It requires all this code to convert between index and Ids

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageRows.vue
```diff
@@ -47,10 +47,7 @@
                 @complete="updateProgress(row.node.id)"
                 @load="handleLoad($refs.rowRefs[index])"
               />
-              <p
-                v-if="row.children.length > 0 && !areAllPopup(row.children)"
-                style="color: white;"
-              >
+              <p v-if="row.children.length > 0" style="color: white;">
```
wynnset, line 50: How come you have removed the `areAllPopup` check here?

wynnset (commented):
Sorry @odotadot just need a few more clarifications here and still hoping you can clean up the PageMenu component by moving a lot of the MenuGroup logic into its own component.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageRows.vue
odotadot: Woops, my mistake. Added it back in.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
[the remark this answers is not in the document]
odotadot: I've created a new MenuGroups.vue component. As for the `div`, I have tried moving the `v-for` and the other attributes to the `b-card`, but for some reason it will break the styling.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
odotadot: The `menuIndex` in the `pageMenuData` object is for switching the lightbox to only show the contents of one menu at a time. In PageRows.vue, it uses the menuIndex to determine which menu group to display. If I just passed the nodeId, I would still have to perform the calculation to determine which menu the node belongs to.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
```diff
@@ -35,30 +35,27 @@
...
-          <page-menu-item
-            v-for="row in rows"
-            :key="row.node.id"
-            :node="row.node"
-            :lockRows="lockRows"
-            :disabled="disabledRow(row.node)"
-            @scroll-to="scrollToRef"
-          />
-        </ul>
+        <menu-groups
```
wynnset, line 38: Most of our components are single items, so I think it would be better if this is a representation of a single menu group rather than all menu groups. That way, you can put the v-for on the single menu-group component and have the component only responsible for what's inside of that one menu group. This is generally the preferred pattern in Vue.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/MenuGroups.vue
```diff
@@ -0,0 +1,148 @@
+<template>
+  <b-container>
+    <div v-for="(menu, index) in filteredMenuGroups" :key="index" class="py-1">
+      <b-card>
```
wynnset, line 4: I think I mentioned this before, by I don't see why you need a div wrapper here - what's wrong with just having the card by itself without a wrapper? Why add a div?

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/MenuGroups.vue
```diff
@@ -0,0 +1,148 @@
+<template>
+  <b-container>
+    <div v-for="(menu, index) in filteredMenuGroups" :key="index" class="py-1">
+      <b-card>
+        <b-card-text
+          v-b-toggle="index === 0 || menu.length === 0 ? `` : `collapse-${index}`"
```
wynnset, line 6: If you change this component to a single menu group, you can easily pass this as a prop so this component doesn't have to deal with that responsibility.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/MenuGroups.vue
```diff
@@ -0,0 +1,148 @@
...
+    getNodeIdFromMenuIndex(index) {
+      return index === 0 ? this.node.id : this.menuGroups[index][0].node.id
+    },
+    handleMenuTitleClick(index) {
+      const pageMenuData = {
+        menuIndex: index,
+        nodeId: this.getNodeIdFromMenuIndex(index),
+        context: index === 0 ? "" : "page",
+      }
+      this.$emit("handle-page-menu-click", pageMenuData)
```
wynnset, line 130: Since this is an event, it's better to just call the even something like "title-clicked" or something as the components generally just tell the parent what events are happening and it's up to the parent component to decide to handle that event.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/MenuGroups.vue
```diff
@@ -0,0 +1,148 @@
...
+    },
+    handleMenuItemClick(nodeId) {
+      this.$router.push({
+        ...this.$route,
+        query: { ...this.$route.query, row: nodeId },
+      })
+      const pageMenuData = {
+        menuIndex: this.getMenuIndexFromNodeId(nodeId),
```
wynnset, line 138: If you change this component to a single menu group, you would not need to pass menuIndex to the parent as the parent would be keeping track of the menu groups.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/MenuGroups.vue
```diff
@@ -0,0 +1,148 @@
...
+              children: [],
+              node: childNode,
+            }
+            newMenu.push(newRow)
+          })
+          filteredMenu.push(newMenu)
+        }
+      })
+      return filteredMenu
+    },
```
wynnset, line 88: If you change this to a single menu group, you can move all of this to the parent.

templates/vue/src/components/Lightbox/index.vue
```diff
@@ -24,6 +25,7 @@
       :node="node"
       :rowRefs="rowRefs"
       :dimensions="dimensions"
+      @handle-page-menu-click="handlePageMenuClick"
     />
```
wynnset, line 29: Can we move this into `multi-content-media` as it's only applicable when that's the content type? I want to keep Lightbox uncluttered if possible.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/MenuGroups.vue
```diff
@@ -0,0 +1,148 @@
...
+    menuGroups() {
+      const menu = []
+      const mainMenu = []
+      this.rows.forEach(row => {
+        row.node.typeData.isSecondaryNode ? menu.push([row]) : mainMenu.push(row)
+      })
+      menu.unshift(mainMenu)
+      return menu
+    },
```
wynnset, line 68: I see that you have this both in here and also in the PageRows component. Is there a way to centralize this or pass it to both components as a prop?

wynnset (changes requested):

wynnset:
I'm experiencing a bizarre bug on this branch where I click to open one multi-content node, but then another one opens instead. @odotadot do you know about this issue? That is not happening on master.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/PageMenuItem.vue
```diff
@@ -95,18 +96,22 @@ export default {
...
       )
     },
-    handleTitleClick() {
-      this.scrollToRow()
+    nestedItemClick() {
+      this.itemClick()
     },
-    scrollToRow(nodeId) {
-      if (!nodeId) {
+    itemClick(nodeId) {
```
wynnset, line 102: It's better to name event handlers "handle..." so this is better called "handleItemClick"

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/PageMenuItem.vue
```diff
@@ -95,18 +96,22 @@ export default {
         !node.unlocked
       )
     },
-    handleTitleClick() {
-      this.scrollToRow()
+    nestedItemClick() {
+      this.itemClick()
```
wynnset, line 100: Why have this function when you can just call `itemClick` (or `handleItemClick`) directly

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/MenuGroup.vue
```diff
@@ -0,0 +1,86 @@
+<template>
+  <b-card>
+    <b-card-text v-b-toggle="canToggleMenu" @click="titleClicked">
```
wynnset, line 3:
```suggestion
    <b-card-text v-b-toggle="canToggleMenu" @click="handleTitleClick">

```

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
```diff
@@ -98,11 +104,36 @@ export default {
...
+        row.node.typeData.isSecondaryNode
+          ? menuGroups.push([row])
+          : firstMenuGroup.push(row)
+      })
+      menuGroups.unshift(firstMenuGroup)
+      return menuGroups
     },
-    disabledFrom() {
-      return this.rows.findIndex(row => !row.node.completed)
+    filteredMenuGroups() {
```
wynnset, line 118: Could you please add some comments before rows, menuGroups, and filteredMenuGroups to explain what the **_role_** (or _purpose_) of each of these are and what their data structure is? It would also be great if you could outline the data structure of these in this PR fo clarify. I am having trouble wrapping my head around all of these things that are being used and also some passed into the menu groups as props...

wynnset (changes requested):
Great job with refactoring the menu groups! I am still having some trouble wrapping my head around all of the various data structures being used in the code and was hoping you could outline some of those things for me in this PR and add additional comments in the code as well. Thank you!

(Note I pushed some changes as well so you should do a git pull first)

wynnset:
Many issues after testing:
- [ ] Node in full-screen does not show up as full-screen
- [ ] Secondary menu not showing up in menu
- [ ] Secondary node is showing up at the bottom of the primary node
image

wynnset:
@EldadZZipori as discussed, let's create a new PR with reduced functionality in the interest of time. This new PR should just allow showing the secondary nodes under their own menu underneath the main menu.

wynnset:
This PR is obsolete at this point.

PR #1069, opened 2021-07-01
1068 - Create a compact mode for image file uploads

### LINKED ISSUE #1068: Create a compact mode for image file uploads

wynnset:
## Changes
* Add a compact mode for the file upload component that takes only the thumbnail space

* Add compactMode property to the FileUpload component to determine whether to display in compact mode. Defaults to false.
## Screenshot
image
image
image
## Issue Linkage
Closes #1068
## PR Dependency
Depends on: N/A
## Automated Testing
* No automated testing added


EldadZZipori:
Compact mode will only work if show-image-preview is set to true. 
Should we only note that or make show-image-preview a computed prop as well that will always return true if compact mode is enabled?

wynnset:
> Compact mode will only work if show-image-preview is set to true.
> Should we only note that or make show-image-preview a computed prop as well that will always return true if compact mode is enabled?

Yes, good catch @EldadZZipori I thought about this but never put anything in there (: 
I think it's a good idea to note that and also automatically enable the image preview if compact mode is on. If someone has time, please make this change.

PR #1076, opened 2021-07-14
1074 - Save the last selected node for user and select it on load 

odotadot:
## Changes
- When Tapestry loads, the selected node is the user's last selected node rather than the root node
- If the user is not logged in, the last selected node will be the root node
## Screenshot
https://user-images.githubusercontent.com/34692466/125696267-e3368c29-206f-4648-b8c9-478831dafe2a.mp4
## Issue Linkage
Closes (issue #1074 )
## PR Dependency
N/A
## Automated Testing
N/A


odotadot:
Feels like it works a bit better without the 3 second delay; if you leave the Tapestry too quickly (before 3 seconds) the last selected node isn't saved properly.

EldadZZipori:
We should add a check on whether there is a user before using this PR functionality because when accessing without a user to a tapestry I get this error and the tapestry does not load. 
image

There is a test that is failing on this PR, on local as well. I think we should merge master in and see if any test fails and resolve that.

Other then that code looks clean and clear. The PR works great.

EldadZZipori (approved):
Everything looks good. All tests pass. Code looks simple and clear.

templates/vue/src/components/TapestryMain/TapestryNode/index.vue
```diff
@@ -498,6 +498,9 @@ export default {
           ? this.openNode(this.node.id)
           : this.updateRootNode()
       }
+      setTimeout(() => {
+        client.updateUserLastSelectedNode(this.node.id)
+      }, 3000)
```
wynnset, line 503: There's a couple of problems here. One is that we are not necessarily updating the selected node here (see the lines before this) so this should be moved to the `updateRootNode()` function instead.

Secondly, part of the requirement was that we do not send the request to save a node if the node was changed within a few seconds. So we should clear this interval if the node is changed. This means we should actually emit an event for this and handle it in the parent component as this is only for this one node. Add a data variable to TapestryMain to keep track of the set timeout and then clear that timeout using the `clearTimeout` function when a node is changed. I would also increase the delay to 5 seconds.

Here's the code just fyi:
```
      clearTimeout(this.updateLastSelectedNodeTimer)
      this.updateLastSelectedNodeTimer = setTimeout(() => {
        client.updateUserLastSelectedNode(this.node.id)
      }, 5000)
```

templates/vue/src/components/TapestryMain/index.vue
```diff
@@ -100,6 +101,9 @@ export default {
     selectedId: {
       immediate: true,
       handler(nodeId) {
+        setTimeout(() => {
+          client.updateUserLastSelectedNode(nodeId)
+        }, 3000)
```
wynnset, line 106: You have this function in two places at the moment which means you are sending 2 requests everytime the node is changed.

wynnset (changes requested):
Almost there!

templates/vue/src/components/TapestryMain/index.vue
odotadot: Removed. Now only calls updateUserLastSelectedNode in TapestryMain

templates/vue/src/components/TapestryMain/TapestryNode/index.vue
odotadot: Fixed timeout and saving

wynnset (approved):

wynnset:
This PR should be closed if #1110 gets merged as that is an improved implementation

wynnset:
Closed as superseded by #1110

wynnset:
Closed as superseded by #1110

PR #1078, opened 2021-07-15
1077 - Add ability to reverse link parent-child direction

EldadZZipori:
## Task Description 
At the moment, when a node (B) is added to another node (A), node B is automatically assumed to be the child of node A. This is sometimes not correct, so we want to amend this behaviour. We will do this as follows:
- When a user clicks on the link between two nodes, instead of trying to delete the link, we will show a modal (see attached mockup).
  - This modal will have an arrow showing the current parent-child relationship (in the code this is referred to as "source" and "target" associated with each link)
  - There should be a button "Reverse" underneath that basically switches the "source" and "target". If this is changed, the user should see a warning that their changes are not saved until they press the save button.
  - There should also be a delete button to delete the link (only if allowed, otherwise this should be disabled)
    - If not allowed, show the reason underneath (like in mockup)

Currently the link becomes red when hovered, but we want to change this to blue (same color as the node progress bar color)

## Changes
No additions to the data structure are needed as we just want to by able to reverse the current relationship that exists in the system.
This PR adds the options to flip/reverse a links source/target.

- Moving the delete option of a link to be under the link modal component instead of the tapestry link component
- Adding an option to reverse the link relationship. Includes a store action and mutation, a Tapestry API function, and back-end endpoint addition and function addition to the tapestry class.

## Conceptual Design
image

## Issue Linkage
Closes (issue #1077 )
## PR Dependency
No Dependencies
## Automated Testing
TBA


odotadot:
Testing Notes:
Overall, this PR looks very good. The code looks clean and I could not find many areas for improvement. The source/target link reversals work, even for MC nodes. The only weird behaviour found was for Link Uploaded videos. When replaying, the screen glitches as shown below.
https://user-images.githubusercontent.com/34692466/126225906-0939edc8-2376-4476-8db0-5a142af6bb2a.mp4

odotadot:
Re-tested and everything looks great and works as intended. The failing cypress test is unrelated to this PR.

wynnset (approved):
Code looks great. The only thing missed here is that there should be a confirmation when someone clicks on the Delete button like the one that was there before asking "Are you sure you want to delete the link?" - Make sure to also fix the associated tests. Once fixed, please mark as ready to merge and assign back.

PR #1080, opened 2021-07-16
1079 - Change all lightboxes to have light background with dark text

### LINKED ISSUE #1079: Change all lightboxes to have light background with dark text

Frozemint:
## Changes
This PR changes various lightbox components to use a light background (`background-color: #eeeeee`) with black text (`color: #111`). Components that required an outline has appropriate colours added, such as `tapestry-blue`.

The background color `#eeeeee` is used such that the white background is not too intense, helping reduce eye strain.
## Screenshot
image
## Issue Linkage
Closes #1079 
## PR Dependency
Depends on: N/A
## Automated Testing
This is a visual change, automated testing may not apply


odotadot:
Testing Notes:
The light backgrounds for the lightbox content is looking good. There are still some edge cases to be caught:
1. Answer Nodes have white text on white background
2. Text nodes have dark text on dark backgrounds in multicontent contexts
3. Wordpress posts have dark text on dark backgrounds in multicontent contexts
4.  Accordion rows have a little blue highlight on the right. Is this how it should be?
image
5. For the audio recording activity, the time is white, making it hard to see

Everything else looks good.

odotadot:
Screencaps:
2, 4. 
image
3. 
image
5. 
image

Frozemint:
I've fixed some of the dark on dark and light on light issue. Please test again and let me know if there are still bugs, since I couldn't reproduce some of your issues and I suspect it's because a components is responsible for multiple bugs you are seeing.

odotadot:
Second Testing:
Looking good, a couple more edge cases that remain on  my end:
Only for Page Menu Multicontent nodes:

- Text nodes, Wordpress posts, Answer Nodes still have dark text on dark background. I'm not sure if you can change the Wordpress post text colour, but the other two should be fixable.

image
image

Frozemint:
> Looking good, a couple more edge cases that remain on my end
Should be corrected now.

odotadot:
Re-tested and all content adheres to light mode (i.e having dark text on light background). Code looks good as well.

templates/vue/src/components/Lightbox/media/MultiContentMedia/AccordionRows.vue
```diff
@@ -184,9 +181,9 @@ export default {
     },
     rowBackground() {
       if (this.isMultiContentContext) {
-        let rgb = 40
-        let colorOffset = this.level * 10
-        rgb = colorOffset > rgb ? 0 : rgb - colorOffset
+        let rgb = 187
+        // let colorOffset = this.level * 10
+        // rgb = colorOffset > rgb ? 0 : rgb - colorOffset
```
wynnset, line 186: This dynamic coloring should not be removed as I believe this is what makes the nested accordions visible as they each will have a different shade on top of their parent accordion.

templates/vue/src/components/Lightbox/media/MultiContentMedia/AccordionRows.vue
```diff
@@ -249,13 +249,15 @@ button[disabled] {
...
+  color: #111;
   background: none;
   width: 100%;
   text-align: left;
 }
 
 .accordion-row {
-  background: #262626;
+  color: #eee;
+  background: #eee;
```
wynnset, line 260: Having color and background the same means the text will be invisible. Is this a mistake?

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
```diff
@@ -23,7 +23,7 @@
         data-qa="page-nav-toggle"
         @click="opened = !opened"
       >
-        <i v-if="!opened" class="fas fa-bars fa-lg"></i>
+        <i v-if="!opened" class="fas fa-bars fa-lg" style="color: black;"></i>
```
wynnset, line 26: Ideally, the color should be applied to the element that has a background color, not on the individual children. We want to reduce the number of CSS rules, so it's better to apply it as far up the dom chain as possible to reduce the CSS rules. Remember that if a parent element has a certain color, the child will inherit it unless there's an exception somewhere. If there is an exception, we should remove that exception, preferably. 

If there are WP css, we should override those in the tapestry.css file instead rather than individual components.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageRows.vue
```diff
@@ -165,9 +165,9 @@ export default {
     },
     rowBackground() {
       if (this.isMultiContentContext) {
-        let rgb = 40
-        let colorOffset = this.level * 10
-        rgb = colorOffset > rgb ? 0 : rgb - colorOffset
+        let rgb = 187
+        // let colorOffset = this.level * 10
+        // rgb = colorOffset > rgb ? 0 : rgb - colorOffset
```
wynnset, line 170: Again, this functionality should not be removed. I suggest using the tinycolor2 functionality instead of what's in here.

templates/vue/src/components/Lightbox/media/MultiContentMedia/index.vue
```diff
@@ -310,11 +310,14 @@ button[disabled] {
...
 
 .title {
-  color: #fff;
+  color: #111;
   margin-bottom: 1em;
 }
 
 .media-container {
+  background: #eee;
+  background-color: #eee;
```
wynnset, line 319:
```suggestion
  background-color: #eee;
```

templates/vue/src/components/Lightbox/media/MultiContentMedia/index.vue
```diff
@@ -310,11 +310,14 @@ button[disabled] {
...
+  color: #111;
```
wynnset, line 320: Again, not sure if we need this on a child element - can we keep it on the parent elements instead? Is there an overwrite somewhere that should be removed instead?

templates/vue/src/components/Lightbox/media/MultiContentMedia/index.vue
```diff
@@ -329,6 +332,7 @@ button[disabled] {
 }
 
 .nested-media-title {
+  color: #111;
```
wynnset, line 335: Again, not sure if we need this on a child element - can we keep it on the parent elements instead? Is there an overwrite somewhere that should be removed instead?

templates/vue/src/components/Lightbox/media/TapestryMedia.vue
```diff
@@ -163,7 +163,9 @@ export default {
 
 <style lang="scss" scoped>
 .media-wrapper {
+  color: #111;
```
wynnset, line 166: Again, not sure if we need this on a child element - can we keep it on the parent elements instead? Is there an overwrite somewhere that should be removed instead?

templates/vue/src/components/Lightbox/media/TextMedia.vue
```diff
@@ -39,13 +39,15 @@ export default {
 
 <style lang="scss" scoped>
 .article {
+  color: #111;
```
wynnset, line 42: Again, not sure if we need this on a child element - can we keep it on the parent elements instead? Is there an overwrite somewhere that should be removed instead?

templates/vue/src/components/Lightbox/media/WpPostMedia.vue
```diff
@@ -61,6 +61,7 @@ export default {
   text-align: left;
 
   &-title {
+    color: #111;
```
wynnset, line 64: Again, not sure if we need this on a child element - can we keep it on the parent elements instead? Is there an overwrite somewhere that should be removed instead?

templates/vue/src/components/Lightbox/media/WpPostMedia.vue
```diff
@@ -71,15 +72,18 @@ export default {
   }
 
   article {
-    color: #47425d;
+    color: #111;
```
wynnset, line 75: Again, not sure if we need this on a child element - can we keep it on the parent elements instead? Is there an overwrite somewhere that should be removed instead?

templates/vue/src/components/Lightbox/media/WpPostMedia.vue
```diff
@@ -71,15 +72,18 @@ export default {
...
     font-family: "Source Sans Pro", sans-serif;
     font-size: 16px;
   }
 }
 
 .wp-media-title {
+  color: #111;
```
wynnset, line 82: Again, not sure if we need this on a child element - can we keep it on the parent elements instead? Is there an overwrite somewhere that should be removed instead?

templates/vue/src/components/Lightbox/media/WpPostMedia.vue
```diff
@@ -71,15 +72,18 @@ export default {
...
   text-align: left;
-  margin-bottom: 0.9em;
+  margin-bottom: 0.5em;
+  margin-top: 0.9em;
+  margin-left: 25px;
```
wynnset, line 86:
```suggestion
  margin: 0.9em 0 0.5em 25px;
```

templates/vue/src/components/Lightbox/media/common/CompletedActivityMedia/index.vue
```diff
@@ -77,7 +77,8 @@ export default {
 
 <style lang="scss" scoped>
 .completed-activity-media {
-  background: #262626;
+  background: #eeeeee;
+  color: #111;
```
wynnset, line 81: Again, not sure if we need this on a child element - can we keep it on the parent elements instead? Is there an overwrite somewhere that should be removed instead?

wynnset (changes requested):
This looks good, just want to make sure we are not adding more CSS specificity than needed and that we don't remove important styling that's there now.

templates/vue/src/components/Lightbox/media/MultiContentMedia/AccordionRows.vue
Frozemint: reversed

templates/vue/src/components/Lightbox/media/MultiContentMedia/AccordionRows.vue
Frozemint: Yeah, it's a typo... it's supposed to be the background only

Frozemint: yeah, it's supposed to be the background only... fixed.

templates/vue/src/components/Lightbox/media/WpPostMedia.vue
Frozemint: Yep, moved to parent

templates/vue/src/components/Lightbox/media/WpPostMedia.vue
Frozemint: Yep, moved to parent (per previous...), moving to parent doesnt seem to break anything on my end

templates/vue/src/components/Lightbox/media/WpPostMedia.vue
Frozemint: yep, per previous commits, the title now inherits the colour

templates/vue/src/components/Lightbox/media/MultiContentMedia/index.vue
Frozemint: Yep, I just checked, this rule is redundant... the ultimate element .media-wrapper already has its own rule

templates/vue/src/components/Lightbox/media/MultiContentMedia/index.vue
Frozemint: yep, the media wrapper takes care of this rule... just checked

templates/vue/src/components/Lightbox/media/common/CompletedActivityMedia/index.vue
Frozemint: yeah, it looks like only the background is necessary... everything else is inherited.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
Frozemint: This is the for the navigation sidebar, where every button by default inherits a `color: #fff`. Since the background is changed to a light colour, it's necessary to override the colour of the button to be dark here.
image

templates/vue/src/components/Lightbox/media/TapestryMedia.vue
Frozemint: Yeah, the rule turns out to be unnecessary, since the wrapper only houses the real element that display text below

templates/vue/src/components/Lightbox/media/TextMedia.vue
Frozemint: yeah, turns out the rule isnt needed since it's inhereited.

wynnset (approved):
@Frozemint I reviewed and found some places that were missed, some CSS that was unnecessarily added or not removed, and some spacing issues. I've fixed all of these and going to merge it in now. Thanks!

PR #1084, opened 2021-07-20
1083 - Mark questions as optional and exclude for completion criteria

odotadot:
## Changes
- Allows users to mark certain questions in an Activity node as optional in the ActivityForm
- When a question is optional, it is considered completed in the calculation that determines whether the Activity Node as a whole is complete
- Optional questions have a "Skip" button next to the Submit button for each question type
- Editing the 'optional' value for Questions will reflect in the Activity Node progress bar upon publishing
- Editing the 'optional' value for Questions will update the nodes' 'completed' and 'progress' values, as well as lock and re-lock nodes that rely upon the completion of the Activity Node

## Change Details
- Combined `updateNodeProgress()` and `completeNode()` into `updateNodeProgressAndCompletion()`
- removed `unlockNodes()` as `updateLockedStatus()` seems to have the same function
## Screenshot
https://user-images.githubusercontent.com/34692466/131008593-949b50cb-89fd-450c-baef-07e72e9ecfeb.mp4
## Issue Linkage
Closes (issue #1083 )
## PR Dependency
N/A
## Automated Testing
- Question marked as optional will be considered a completed question when calculating Activity progress


EldadZZipori:
I think we should mark/unmark a question after an admin changes an option to be optional or not. Currently if I make a question optional and then required it will be marked as completed until I click on the the node. 
We can either mark or unmark every question when an admin changes a node or block them from editing. (functionality available in drag&drop.)

Other then that everything looks great! Functionality works as intended, code looks really clean and good.

odotadot:
> I think we should mark/unmark a question after an admin changes an option to be optional or not. Currently if I make a question optional and then required it will be marked as completed until I click on the the node.

Is there a way to do this...? The only way I can think of is to make some endpoints to save and update Activity node progress each time the user publishes a node. I tried to $emit an event to ActivityMedia each time the user publishes a node, but since ActivityMedia is unmounted, it will never notice. As of right now, the Activity node progress is updated each time it mounts and dismounts.

> We can either mark or unmark every question when an admin changes a node or block them from editing. (functionality available in drag&drop.)

Currently, a question is considered complete if it is optional or has a saved answer. If a question was optional before, but the user answered it, it will still be considered complete when it is changed to be required. Should question completion be reset after edits? Or should edits even be allowed?

@wynnset  How should we approach this?

wynnset:
> Currently, a question is considered complete if it is optional or has a saved answer. 
@odotadot from this statement, it sounds to me that we do not "save" a node as completed, but rather calculate it on the fly using the questions in the activity (if all of them are either completed OR optional). That means if an author changes the optional flag for a question, the completion flag for the activity node would automatically be changed. Is that right? If that's the case, then we do not need to do what @EldadZZipori is suggesting.

If that is not the case, I suggest we change it to be the calculated on the fly if it's not too hard. If it's going to take more than a couple of hours to change it, then we can just forego this use case and open a task in the Dev Backlog on Asana and address it in the future.

Thanks

odotadot:
Seeing as the Activity node completion is already calculated on the fly, I believe this PR meets the specifications. The concern was about the timing the Activity node progress bar updates - it only updates when its lightbox opens and closes.

odotadot:
> Seeing as the Activity node completion is already calculated on the fly, I believe this PR meets the specifications. The concern was about the timing of the Activity node progress bar updates - it only updates when its lightbox opens and closes.

odotadot:
The completion flag and progress values for an Activity node now change based on on-the-fly calculations made after any edits. The progress bar for the Activity node only changes when its lightbox is opened or closed, not immediately after it is published.

EldadZZipori:
I think we should put the optional switch under a new title, maybe "Additional question options" .
image

EldadZZipori (approved):
Code looks really good and clean

wynnset:
Could you please rebase this on top of 1086 or otherwise @EldadZZipori can rebase 1086 on top of this? That way we will avoid looming merge conflicts

wynnset (changes requested):
Code looks good! Found a bug with the locked nodes: If I have a single optional question in an activity and another node has a lock condition on that activity, it does not unlock unfortunatel (at least in multi-content).

odotadot:
As long as I click on the activity node tab in the multicontent lightbox first, the completion will update. Should it work without having to first click on the activity node tab?
https://user-images.githubusercontent.com/34692466/130296657-5d3fd631-4240-40d5-b081-4dad6992ec31.mp4

wynnset:
> As long as I click on the activity node tab in the multicontent lightbox first, the completion will update (this is the same issue with the activity progress unable to update until it's mounted first). Should it be working without having to first click on the activity node tab?

Looking at the first half of the video, I see that the other node was locked until you went in and pressed "Skip". It shouldn't be required for the user to do that - the other node should be unlocked from the beginning ideally. Is this really hard to implement?

wynnset:
I don't think my previous comment above has been addressed yet.

I also found another issue: I just tested this and it doesn't seem to be working. When I complete an activity with 1 required and 1 optional question, the progress is not updating to 100%. When I inspect the network, I see that when the required question is complete, the progress sent up is 0.5 - this is wrong and it should be 1 so that that whole node is marked as complete.

wynnset:
Let's hold on this while we work on higher priority issues. It seems like there is not a huge demand for this.

wynnset:
Ignore that last comment. This does need to get done as soon as possible.

wynnset:
As this PR had a lot of issues, I've implemented this task in PR #1121 with all the needed fixes.

PR #1091, opened 2021-07-27
1090 add ability to mark page chlid as half width

EldadZZipori:
# Task Description
Currently all nodes listed in a page are full width. We want to add the ability to show nodes in a 1/2 + 1/2 layout like in the attached mockup (ignore everything except the layout).

To do this, we will use the Bootstrap grid (do not custom create CSS). 

Add a new checkbox for page child nodes "Show this as half width" under the appearance tab Content section.

## Changes

- Convert page rows to use bootstrap vue
- Add option to make a page child 50% width

To achieve this goal the PageRows component assess where to insert an empty 50% column, a spacer, and to make a row 50%.
The spacer -
```
<div class="w-100"></div>
```
Insures the next node will be displayed in a new row.

See "Equal-width multi-line" [here](https://bootstrap-vue.org/docs/components/layout).
 
```
insertInstructions: [],
```
Is used to determine what additional tag is needed after every column.

## Conceptual design
image
image

## Issue Linkage
Closes (issue #1090 )
## PR Dependency
Depends on: PR #1047


templates/vue/src/components/modals/NodeModal/index.vue
```diff
@@ -432,6 +435,11 @@ export default {
     isMultiContentNodeChild() {
       return this.parent && this.parent.mediaType == "multi-content"
     },
+    isPageMultiConentNodeChild() {
+      return (
+        !!this.isMultiContentNodeChild && this?.parent?.presentationStyle === "page"
```
odotadot, line 440: reduce to `this.isMultiContentNodeChild && ...`

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageRows.vue
```diff
@@ -5,85 +5,102 @@
...
+                    style="color: white; margin-bottom: 24px;"
+                    @complete="updateProgress(row.node.id)"
+                    @load="handleLoad($refs.rowRefs[index])"
+                  />
+                  <p
+                    v-if="row.children.length > 0 && !areAllPopup(row.children)"
+                    style="color: white;"
+                  >
+                    https://bootstrap-vue.org/docs/components/layoutypeData.subAccordionText
+                    }}
```
odotadot, line 59: What is this URL for?

templates/vue/src/components/modals/NodeModal/forms/AppearanceForm.vue
```diff
@@ -136,6 +136,14 @@
           Crop video to fill window (not recommended)
         </b-form-radio>
       </b-form-group>
+      <b-form-group v-if="isPageChild">
+        <b-form-checkbox
+          v-model="node.halfWidth"
+          data-qa="page-child-node-behaviour-half-width"
+        >
+          Show this as half width
```
odotadot, line 144: "Show this node as half width" might sound clearer

templates/vue/src/components/modals/NodeModal/index.vue
```diff
@@ -68,7 +68,10 @@
             :active="tab === 'appearance'"
             @click="changeTab('appearance')"
           >
-            <appearance-form :node="node" />
+            <appearance-form
+              :node="node"
+              :is-page-child="isPageMultiConentNodeChild"
```
odotadot, line 73: Isn't the prop named `isPageChild` ? not kebab-case?

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageRows.vue
EldadZZipori: Oops, not on purpose

templates/vue/src/components/modals/NodeModal/index.vue
EldadZZipori: I believe vue/js knows how to do that transformation by itself because it is common to use kebab-case for tags but JS properties are written camelCase

odotadot:
This PR works very well. There's only 1 concern with half-width content: External links dimensions looks a bit weird:
image

Other than that, all other content can be seen clearly in half-width mode. Code looks good.

EldadZZipori:
@odotadot You are right the External link was a bit off. I replaced it to use bootstrap-vue. Can you give it a look again to see that it is ok?

odotadot:
Re-testing: 
Some new concerns I didn't notice before have shown up:
1.  This error appears in the console log
image
2. When opening the regular lightbox for Activity and Answer nodes, they take up half the screen, even if they have not been selected to be half-width:
image
3. Did the background colour for the lightboxes change to be white in this PR? I thought they were dark usually. 
4. The half-width Link content is a bit weird still, as the image is beneath the text:
image

templates/vue/src/components/modals/NodeModal/index.vue
EldadZZipori: Turns out this is needed, otherwise it might return null sometimes

odotadot:
@wynnset Should this PR also prepare for nested Page Menus in Page Menu nodes? 
image

wynnset:
> Should this PR also prepare for nested half-width Page Menus in Page Menu nodes?

Hi @odotadot - I'm not sure what you mean by "nested half-width Page Menus" or "Page Menu nodes" because the page menu is not a node and also it doesn't appear as half width. Only the page nodes would appear as half-width. The page menu for a page multi-content node would appear the same way as it does currently - no changes to the page menu should be required unless you have a specific concern that I'm not getting.

wynnset:
@odotadot @EldadZZipori these changes should be on master and then we can pull them into tyde2. Please rebase this branch to master (make sure not to take anything from tyde2 into master).

EldadZZipori:
This PR is superseded by #1099

EldadZZipori:
This PR is superseded by #1099

PR #1072, opened 2021-07-02
1067 - TYDE2 - migrate community view onboarding

odotadot:
## Changes
- Manually migrated Community View Onboarding Files from branch 992 to this branch
## Screenshot
image
## Issue Linkage
Closes #1067
Closes #992
## PR Dependency
Depends on: #1048
Depends on: #993
## Automated Testing
n/a

PR #1095, opened 2021-08-03
1094 - TYDE CoS - Circle view onboarding

odotadot:
## Changes
- Creating circle view onboarding for the Circle of Support
## Screenshot
TBD
## Issue Linkage
Closes (issue #1094 )
## PR Dependency
Depends on: (PR #1072 )
## Automated Testing
N/A


templates/vue/src/components/tyde/activities/CircleOfSupport/CosModal.vue
```diff
@@ -0,0 +1,52 @@
...
+}
+.content {
+  position: relative;
+  padding: 2rem;
+  background: white;
+  border: var(--cos-border);
+  border-radius: 1rem;
+  max-width: 700px;
+}
+</style>
```
EldadZZipori, line 52: Don't we have the exact same component under CommunityView? I think you can delete the on under CommunityView and only use this one, because it is relevant to more then one component.

templates/vue/src/components/tyde/activities/CircleOfSupport/onboarding/TopTooltip.vue
```diff
@@ -0,0 +1,123 @@
...
+
+.right::before,
+.right::after {
+  right: 25px;
+}
+.left::before,
+.left::after {
+  left: 25px;
+}
+</style>
```
EldadZZipori, line 123: This seems almost identical to Tooltip. Can we combine them to one component with a prop that determine their position?

odotadot:
Initialization was changed: 
1. If no communities, start community view onboarding
2. If no connections, start connections onboarding and continue to circle view onboarding
3. If no connections in circle view, start circle view onboarding

As such, if the user deletes their connections (so the connections tab is empty), they will have to go through onboarding from (2) again. 

Failing test is unrelated to this PR.

EldadZZipori:
I think continue is good because this looks a bit long compared to the other buttons. 
image

This does not get positioned perfectly every time. 
image

When I do not have connections in circle view the onboarding does not start, even if I have communities and connections.
image

I couldn't check the whole onboarding because I am unable to continue to the COS onboarding with the subscriber user or the public user.

EldadZZipori:
This tooltip is still not in the correct position for me
image

Is there any difference between these two for the user? 
image


Other than that everything works for me.

wynnset:
The order of this onboarding needs an update:
1. "Click here to toggle to the circle view" (this is fine as-is)
2. "Try adding some connections into your circle" should appear right after switching to the circle view. The text should be changed to read "Let's add some connections into your circle. Click on this button to start."
3. Once the button is pressed to pull up the connections, that's when we should show the drag and drop gif and instruct them to drag their connections. The text in this one is fine as-is (I might make some adjustments after)
4. After they add their connections to the circle and press the down button to collapse the connection panel, we should show them a message saying "Great job! You have added your connections in your circle now." then have a "Next" button. 
5. Once the next button is pressed, show another message "You can always move connections by dragging them between circles. You can remove connections from your circle by dragging them outside of the circles." - we should NOT show them how to delete a connection (using the "Delete connection" button) here as that might come across like it will just remove them from the circles whereas it's actually deleting the connection altogether. Press next to continue.
6. Lastly, show them the circle gif and change the text to "To view connections in only one of the circles, click on this circle to toggle between the circles and keep going to view all of them again." Press next to continue.
7. The last popup should say "Great job! Don't forget to keep your circle of support updated. You can always come back here to add connections, communities, and update your circles."

Also, we need to update the GIFs. Currently, they are showing an outdated version of the app and also there are these black backgrounds that should not be appearing. We should make sure to use the Twenty Nineteen Wordpress theme to avoid such issues.

templates/vue/src/components/tyde/activities/CircleOfSupport/CircleView/index.vue
```diff
@@ -5,15 +5,19 @@
       class="tab"
       :connections="connections"
       :communities="communities"
+      :toolTipPositioned="toolTipPositioned"
       :draggable="!dragDisabled"
       @back="handleBack"
-      @add-connection="$emit('add-connection', $event)"
+      @add-connection="handleConnectionOpen"
+      @connection-opened="handleConnectionOpened"
```
wynnset, line 12: "handleConnectionOpen" and "handleConnectionOpened" sound very similar - it's preferable to keep these names more distinct. Couldn't the add connection handler be called "handleAddConnection"? I think that's also more self-explanatory.

templates/vue/src/components/tyde/activities/CircleOfSupport/CircleView/index.vue
```diff
@@ -5,15 +5,19 @@
...
       @edit-connection="handleEditConnection"
       @add-community="$emit('add-community', $event)"
       @delete-connection="$emit('delete-connection', $event)"
       @drag:start="handleDragStart"
       @drag:move="handleDragMove"
       @drag:end="handleDragEnd"
+      @connection-submitted="$emit('connection-submitted')"
+      @connection-closed="handleConnectionClosed"
```
wynnset, line 20: Can we put all these event listeners beside each other? It's a bit messy to see some down here and the rest up there ^

templates/vue/src/components/tyde/activities/CircleOfSupport/onboarding/ObFinishView.vue
```diff
@@ -32,16 +32,24 @@
...
+        </b-button>
+        <b-button
+          v-else
+          pill
+          variant="secondary"
+          class="secondary mx-2"
+          @click="$emit('done')"
+        >
+          Done
+        </b-button>
```
wynnset, line 52: Since the siblings of these elements are `b-col`, it doesn't make sense that this is not wrapped in a `b-col`. It's incorrect to mix `b-col` with other elements like this.

templates/vue/src/components/tyde/activities/CircleOfSupport/onboarding/onboardingMachine.js
```diff
@@ -2,9 +2,16 @@ import { createMachine } from "xstate"
 
 export const OnboardingStates = {
   Welcome: "Welcome",
+  MoveConnections: "MoveConnections",
```
wynnset, line 5: I think it would look more clear if we order these in the way in which they appear

wynnset (changes requested):

templates/vue/src/components/tyde/activities/CircleOfSupport/CircleView/index.vue
EldadZZipori: yes, that one was there before I believe

PR #1096, opened 2021-08-03
1085 - Create theme control

EldadZZipori:
## Task description
User variables to keep track of our colour sets

Add our current light colours as the default

Add a new set of colours for the dark mode

Allow user to change to dark mode in the user setting menu (where we also have avatar settings). 

We should do some research to figure out what's the most proper way of doing this, keeping in mind that we use Bootstrap, Vue, Bootstrap-Vue, and SCSS which all give us many options for doing this.

## Changes

- Add a theme property to the store
- Add a user settings button and data structure (currently only includes theme)

## Screenshot
image
## Issue Linkage
Closes (issue #1085 )

## Documentation Tasks

- [ ] Add color use tutorial to notion

## PR Dependency
Depends on: (PR #1079 #1090) 
## Automated Testing
 - Check user settings are opening and able to change themes.


EldadZZipori:
This can be tested already. I am looking for a way to check the store using cypress .

Frozemint:
image
image

Lightbox backgrounds doesnt seem to be applying? Also, may I suggest adding icons for the theme toggle buttons? Check here:  
https://www.flaticon.com/free-icons/light-mode

image

EldadZZipori:
I am checking the failing test. It is not directly related to the functionality itself.

Frozemint:
I just did a pull and checked out the nodes etc, and it seems like text colours are still not changing colour when dark mode is on. Light mode works great however.

image
image
image

Frozemint:
It looks much better. There are still a number of bugged items tho:

The answer activity buttons still have light text in light mode
image

The drag and drop activity drop zone have the same colour as the background so it becomes invisible
image

The previous answer components doesnt have a distinct background colour so it looks like the text are floating 
image

The audio record button in dark mode is still black, so it becomes kind of invisible until you mouse over it. It's also impossible to tell the re-record button is disabled here.
image

Lastly, the page accordion nodes title are black on black so it's invisible
image

EldadZZipori:
Currently, the svg for drag&drop is not changing completely.
image

This is limited by the svg itself and cannot be controlled by CSS. @wynnset what should I do with this one?

wynnset:
I think it's fine since it has borders that would work on both dark and light backgrounds

Frozemint:
image
The external link node text using dark mode doesnt change colour. Otherwise it looks like most of the text-related bugs are gone.

Frozemint:
All the nodes looks good and the theme changes correctly now. 
Just one last thing: did you make changes to the audio recording buttons? When I hover over them they don't change colour, so it's hard to tell they're enabled or that they could be pressed. The colour of the buttons are also hard to distinguish from the background in both light and dark mode.
image
image

odotadot:
Testing notes:
1. WordPress content is dark text on dark background. A bit hard to see.
image
2. If a Multicontent node has a Page node as a child, the background is white. 
image
3. For Video, h5P, link and wordpress media types, if they are in a Page node, the titles seem close to the edge.
image

Other than these, this PR looks very good.

classes/class.tapestry-user-progress.php
```diff
@@ -120,6 +120,30 @@ public function getH5PSettings()
         return $this->_getUserH5PSettings();
     }
 
+        /**
```
odotadot, line 123: align comment

interfaces/interface.tapestry-user-progress.php
```diff
@@ -56,6 +56,24 @@ public function updateH5PSettings($h5pSettingsData);
      */
     public function getH5PSettings();
 
+        /**
```
odotadot, line 59: align comment

templates/vue/src/App.vue
```diff
@@ -61,6 +63,12 @@ export default {
     },
   },
   mounted() {
+    client.getTheme().then(res => {
```
odotadot, line 66: perhaps you can put this into the `data `variable on line 84 and the following code inside that `then` statement as well

PR #1101, opened 2021-08-05
1100 - Bug Fix: Fix page navigation in full-screen mode

### LINKED ISSUE #1100: Fix page navigation in full-screen mode

Frozemint:
## Changes
This PR fixes the issue where selecting a row from the page menu while in full-screen mode, the row will not be selected. 
The issue was caused by the page menu only causing a page scroll, so the selected node wasn't actually updated in the router. The issue was hence fixed by doing a push in the router.

## Screenshot
N/A
## Issue Linkage
Closes #1100 
## PR Dependency
Depends on: N/A
## Automated Testing
Minor Bug Fix so N/A


odotadot:
Page navigation in full-screen mode now works properly. Code looks clean and minimal as well.

wynnset (approved):

PR #1104, opened 2021-08-10
1103 editable

Frozemint:
## Check your PR title (you can delete this section)
* Ensure it's exactly the same as the title on Asana and Github issues
* Starts with an issue number and a dash. E.g. `583 - Add tapestry zoom feature`
* If it's a bug, add "Bug fix:" after the issue number. E.g. `583 - Bug fix: Fix tapestry not zooming`
## Changes
* Write down exactly what changes this PR makes to the base branch. This will be used as the commit message once your PR is merged in. Note that all of your commit messages will be deleted when merging this PR. Please delete this line
## Screenshot
* This is a space to include a screenshot or GIF if this is a frontend change. Delete this section if this is not a frontend change.
## Issue Linkage
Closes (issue #)
## PR Dependency
Depends on: (PR # or N/A)
## Automated Testing
* Please enter what automated testing you have added so that testers will know what they should NOT test

### DIFF
--- a/templates/vue/src/components/modals/NodeModal/forms/ContentForm/ActivityForm/index.vue
+++ b/templates/vue/src/components/modals/NodeModal/forms/ContentForm/ActivityForm/index.vue
@@ -33,12 +33,12 @@
         </b-row>
         <b-collapse :id="`question-${index}-container`" visible>
           <b-card
-            sub-title="Show answer to a previous activity first"
+            sub-title="Question properties"
             bg-variant="light"
             text-variant="dark"
             class="mb-3"
           >
-            <b-form-group>
+            <b-form-group label="Show answer to a previous activity first">
               <b-form-checkbox
                 v-model="question.followUp.enabled"
                 switch
@@ -74,6 +74,11 @@
                   </p>
                 </template>
               </combobox>
+            </b-form-group>
+            <b-form-group label="Allow user to edit this answer">
+              <b-form-checkbox switch v-model="question.followUp.allowAnswerEdit">
+                {{ question.followUp.allowAnswerEdit ? "Yes" : "No" }}
+              </b-form-checkbox>
             </b-form-group>
           </b-card>
           <b-card
@@ -269,6 +274,7 @@ const defaultQuestion = {
     text: "",
     nodeId: null,
     questionId: null,
+    allowAnswerEdit: false,
   },
   answerTypes: {
     text: {
--- a/templates/vue/package-lock.json
+++ b/templates/vue/package-lock.json
@@ -11523,12 +11523,6 @@
             "psl": "^1.1.28",
             "punycode": "^2.1.1"
           }
-        },
-        "ws": {
-          "version": "7.3.1",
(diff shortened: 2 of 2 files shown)


EldadZZipori:
Functionality works for questions that where answered before, but answered question are not accessible if we check that the user can not edit them.
ezgif com-gif-maker

Frozemint:
Fixed. I incorrectly assumed that by default, we do not show the summary screen. I forgot that I have long changed that, since we should be always trying to show the summary screen when possible.

EldadZZipori:
It seems that 3 activities tests are failing due to the changes in this PR.

wynnset (changes requested):
I see that this was added for activity nodes, but it's supposed to be added only for answer nodes.

wynnset (changes requested):
Unfortunately this is still not implemented for the right node type. Again, the task says this is for an **_Answer_** node, but this is still implemented for the "previous answer" portion of an activity node question.
image

templates/vue/src/components/modals/NodeModal/forms/ContentForm/ActivityForm/index.vue
```diff
@@ -269,6 +269,7 @@ const defaultQuestion = {
     text: "",
     nodeId: null,
     questionId: null,
+    allowAnswerEdit: false,
```
wynnset, line 272: Unused data variable should be removed

templates/vue/src/components/modals/NodeModal/forms/ContentForm/ActivityForm/index.vue
```diff
@@ -33,12 +33,12 @@
...
         <b-collapse :id="`question-${index}-container`" visible>
           <b-card
-            sub-title="Show answer to a previous activity first"
+            sub-title="Question properties"
             bg-variant="light"
             text-variant="dark"
             class="mb-3"
           >
-            <b-form-group>
+            <b-form-group label="Show answer to a previous activity first">
```
wynnset, line 41: All changes in this file should be reverted as they are not necessary.

templates/vue/src/components/Lightbox/media/ActivityMedia/Question/index.vue
```diff
@@ -120,6 +144,22 @@ export default {
     isLoggedIn() {
       return wp.isLoggedIn()
     },
+    hasPreviousSubmission() {
+      return Object.keys(this.answers).length > 0
+    },
+    isEditable() {
+      return this.question.followUp.allowAnswerEdit
```
wynnset, line 151: We should not be storing this under the followup for a question as that is not what we are concerned with here. This is for ANSWER nodes and so it should be saved as a property of `typeData` (maybe called `isEditable`)

wynnset (changes requested):
This seems to be affecting BOTH answer nodes AND completed activity nodes. This checkbox should ONLY affect answer nodes and not activity nodes.

PR #1106, opened 2021-08-12
1102 - Bug Fix: Fix link uploaded videos dimensions

### LINKED ISSUE #1102: Fix link uploaded videos dimensions

Frozemint:
## Changes
This PR changes problematic CSS styles that are causing the buggy behaviour of the YouTube player, where the media control would be cut off. The cause of the bug was a CSS style in `templates/vue/src/components/modals/HelpModal.vue`, which had an `iframe` CSS rule that did not have a selector. Consequently the rule applied to _all_ iframes in the app indiscriminately, including the YouTube iframe and causing the buggy behaviour.

The issue was fixed by restricting the rule to only apply to iframes within the HelpModal. I also slightly tweaked the YouTubeMedia class so that the iframe is always the size of the parent div, which seems to be the intended behaviour. 
## Screenshot
image
## Issue Linkage
Closes #1102 
## PR Dependency
Depends on: N/A
## Automated Testing
This is a UI bug change, automated testing need not apply


odotadot:
The link uploaded video dimensions are now all good - the media bar is visible and the video is in clear view. Succinct solution.

EldadZZipori (approved):
Everything works on my end as well. fix is simple and logical.

templates/vue/src/components/Lightbox/media/VideoMedia/YouTubeMedia.vue
```diff
@@ -254,7 +254,6 @@ export default {
 
 <style lang="scss" scoped>
 .embed-responsive {
-  max-height: calc(100vh - 120px);
```
wynnset, line 257: Does this need to be removed? Same as the other change above... In the PR description it says that it was the help modal causing issues so I'm wondering if these changes are also part of the issue and that this fixes those issues without causing new problems.

wynnset (changes requested):

templates/vue/src/components/Lightbox/media/VideoMedia/YouTubeMedia.vue
Frozemint: The line was removed to avoid a possible second bug, where if the browser height is decreased, the iframe is also resized vertically: 
image
If this is intended behaviour, scoping the CSS rule seem to be the only thing necessary to fix the video player controls being cut off bug from my testing.

wynnset: Yes, this is intended. I will remove this code then.

wynnset:
This issue is already resolved in a different commit so I will close this PR.

PR #1110, opened 2021-08-18
1074 - Save the last selected node for user and select it on load (using router hook)

### LINKED ISSUE #1074: Save the last selected node for user and select it on load

wynnset:
## Changes
- When Tapestry loads, the selected node is the user's last selected node rather than the root node
- If the user is not logged in, the last selected node will be the root node

Note: This PR is a rewrite of #1076 that uses a router hook for saving the selected node
## Issue Linkage
Closes #1074
## PR Dependency
Depends on #1066 
## Automated Testing
N/A

### DIFF
--- a/classes/class.tapestry-user-progress.php
+++ b/classes/class.tapestry-user-progress.php
@@ -132,25 +132,25 @@ class TapestryUserProgress implements ITapestryUserProgress
     }
 
     /**
-     * This function checks if any user has answered a question
+     * This function checks if any user has answered a question.
      *
-     * @param string $postId 
-     * @param string $nodeMetaId 
-     * @param string $questionId 
-     * 
-     * @return boolean $hasAnswer
+     * @param string $postId
+     * @param string $nodeMetaId
+     * @param string $questionId
+     *
+     * @return bool $hasAnswer
      */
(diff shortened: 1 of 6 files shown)


wynnset:
Hi @odotadot I took the code you had and instead of saving the selected node in various components, I instead used a [navigation hook](https://router.vuejs.org/guide/advanced/navigation-guards.html#global-after-hooks). I also realized we do not use the Vuex selectedNodeId as we are using the router for determining the currently selected node centrally. Could you please review and test this and let me know if you find any issues since you are familiar with the task? I also thought it would be a good learning exercise for you. Thanks!

wynnset:
P.S. I have set the branch to the secondary menu branch as I think it's also touching a lot of the things that this is also touching and want to make sure we deal with any issues in advance.

odotadot:
@wynnset Tested this implementation and everything works the same as before. After some thought I think that whatever outstanding implementation that needs to be done to save the page node lightbox state for the user might be better started on some new PR based off of  the TYDE branch. I believe the `changeRow()` method which controls all the multicontent routes may need to be refactored to work with the navigation hooks. Let me know if there is anything else I should investigate or finish up as part of this PR.

PR #1116, opened 2021-08-30
1112 - Dev bug: Cypress CI only records 1 second of content

### LINKED ISSUE #1112: Dev bug: Cypress CI only records 1 second of content

Frozemint:
## Changes
* This PR bumps `Cypress` and [`@cypress/code-coverage`](https://github.com/cypress-io/code-coverage) (as required by Cypress) to the latest versions, 8.3.1, and 3.9.10 respectively. Updating Cypress seems to resolve the issue of video recording being 1 second long.

Some tests were broken from updating Cypress. In particular, the behaviour of `.should('not.be.visible')` [has been changed](https://docs.cypress.io/guides/references/migration-guide#Migrating-to-Cypress-6-0). Previously `.should('not.be.visible')` would pass if the element in question is either not visible _or does not exist in the DOM_. 

[This behaviour was deemed problematic by the Cypress team](https://github.com/cypress-io/cypress/issues/9348), as this behaviour leads to `should('not.be.visible')` being used "vaguely": it can be used for testing the nonexistence or invisibility of an element. This weakness of `should('not.be.visible')` could also lead to false negatives, in which tests incorrectly passes in cases such as typos.

The behaviour was thus changed such that `should('not.be.visible')` now fails when used on an element that was nonexistent. To test for nonexistence (elements that does not exist in the DOM, such as a dismissed modal popup) we are now required to use `should('not.exist')`; to test for invisibility (elements that **DOES EXIST** in the DOM and is just invisible to the user), we are now required to use `should('not.be.visible')`, and `not.be.visible` should be used in this manner only.

* Due to the above some now-outdated tests has been updated accordingly. 

A consequence of this new Cypress behaviour also means that dismissed modals/elements should be removed from the DOM entirely for them to pass. This means temporary workarounds that does not render modals/elements nonexistent will now fail tests.
## Issue Linkage
Closes #1112 
## PR Dependency
Depends on: N/A
## Automated Testing
* N/A, this is a package update


EldadZZipori:
Now that we use a newer version of cypress we have to also check that all the functionality work and replace outdated function.
Note that two tests are currently failing.

legendword:
The cypress videos are being correctly saved. Waiting for #1163 to be merged to see if all tests pass.

templates/vue/src/components/Lightbox/media/VideoMedia/UrlVideoMedia.vue
```diff
@@ -3,6 +3,8 @@
     <video
       ref="video"
       controls
+      muted
```
zifgu, line 6: Just wondering why we want to mute all videos. Could this end up being confusing for a user?

zifgu (commented):
Looks good to me. The tests are passing locally for me, and Cypress is saving the full videos.

I have left one comment, since the PR also mutes video nodes by default and I was just wondering why that is.

templates/vue/src/components/Lightbox/media/VideoMedia/UrlVideoMedia.vue
legendword: I'm not sure why the original author included this change either. I removed it and all tests are still passing. I think we should keep it removed.

wynnset (approved):

PR #1117, opened 2021-08-30
1114 - List activities in a tapestry and allow viewing all user answers

### LINKED ISSUE #1114: List activities in a tapestry and allow viewing all user answers

eabean:
## Changes
- Added UserAnswers button and modal to the toolbar. Only admin users can see and access it.
- Admins can drill down from activities to questions and display answers of the users who answered the question in tabular format for a single selected question.
- Can export all answers for a Tapestry (all activities, questions of all users) to .xlsx file (Excel). One sheet per activity within the workbook. 
- Adds [xlsx](https://www.npmjs.com/package/xlsx) to the package.json

Changes to routes:
- Settings modal is now at `/nodes/:nodeId/settings/:tab`
- Newly added UserAnswers modal is at `/nodes/:nodeId/user-answers/answers`

## Change Details

Created a new backend function `getAllUserAnswers()` that returns an object that has all the answers of all users for a Tapestry.
`allAnswers` is an object of objects like` { activityId: activityAnswers }`. `activityAnswers` is an object of objects like `{ questionID: userAnswers }`. `userAnswers` is an array of `userAnswers`, each `userAnswers` an object with these keys `{ ID, display_name, text, audio, multipleChoice, dragDrop }` with each value the answer that the user submitted for that answer type.

## Screenshot

image

https://user-images.githubusercontent.com/44929104/170800208-fbb16194-7f86-471c-b3f4-be0175ee20a8.mov

image

Old video (before frontend filtering):

https://user-images.githubusercontent.com/34692466/131918221-d70f1a98-7ed2-46c6-a7b5-a804d6e73be2.mp4

## Issue Linkage
Closes #1114
## PR Dependency
Depends on #1163
## Automated Testing
TBD


odotadot:
@wynnset I'm not sure how to format the drag and drop answers in the Excel file. Should this be included, and if so, what should it look like? An example of the Excel sheet is in one of the screenshots above.

wynnset:
> I'm not sure how to format the drag and drop answers in the Excel file. Should this be included, and if so, what should it look like? An example of the Excel sheet is in one of the screenshots above.
> 
You can just list the answers for each bucket as a list in the answer field. Eg:

- Bucket 1: Item 3, Item 4
- Bucket 2: Item 2
> 
> 
> Also, I am using [xlsx](https://www.npmjs.com/package/xlsx) to export to an Excel file. Is this fine? 

This should be ok. I was thinking CSV would be more simple but this is fine as well.

templates/vue/src/components/modals/UserAnswersModal/UserAnswersForm.vue
```diff
@@ -0,0 +1,250 @@
...
+      })
+      dragDropInfo.items.forEach(item => {
+        itemInfo[item.id] = item.text
+      })
+      answers.forEach(answer => {
+        if (answer.items.length > 0) {
+          let bucketKey = Object.keys(bucketInfo).find(
+            bucketId => bucketId === answer.bucketId
+          )
+          let bucketName = bucketInfo[bucketKey]
```
zifgu: Can we replace this with `bucketInfo[answer.bucketId]`?

templates/vue/src/components/modals/UserAnswersModal/index.vue
```diff
@@ -0,0 +1,85 @@
...
+  methods: {
+    closeModal() {
+      this.$emit("close")
+    },
+  },
+}
+</script>
+
+<style lang="scss" scoped>
+#save-button {
```
zifgu, line 72: I couldn't find an element with this id in the component, so I think this CSS can be deleted.

templates/vue/src/config/routes.js
```diff
@@ -57,6 +58,16 @@ export default [
...
+      {
+        path: `user-answers`,
+        redirect: `${ROOT_PATH}/user-answers/answers`,
+        children: [
+          {
+            path: `/:tab`,
+            name: names.USERANSWERS,
+          },
+        ],
+      },
```
zifgu, line 70: I would just like to make sure that these additions are necessary, since the User Answers modal seems to open at #/answers and not #/nodes/x/user-answers/answers.

zifgu (changes requested):
Confirming that the user answers feature works as described. Tested that only admin users have the UserAnswers button. Tested behaviour with:
- All 4 question types
- No activities (in which case clicking the export button does nothing), one activity, several activities in the tapestry
- Questions answered by no one, one user, and multiple users
- Also tested exporting to xlsx

In the case where no one has answered yet, the expected behaviour is that the answer column doesn't show up at all. I'm wondering if this could be confusing to new users and if we should leave an empty column instead when no one has answered.

image

classes/class.tapestry-node.php
```diff
@@ -424,6 +424,21 @@ public function addReview($comments)
         ];
     }
 
+    public function getMediaType($node)
```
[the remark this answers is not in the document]
legendword: Actually, these properties are private and thus cannot be directly accessed. They did this so that they can access the properties through a public function. It is a bit confusing, but it may be necessary in this case.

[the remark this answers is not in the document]
zifgu: That is rather unfortunate but I see why it's needed. Thanks for looking into this!

templates/vue/src/components/modals/UserAnswersModal/UserAnswersForm.vue
```diff
@@ -0,0 +1,250 @@
...
+        <b-col cols="12">
+          <div>
+            <b-form-group data-qa="select-activity" label="Activity">
+              <combobox
+                v-model="activityId"
+                :options="activityNodes"
+                data-qa="choose-activity"
+                item-text="title"
+                item-value="id"
+                empty-message="Please select an activity first"
```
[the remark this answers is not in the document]
legendword: Changed to "There are no activities in this Tapestry."

legendword:
> In the case where no one has answered yet, the expected behaviour is that the answer column doesn't show up at all. I'm wondering if this could be confusing to new users and if we should leave an empty column instead when no one has answered.

Good point. The problem is, however, that the column header is unknown when there is no answer to a question. The header is generated from the JSON object returned from server which would just contain ID and display_name if there are no answers.

templates/vue/src/config/routes.js
zifgu: Thank you for fixing this! To me, it looks good as-is. We might like to add more tabs in the future. Also, since the UI contains a tab, it seems nice that the route is consistent with that structure (similar to user-settings/theme, even though it only has one tab as well).

zifgu:
Update on the behaviour of the PR:
- If a user has not answered a question, we are now not showing the row of that user in the frontend.
- If there are no answers to a question (0 rows), show "question has not been answered by any user"
- However, all users will be included in the exported spreadsheet so that users can find which users have not answered yet.

zifgu:
Tested the new changes locally – everything looks good.

wynnset (approved):

PR #1118, opened 2021-08-31
1113 - Add units multi content presentation

EldadZZipori:
## Changes

- Add Units type to multi-content nodes
- Allow only page nodes to be added under units nodes

## Screenshot
image
## Issue Linkage
Closes (issue #1113 )
## PR Dependency
Depends on: (N/A)
## Automated Testing
TBA


templates/vue/src/components/Lightbox/media/MultiContentMedia/index.vue
```diff
@@ -121,7 +125,9 @@ export default {
     ]),
     ...mapState(["favourites", "rootId"]),
     rows() {
-      return this.node.childOrdering.map(id => {
+      let node = this.node
+
+      return node.childOrdering.map(id => {
```
wynnset, line 130: This seems unnecessary

templates/vue/src/components/Lightbox/media/MultiContentMedia/index.vue
```diff
@@ -310,7 +316,6 @@ button[disabled] {
 }
 
 .title {
-  color: #111;
```
wynnset, line 313: Is this related to this PR?

templates/vue/src/components/Lightbox/index.vue
```diff
@@ -297,6 +301,9 @@ export default {
         height: this.lightboxDimensions.height,
       }
     },
+    handleUnitChange(newNode) {
```
wynnset, line 304: Since this is the node ID, it's better to name it `nodeId` or `selectedNodeId` rather than `newNode`

wynnset:
Unfortunately this is not working for me. When I click to view the Units node called "Multi", instead of opening up my units node, it opens up one of its children. 
https://user-images.githubusercontent.com/8981258/133918879-e0f62471-9e26-4e44-bbb2-caeadb1840a1.MOV

templates/vue/src/components/TapestryApp.vue
```diff
@@ -50,7 +51,12 @@ export default {
   },
   mounted() {
     this.$root.$on("open-node", id => {
-      this.openNode(id)
+      let urlNodeId = id
+      const node = this.getNode(id)
+      if (node.presentationStyle === "units") {
+        urlNodeId = node.childOrdering[0]
+      }
+      this.openNode(urlNodeId)
```
wynnset, line 59: I think it would be more clear if you figure out and pass the correct node id of the actual page that you want to open when you emit the event rather than having to modify this file with this logic. I don't know why this component should get concerned with this logic.

templates/vue/src/components/modals/NodeModal/forms/ContentForm/index.vue
```diff
@@ -135,31 +136,47 @@ export default {
...
     }
   },
+
   computed: {
     ...mapGetters(["isMultiContentRow"]),
+    mediaTypes() {
+      if (
+        this.parent?.mediaType === "multi-content" &&
+        this.parent?.presentationStyle === "units"
+      ) {
```
wynnset, line 150:
```suggestion
      if (this.isUnitChild) {
```

templates/vue/src/components/modals/NodeModal/forms/ContentForm/index.vue
```diff
@@ -204,6 +227,15 @@ export default {
       }
       this.$emit("type-changed", evt)
     },
+    selectUnitChild() {
+      if (
+        this.parent?.mediaType === "multi-content" &&
+        this.parent?.presentationStyle === "units"
+      ) {
```
wynnset, line 234:
```suggestion
      if (this.isUnitChild) {
```

wynnset (changes requested):

wynnset:
@EldadZZipori it still wasn't working for me so I debugged and fixed it. Everything is working as expected now! Just need to do the code review a bit more thoroughly and this one should be good to get merged in.

wynnset (approved):
I made some visual updates, refactored a bit, and also changed the type to be called "unit" instead of "units" to be more clear. I've reviewed all the code and everything looks good.

wynnset:
@EldadZZipori sorry can you fix the test on this one? I broke it!

EldadZZipori:
Actually all the cypress tests are passing the one that does not pass is related to the preview link api? @wynnset

wynnset:
We need to fix the integration test, otherwise this is ready to get merged in.

EldadZZipori:
@wynnset integration test is now passing. I had to compute the variables in pagemenu in case they are undefiened.

PR #1120, opened 2021-09-07
1119 - Activities do not fit the lightbox on tyde2

EldadZZipori:
## Changes
Sets the media-wrapper div to have width:100%
## Issue Linkage
Closes (issue #1119 )
## PR Dependency
Depends on: (N/A)

### DIFF
--- a/templates/vue/src/components/Lightbox/media/TapestryMedia.vue
+++ b/templates/vue/src/components/Lightbox/media/TapestryMedia.vue
@@ -159,6 +159,7 @@ export default {
   border-radius: 15px;
   overflow: auto;
   height: 100%;
+  width: 100%;
   padding: 0;
 
   &-no-scroll {


wynnset:
@EldadZZipori is this working on master? Is the issue only with tyde2?

wynnset:
As we discussed, once done with higher priority issues, we should look and see how come this is not an issue on master even though the code is the same (or maybe the latest master is not pulled into tyde2)

EldadZZipori:
I looked locally and it seems that master is not merged into tyde2

wynnset:
Ok, so if we merge master into tyde2 this will be fixed? Have you tested it @EldadZZipori ? If so, we can close this PR.

EldadZZipori:
Yes right. On master, it is working. Closing this PR

PR #1125, opened 2021-09-19
1059 - TYDE2 - Fix aspect ratio for h5p videos

### LINKED ISSUE #1059: Fix aspect ratio for h5p videos

wynnset:
## Changes
* H5P stores the aspect ratio for some interactive videos. We want to use this to figure out the height of the video and its container as the current method has not proven very reliable
* If no aspect ratio is available, use a default aspect ratio of 16:9 (we should ideally let users define this default aspect ratio in the tapestry settings and also be able to override it per node)

Note: this is only relevant to h5p videos from youtube
## Screenshot
N/A
## Issue Linkage
Closes #1059
## PR Dependency
N/A
## Automated Testing
N/A

### DIFF
--- a/classes/class.tapestry.php
+++ b/classes/class.tapestry.php
@@ -654,6 +654,10 @@ class Tapestry implements ITapestry
                     $H5PIndex = array_search($H5PId, array_column($allH5Ps, 'id'));
                     if ($H5PIndex || 0 == $H5PIndex) {
                         $nodes[$i]->typeData->h5pMeta = $allH5Ps[$H5PIndex];
+                        // Could not get h5pMeta to populate with fresh data,
+                        // so had to create a new property here.
+                        // TODO: Figure out a way not to have to do this
+                        $nodes[$i]->typeData->h5pMeta2 = $allH5Ps[$H5PIndex];
                     }
                 }
             }
--- a/templates/vue/src/components/Lightbox/media/VideoMedia/H5PVideoMedia.vue
+++ b/templates/vue/src/components/Lightbox/media/VideoMedia/H5PVideoMedia.vue
@@ -314,7 +314,7 @@ export default {
       const aspectRatio = aspectRatioStr.split(":")
       this.frameHeight = (this.dimensions.width / aspectRatio[0]) * aspectRatio[1]
       // Add height for the video controls
-      this.frameHeight += 38
+      this.frameHeight += Math.min(38, this.frameHeight * 0.08)
       this.frameWidth = 0
 
       this.$emit("change:dimensions", { height: this.frameHeight })
--- a/templates/vue/src/components/Lightbox/media/VideoMedia/H5PVideoMedia.vue
+++ b/templates/vue/src/components/Lightbox/media/VideoMedia/H5PVideoMedia.vue
@@ -6,7 +6,7 @@
       'context-multi-content': hasMultiContentContext,
     }"
     :style="{
-      height: `${dimensions.height}px`,
+      height: `${frameHeight}px`,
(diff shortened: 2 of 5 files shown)


classes/class.tapestry.php
```diff
@@ -654,6 +654,10 @@ private function _addH5PMeta($nodes)
                     $H5PIndex = array_search($H5PId, array_column($allH5Ps, 'id'));
                     if ($H5PIndex || 0 == $H5PIndex) {
                         $nodes[$i]->typeData->h5pMeta = $allH5Ps[$H5PIndex];
+                        // Could not get h5pMeta to populate with fresh data,
+                        // so had to create a new property here.
+                        // TODO: Figure out a way not to have to do this
+                        $nodes[$i]->typeData->h5pMeta2 = $allH5Ps[$H5PIndex];
```
wynnset, line 660: Preferably we should fix this

templates/vue/src/components/Lightbox/media/VideoMedia/H5PVideoMedia.vue
```diff
@@ -131,8 +131,9 @@ export default {
         this.settingsLastUpdated = currTimestamp
       }
     },
-
     setFrameDimensions() {
+      // Switching to aspect ratio method for TYDE only
```
wynnset, line 135: The aspect ratio method should be something that we can do for all videos through a new Tapestry setting and then be able to override per video. We should probably add this feature for all videos that use an iframe (e.g. H5P, YouTube, Kaltura) so that the author can set the aspect ratio and that would cause this code here to not run.

templates/vue/src/components/Lightbox/media/VideoMedia/H5PVideoMedia.vue
```diff
@@ -297,25 +299,35 @@ export default {
...
 
-      this.frameHeight = this.dimensions.height
-      // Check to see whether this is an H5P recorder
-      // If it is, we can emit an event to load the recorded audio (if exists)
-      // and terminate
-      if (h5pInstance.recorder && loadedH5PId) {
-        this.loadedH5PRecorderId = loadedH5PId
-        this.h5pRecorderSaverIsLoaded()
-        return
-      }
```
wynnset, line 311: This is good - let's make sure this piece goes into master.

templates/vue/src/components/Lightbox/media/VideoMedia/H5PVideoMedia.vue
```diff
@@ -297,25 +299,35 @@ export default {
...
       const mediaProgress = this.node.progress
+
+      // This is for TYDE only
+      // Note: This method hardcodes a default 16:9 aspect ratio which is not a good idea
+      // TODO: Allow users to define the aspect ratio or otherwise figure out a way
+      // to get it automatically. Currently it seems H5P has the right aspect ratio for
```
wynnset, line 309: Yes

wynnset (commented):
This is no longer needed for TYDE and also it's got some commented out code and hardcoded values that is not great. That said, I think the functionality here is useful and we should try and finish this work and integrate a more proper implementation of it into master that also allows authors to set aspect ratios for videos on a global (Tapestry-wide) and also per-video basis. We could also look into figuring out the aspect ratio of the video when the node is added, but that might be a bit more complicated to do.

This is not urgent at the moment, so I'll mark it deferred and we can come back to it when we have more time.

wynnset:
Closing this as all issues with H5P video resizing have now been resolved in master and are getting merged into tyde2 soon as well.

PR #1135, opened 2021-10-06
1133 - Add edit link to lightbox and multi-content children and add breadcrumb for multi-content children

### LINKED ISSUE #1133: Add edit link to lightbox and multi-content
The link should appear beside the favourite button on the corner of lightbox and multi-content rows

### LINKED ISSUE #1142: Add breadcrumb to edit modal for children of multi-content

wynnset:
## Changes
Currently, there's no way to edit a piece of content other than to go to the tapestry and click on the small edit button on the node. This is not always the easiest way as users might want to edit things as they are viewing the content. This PR adds an edit link to the corner of the lightbox and also for each multi-content child node. This link takes the user to the edit modal for the node they clicked edit for. Once done, it sends the user back to view the content.

We also want to add a breadcrumb to show at the top of the edit node modal for multi-content so that the user can easily see the hierarchy for the node they are viewing and be able to edit the ancestors. Users should be able to edit the ancestors if they are coming from the lightbox (the requirement above) and once they cancel or publish, they should be taken back to the lightbox.
## Screenshot
image
image

## Issue Linkage
Closes #1133
Closes #1142 
## PR Dependency
N/A
## Automated Testing
N/A

### DIFF
--- a/templates/vue/cypress/integration/lightbox/multicontent.spec.js
+++ b/templates/vue/cypress/integration/lightbox/multicontent.spec.js
@@ -127,224 +127,224 @@ describe("Multi-content", () => {
     })
   })
 
-  // describe("Accordion", () => {
-  //   it("should be able to make a node an accordion", () => {
-  //     cy.setup("@oneNode")
+  describe("Accordion", () => {
+    it("should be able to make a node an accordion", () => {
+      cy.setup("@oneNode")
 
-  //     cy.getSelectedNode().then(node => {
-  //       cy.openModal("edit", node.id)
-  //       cy.changeMediaType("multi-content")
-  //       cy.submitModal()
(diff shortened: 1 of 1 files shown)


EldadZZipori:
This is cool addition. There is one thing that might be a bit confusing. 
If you open a node under a multi-content node directly and click the new edit button you will see "back to [name of multi-content]"
image

when clicking on it you will be directed to the correct node, i.e. the child node that was originally opened. 
The only thing is the imprecise test.

EldadZZipori:
There was a problem in this PR where a statement was evidently an assignment of a variable so a test failed. Now the test is falling because of a problem that when you delete a child node of a multi-content node, both directly or from the multi-content node, all nodes disappear. 

@wynnset

PR #1141, opened 2021-10-13
1138 - Open text node links in new window

### LINKED ISSUE #1138: Links in text nodes should open in a new window

EldadZZipori:
## Changes
Adds a target="_blank" when adding a link to text node .
## Issue Linkage
Closes #1138
## PR Dependency
Depends on: (N/A)

### DIFF
--- a/templates/vue/src/components/Lightbox/media/TextMedia.vue
+++ b/templates/vue/src/components/Lightbox/media/TextMedia.vue
@@ -21,7 +21,13 @@ export default {
   },
   computed: {
     content() {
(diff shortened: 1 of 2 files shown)


EldadZZipori:
@wynnset Do we want to implement backward compatibility? If so we will have to compute all link tags to have the attribute target="_blank" but then the additional computation will always be there.

wynnset:
@EldadZZipori yes we need to provide backwards compatibility, so the solution would be to preprocess the HTML before parsing it so that we add `target="blank"` to all the links in the HTML.

PR #1147, opened 2021-10-27
1146 - Show completion screen only when all question are done

### LINKED ISSUE #1146: Do not show default confirmation screen when moving from one question to the next
If a user is doing an activity with multiple questions in it, when they submit a question, we always show them the confirmation screen with a button to continue to the next question. This can get quite annoying if the confirmation screen is always just saying "Thanks" and it also disrupts the flow for the user. For that reason, if a question doesn't have a customized confirmation screen, when they submit it we should take them directly to the next question. If it's the last question being submitted, we can show them the confirmation screen. 

Once the user clicks the continue button on the confirmation screen, we should take them back to the answer of the first question (currently it shows the answer to the last question)

In addition, whenever a question is submitted, we should immediately scroll the user back up to the top of the activity in multi-content context. Sometimes a question is long and when you submit it, it stays at the same scroll position on the page (which could be a content below) which is very confusing for the user.

EldadZZipori:
## Changes
* Checking whether the question is last before showing completion screen
* Scrolling the user back to the top of the section in MultiContent view whenever necessary
* (In MultiContent view) Going back to the answer of the first question when the "Done"/"Continue" completion button is clicked, and there are no more unanswered questions after the current one.
* Adding a toggle for "customized/not customized" to the Confirmation customization section of the ActivityForm. If this toggle is on, display the form to customize the title and message, and count the confirmation screen as customized.

## Change Details
For backwards compatibility, customization is now considered "on" (and the toggle will be toggled on) if title or message strings are not empty.

## Screenshots
Custom confirmation toggle:

https://user-images.githubusercontent.com/44929104/170799443-f32dbdc7-730a-4d5f-932c-c1330d0a6fc0.mov

## Issue Linkage
Closes #1146 
## PR Dependency
Depends on: N/A


legendword:
@zifgu Try and see if you can replicate https://github.com/tapestry-tool/tapestry-wp/issues/1146#issuecomment-1133393108, if not we don't have to worry about it for now. Just test against the descriptions here: https://github.com/tapestry-tool/tapestry-wp/issues/1146#issue-1035491940

zifgu:
Confirming that only custom confirmation screens and the confirmation screen on the last question are shown.

The completion button sometimes goes back to the answer of the first question even when some questions are not completed – e.g. if the first question is not completed. I think this is because it only looks for unanswered questions after the active question. I might suggest either editing the description or the logic to match.

https://user-images.githubusercontent.com/44929104/170583893-bec24b9c-7849-4a03-9a05-cff0a235ada2.mov

Also, if you first set a custom confirmation title/message, then delete it, a confirmation screen still shows up even though the text is no longer customized.

https://user-images.githubusercontent.com/44929104/170584379-c9ce1428-a93a-49bc-a1b5-95703c355d56.mov

> @zifgu Try and see if you can replicate [#1146 (comment)](https://github.com/tapestry-tool/tapestry-wp/issues/1146#issuecomment-1133393108), if not we don't have to worry about it for now.

I was not able to replicate this issue when submitting the second-last question, or when editing the answer of the second-last question. Do you have any more information about the conditions where this issue occurs?

legendword:
Regarding your comment on deleted confirmation message:

I tested it on my end and deleting the confirmation title works, but deleting the message does not. I took a look into the actual stored content of that and it appears that the rich text editor retains an empty paragraph tag when you delete a line. This causes the code to detect that there actually is custom message. I've known this issue in almost every rich text editor... I guess this is why the code authors working on this PR before didn't include the check for non-empty confirmation messages. So the best solution I could come up with is simply not check for non-empty confirmation message, meaning we ASSUME that a custom confirmation message must always be accompanied by a custom confirmation title.

image

legendword:
> I was not able to replicate this issue when submitting the second-last question, or when editing the answer of the second-last question. Do you have any more information about the conditions where this issue occurs?

It was a very random thing that happened, I myself could not even replicate that issue, so I'd suggest not worrying about it too much until we see it consistently.

Lastly, returning to the answer of the first question is intended behavior according to Aidin, but obviously if you leave the first question unanswered it would be kind of weird. I am not sure what to do about this; personally I would say this situation is very rare, but I will let Aidin decide on this one.

zifgu:
> I'd suggest not worrying about it too much until we see it consistently.
> 
> Lastly, returning to the answer of the first question is intended behavior according to Aidin, but obviously if you leave the first question unanswered it would be kind of weird. I am not sure what to do about this; personally I would say this situation is very rare, but I will let Aidin decide on this one.

Sounds good. I also think the current behaviour isn't confusing, especially since the answer screen shows an informative message when the question is unanswered. In that case, I can update the PR description to match the behaviour of the completion button.

zifgu (commented):
The code changes look good to me and tests are passing. The deleted confirmation message issue is tricky, so maybe we can figure out what to do about it and then finalize this PR.

zifgu:
Update on the behaviour of the PR in the above edge case:
- Add a toggle for "customized/not customized" in the "Confirmation customization" section. If this toggle is on, we will display the form to customize the title and message, and count the confirmation screen as customized.

templates/vue/src/components/modals/NodeModal/forms/ContentForm/ActivityForm/index.vue
```diff
@@ -390,6 +409,15 @@ export default {
...
     },
+    setCustomConfirmation(question, enabled) {
+      if (enabled) {
+        question.confirmation.title = ""
+        question.confirmation.message = "<p></p>"
+      } else {
+        question.confirmation.title = ""
+        question.confirmation.message = ""
+      }
+    },
```
zifgu, line 420: It could be helpful to add a comment explaining why the custom confirmation toggle is implemented this particular way.

wynnset (changes requested):
Let's keep the style consistent, so the toggle should say either "Default" or "Customized" and it should be placed to the right side similar to the "Required" toggle button in the previous panel:
image

PR #1159, opened 2022-05-16
Fix link preview instructions to match main README

legendword:
## Changes
* Set link preview API key instructions in templates/vue to be the correct version as in the base dir README
* Update cypress-sample.json in README to reflect the current actual file contents
* Update links to MAMP/WAMP/LAMP setup guides
* Update linking of Github usernames to Slack for mentions to get forwarded there
## Issue Linkage
N/A
## PR Dependency
Depends on: N/A


templates/vue/README.md
```diff
@@ -46,20 +49,20 @@ Cypress is the framework we use for end-to-end (e2e) testing. For it to work, it
...
+  },
+  "baseUrl": "http://localhost:8000",
+  "retries": 1
 }
 ```
 
 Copy the file to `cypress.json` and change the variables to match your own settings.
 
-* `BASE_URL` is the url of your local site.
+* `baseUrl` is the url of your local site.
```
wynnset, line 65: I think this is the URL that Cypress uses or that the docker version uses (local site is on port 8888, I believe)

templates/vue/cypress-sample.json
```diff
@@ -5,7 +5,7 @@
     "SUBSCRIBER_USERNAME": "subscriber_name",
     "SUBSCRIBER_PASSWORD": "subscriber_password"
   },
-  "baseUrl": "http://localhost:8000",
```
wynnset, line 8: Is this change necessary?

legendword: It is for ease of setting up, because for me I had to change the port here to 8888 when copying this sample to my own config. Since the wordpress site has a default port of 8888, I think changing it in the cypress-sample.json would be appropriate.

PR #1161, opened 2022-05-16
1160 - Bug fix: Jest testing router import

### LINKED ISSUE #1160: Bug fix: Jest testing router import

legendword:
## Changes
- updated test.js to match new router.js exports
- bumped @testing-library/vue version to ensure the passing of a VueRouter instance as routes is supported
- ran php-cs-fixer to lint php code
- replaced node-sass with sass for GitHub actions to be able to run npm install
- fixed a calc() whitespace issue
- using a dedicated port (9000) for webpack bundle analyzer to avoid port conflicts with wordpress site

## Issue Linkage
Closes #1160

## PR Dependency
N/A


wynnset:
Hey @legendword I'm going to remove the dependency in this PR because it is being merged to master. You should only add the dependency the other way, otherwise you are creating a circular reference.

legendword:
> Hey @legendword I'm going to remove the dependency in this PR because it is being merged to master. You should only add the dependency the other way, not this way.

Alright, sorry about that!

zifgu (commented):
Confirming that all items in the changes have been implemented as intended, and all Jest integration tests are passing locally for me as well.

The end-to-end tests generally agree with my local results. For activity.spec.js, the tests seem to vary in whether they pass or fail, but since this is being addressed by #1163, I think everything looks good.

We did get a warning that our package-lock.json version is incompatible with the version of npm being used by CI (6.14.16). Do you know which version we should be using? Also would we need to update the npm version used by CI or regenerate the package-lock.json?

legendword:
> We did get a warning that our package-lock.json version is incompatible with the version of npm being used by CI (6.14.16). Do you know which version we should be using? Also would we need to update the npm version used by CI or regenerate the package-lock.json?

I'm using npm v8.10.0 locally which is the newest npm version for node v12. The version of npm used by the CI is probably similar to what I had before I updated npm. I think this warning can be safely ignored as it doesn't seem to have any consequences on the running of the tests. In the future, we could consider using a higher version of node in the CI so their npm becomes compatible with lockfile version 2. (there does not seem to be a way to specify an npm version for the CI to use, only a node version)

zifgu:
> (there does not seem to be a way to specify an npm version for the CI to use, only a node version)

I see, it does look like the version of Node used by the CI (v12.22.12) directly determines the npm version, so our only option would be to upgrade the Node version in the future. Just wondering, do you know if there is a reason we are using Node v12 for the project/anything that would break if we used a higher version of Node?

wynnset:
> do you know if there is a reason we are using Node v12 for the project/anything that would break if we used a higher version of Node

I'm not sure if there's a reason. I think we should try and upgrade it in a future task and see if there are any errors that we'd need to fix.

PR #1163, opened 2022-05-17
1162 - Fix Cypress tests

### LINKED ISSUE #1162: Fix Cypress tests

legendword:
## Changes
- no longer checking for Thanks! and Save Theme
## Issue Linkage
Closes #1162
## PR Dependency
Depends on: #1161

### DIFF
--- a/templates/vue/cypress/integration/lightbox/activity.spec.js
+++ b/templates/vue/cypress/integration/lightbox/activity.spec.js
@@ -47,7 +47,9 @@ describe("Activity", () => {
           .last()
           .trigger("drop", { dataTransfer })
         cy.contains(/submit/i).click()
-        cy.contains("Thanks!").should("be.visible")
+        cy.contains("You can press the button below to continue.").should(
+          "be.visible"
+        )
         cy.contains(/done/i).click()
       })
       cy.lightbox().should("not.exist")
@@ -151,7 +153,9 @@ describe("Activity", () => {
           .first()
           .trigger("drop", { dataTransfer })
         cy.contains(/submit/i).click()
-        cy.contains("Thanks!").should("be.visible")
+        cy.contains("You can press the button below to continue.").should(
+          "be.visible"
+        )
         cy.contains(/done/i).click()
       })
       cy.lightbox().should("not.exist")
@@ -228,7 +232,9 @@ describe("Activity", () => {
           .last()
           .trigger("drop", { dataTransfer })
         cy.contains(/submit/i).click()
-        cy.contains("Thanks!").should("be.visible")
+        cy.contains("You can press the button below to continue.").should(
+          "be.visible"
+        )
         cy.contains(/done/i).click()
       })
       cy.lightbox().should("not.exist")
@@ -271,7 +277,9 @@ describe("Activity", () => {
           .getByTestId(`multiple-choice-question-item-1-checked`)
           .click({ force: true })
         cy.contains(/submit/i).click()
-        cy.contains("Thanks!").should("be.visible")
+        cy.contains("You can press the button below to continue.").should(
+          "be.visible"
+        )
         cy.contains(/done/i).click()
       })
       cy.lightbox().should("not.exist")
@@ -308,7 +316,9 @@ describe("Activity", () => {
           .getByTestId(`multiple-choice-question-item-2-checked`)
           .click({ force: true })
         cy.contains(/submit/i).click()
-        cy.contains("Thanks!").should("be.visible")
+        cy.contains("You can press the button below to continue.").should(
+          "be.visible"
+        )
         cy.contains(/done/i).click()
       })
       cy.lightbox().should("not.exist")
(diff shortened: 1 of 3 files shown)


legendword:
Regarding the final failing test (multicontent.spec.js: should be able to lock accordion rows and progress through them):

The testing is performing as expected: the progression through accordion rows is indeed currently bugged. Viewing an accordion row does not unlock the next one. Upon primitive investigation, the cause seems to be that the nodes in the `rows` computed property in MultiContentRows.vue do not reflect the updated information from the `complete` event triggered by viewing a row. Specifically, the `completed` flag of the nodes in the `rows` computed property is not being updated to true when in the Vuex store it is already true (this is a bit of a speculation and I may be totally wrong). Potentially needs to open an issue and address the problem.

A similar issue is present in the locking of Tapestry nodes. A reload of the web page is required to unlock a node after its unlock condition has been met. Obviously, the node should be unlocked as soon as the Vuex store is updated. A fix should be implemented and after that, the test in node-locked.spec.js can be reverted back (removing the `.reload()` line)

templates/vue/cypress/integration/lightbox/video.spec.js
```diff
@@ -42,6 +44,10 @@ describe("Video", () => {
       })
       cy.closeLightbox()
 
+      cy.logout().visitTapestry()
+
+      cy.login(roles.SUBSCRIBER).visitTapestry()
```
zifgu, line 49: Wondering if there is a particular reason that this test needs to view the video as a subscriber. I tried commenting out line 49 and re-running this particular test, and it is still passing.

Also, if we are to keep line 49, we should probably remove the call to `visitTapestry()` after `logout()` so we avoid visiting twice.

legendword: I think viewing as a visitor works as well. I will make a commit for that.

zifgu (commented):
Confirming that everything in the Changes section has been implemented and looks good! I have left 1 comment on the item "view video as subscriber to make updateNodeProgress working" since I think it may not be needed.

Just to double check, do we need to access the multicontent node as a visitor because the unlocking behaviour does not work for admins? This would match what I observe locally.

zifgu (approved):
This looks good to me!

legendword:
> Confirming that everything in the Changes section has been implemented and looks good! I have left 1 comment on the item "view video as subscriber to make updateNodeProgress working" since I think it may not be needed.
> 
> Just to double check, do we need to access the multicontent node as a visitor because the unlocking behaviour does not work for admins? This would match what I observe locally.

Yes, the progress is not saved for the admin, so we need to view as someone else for progress to work.

zifgu:
> Yes, the progress is not saved for the admin, so we need to view as someone else for progress to work.

Got it, thank you for explaining. My bad, I just realized there was an unused import statement that I missed so we have one more commit here.

legendword:
Yes thank you for that, forgot to delete the roles since we are no longer using that constant.

templates/vue/src/components/Lightbox/media/MultiContentMedia/MultiContentRow.vue
```diff
@@ -1,30 +1,33 @@
...
-    <div v-if="isVisible" class="multi-content-row">
-      <div v-if="presentationStyle !== 'accordion'" class="title-row-icon">
+    <div class="multi-content-row">
+      <div v-if="presentationStyle === 'accordion'" class="button-row">
+        <button class="trigger-row-btn" :disabled="disabled" @click="toggle">
+          <i
+            v-if="!disabled"
+            :class="isVisible ? 'fas fa-minus' : 'fas fa-plus'"
+          ></i>
+          <i v-else class="fas fa-lock fa-sm title-row-ico"></i>
```
wynnset, line 10: @legendword I think the class `title-row-ico` here is misspelled? (missing an `n`).. I know it's not part of the changed code, but can we fix this or remove the class name otherwise if not needed?

legendword: I think it is being used, so I fixed the typo.

wynnset (commented):
The pencil and heart icons on the top-right corner of page rows are no longer having the margin they had on master, but they look correct on accordion rows:
image

wynnset (commented):
I am wondering why the changes in the `MultiContentRow.vue` file were made? I see that some of the logic for the elements have changed and just want to make sure this is a necessary change and that it's related to fixing the Cypress tests.

legendword:
> I am wondering why the changes in the `MultiContentRow.vue` file were made? I see that some of the logic for the elements have changed and just want to make sure this is a necessary change and that it's related to fixing the Cypress tests.

Before I made the changes, the button-row resided outside of the multi-content-row, which means the tests which were looking for the row title within .multi-content-row were failing. This restructuring basically makes .button-row and .row-content both descendants of the .multi-content-row, which works great both for the tests and the css styling.

wynnset:
> > I am wondering why the changes in the `MultiContentRow.vue` file were made? I see that some of the logic for the elements have changed and just want to make sure this is a necessary change and that it's related to fixing the Cypress tests.
> 
> Before I made the changes, the button-row resided outside of the multi-content-row, which means the tests which were looking for the row title within .multi-content-row were failing. This restructuring basically makes .button-row and .row-content both descendants of the .multi-content-row, which works great both for the tests and the css styling.

Sounds good @legendword . Thanks for the explanation. Please just fix the issue I mentioned above and we should be able to merge this in.

PR #1165, opened 2022-05-25
1164 - Upgrade node packages and prepare for node v16

legendword:
## Changes
**The changes require running `npm install` to take effect.**
**Now using node v16. Please update your local node version as well.**
- update webpack to v5 and webpack-related packages to their latest versions
- update babel eslint-parser
- update jest to v27 (not latest version) and jest-related packages to their latest versions
- use core-js instead of @babel/polyfill
- alter config params to work with newer package versions
## Issue Linkage
Closes #1164
## PR Dependency
Depends on: N/A
## Automated Testing
N/A

PR #1167, opened 2022-05-27
1166 - Adhere to vue/no-mutating-props

### LINKED ISSUE #1166: Adhere to vue/no-mutating-props
Removing the anti-pattern of directly mutating props across the code base.

https://eslint.vuejs.org/rules/no-mutating-props.html

legendword:
## Changes
* Store the node that is being edited by NodeModal in the Vuex store (`currentEditingNode`).
* Replace all mutations to the component prop `node` with a mutation that updates the property in the `currentEditingNode`.
* NodeModal no longer passes the node object to its child components. Each component that reads from / writes to the current editing node uses Vuex now.
## Issue Linkage
Closes #1166 
## PR Dependency
Depends on: #1165 (merged); #1185


legendword:
> Note to self: Investigate `Invalid prop: type check failed for prop "license". Expected Object, got String with value "".` for CopyrightForm

Actually caused by calls to `initialize()` in NodeModal. This issue is quite prominent as it gets triggered at inappropriate times where the node object "falls back" to a default one which has an empty string for "license".

legendword:
> > Note to self: Investigate `Invalid prop: type check failed for prop "license". Expected Object, got String with value "".` for CopyrightForm
> 
> Actually caused by calls to `initialize()` in NodeModal. This issue is quite prominent as it gets triggered at inappropriate times where the node object "falls back" to a default one which has an empty string for "license".

Implemented a small fix by removing the default empty string value of license in the default node (used by `initialize()`).

legendword:
It seems that we cannot add video popups at this moment:

The check for this button involves `this.actionType === "add" || node.mediaType !== "multi-content"` for checking whether the node needs to be saved first, but for video nodes this condition is always true.

image

templates/vue/src/components/modals/common/FileUpload.vue
```diff
@@ -230,30 +230,24 @@ export default {
   },
   data() {
     return {
+      // NOTE: compactMode cannot work without showing the image preview
+      showImagePreviewValue: this.compactMode || this.showImagePreview,
```
zifgu, line 234: When is `showImagePreviewValue` used in this component?

zifgu (changes requested):
Tested all affected components and confirmed the values are being updated properly.

templates/vue/src/components/Lightbox/media/ActivityMedia/Question/index.vue
```diff
@@ -138,7 +138,7 @@ export default {
               this.question.followUp.questionId
             ]
           ) {
-            // eslint-disable-next-line vue/no-side-effects-in-computed-properties
+            // eslint-disable-next-line vue/no-side-effects-in-computed-properties, vue/no-mutating-props
```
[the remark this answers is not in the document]
legendword: Yes. I think there is no easy way to make this adhere to the no-mutating-props rule. The question object should be a constant 99% of the time. This is the only exception, so I would say it's not worth changing lots of code just to fix this one thing.

templates/vue/src/components/modals/common/FileUpload.vue
legendword: Oops, forgot to update its references. Fixed now.

templates/vue/src/components/Lightbox/media/ActivityMedia/Question/index.vue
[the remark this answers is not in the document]
zifgu: Got it. Thanks for explaining, I will resolve this point then.

zifgu (approved):
The changes look good to me!

legendword:
Just need to review the new changes to NodeModal.vue in [this commit](https://github.com/tapestry-tool/tapestry-wp/pull/1167/commits/41c4ceac8cdb000a73d49aa78ce3b3175ee84234).

zifgu (commented):
Tested the mutations locally and confirming that the variables are being updated correctly through Vue Devtools.

Before I approve, doing a search for `this.node.` yields some cases where `this.node` seems to be modified directly.
- Do the assignments inside NodeModal `mounted` need to be changed as well?
- Do all the occurrences of the `this.node.mediaDuration = 0` need to be changed?

legendword:
> Tested the mutations locally and confirming that the variables are being updated correctly through Vue Devtools.
> 
> Before I approve, doing a search for `this.node.` yields some cases where `this.node` seems to be modified directly.
> 
> * Do the assignments inside NodeModal `mounted` need to be changed as well?
> * Do all the occurrences of the `this.node.mediaDuration = 0` need to be changed?

You are right. I missed those while making the changes. I also fixed the typo.

zifgu (approved):
Thanks for fixing that.

Tested that the thumbnail and locked thumbnail upload handlers in `mounted` still work. It looks done to me.

legendword:
The above commit fixes a bug that causes video nodes to never be completed. It was a bug discovered by @zifgu and since it deals with mutating the node, we put it here, too.

templates/vue/src/components/modals/common/FileUpload.vue
```diff
@@ -230,30 +230,24 @@ export default {
...
       uploadPercentage: 0,
       uploadBarInterval: null,
       uploadSource: null,
       isUploading: false,
       confirmedUpload: true,
       changeImage: false,
-      imagePatternId: "",
+      imagePatternId:
```
wynnset, line 241: We should make this one a computed value as well.

wynnset (changes requested):
This PR looks great! Just one minor change needed. Once done, please mark as ready to merge.

wynnset (approved):

PR #1172, opened 2022-06-01
1171 - Bug fix: Add video pop-ups in NodeModal

### LINKED ISSUE #1171: Bug fix: Add video pop-ups in NodeModal
It will always prompt "please save the node first" in a video node since `requiresSaving` is checking for node mediaType "multi-content", which prevents users from adding pop-ups directly in the NodeModal.

legendword:
## Changes
* Fix SubItemTable requiresSaving check so that the "Add New Pop-up node" button is working.
* Node created by clicking the "Add New Pop-up node" button has their popup toggle ON by default.

## Screenshot
image

## Issue Linkage
Closes #1171
## PR Dependency
Depends on: N/A
## Automated Testing
N/A


zifgu (approved):
Tested the fix when adding multicontent and video nodes, editing multicontent and video nodes, and changing the type to multicontent or to video. The fix behaves as intended. The code looks good to me as well.

wynnset (commented):
This looks good. Was just testing this and found another minor thing we could fix - can we check the "Show node as popup" checkbox automatically if user has clicked the "Add New Pop-up Node" button to get here? You could probably use a url parameter for that.
image

PR #1174, opened 2022-06-02
1173 - Create embed button

### LINKED ISSUE #1173: Create embed button
## Description

Create a button in the toolbar (with the embed symbol “</>”) that will allow users to embed Tapestries on other sites (similar to YouTube).

We should allow users to indicate what size they want the embedded iframe to be. There should be a default size (e.g. 800x600 or something that works).

The embeds should only include the tapestry and not the header, footer, and other parts of the webpage. We should add a small tapestry logo somewhere that is prominent for these embeds.

## Additional Notes

- The logo, name of the Tapestry, and link to view the Tapestry should appear below the iframe.
- The Wordpress Admin Bar should also be hidden.
- Logged in users should appear as not logged in (no authoring functionalities in iframe).

### LINKED ISSUE #1290: Tapestry sometimes shows large empty spaces in the tapestry when there are invisible nodes
When there are nodes in the tapestry that are available to the user, but invisible (e.g. are locked and hidden, or that are children of multi-content nodes), the tapestry creates space for them. This sometimes causes large empty margins in the tapestry.

In the example below, "Contact and Minorities" node is invisible to the user because it's locked and hidden.

User mode:

image

Admin mode:

image

legendword:
## Changes
* Add iframe mode, triggered by appending `?iframe` after page URL and before `#` (e.g. `http://localhost:8888/tapestry/test/?iframe#/nodes/905`)
* The admin bar, header, and footer all get hidden, and tapestry shows as not logged in in iframe mode
* Add Embed button in toolbar, and a modal to generate the code for the embed
* Provide users with the option to auto-set width & height of iframe
* Provide users with option to show a link to open the tapestry in a new window
* Bug fix: Tapestry sometimes shows large empty spaces in the tapestry when there are invisible nodes
## Screenshot
image
image
#1290 bug fix:
image

## Issue Linkage
Closes #1173 
Closes #1290 
## PR Dependency
Depends on: N/A
## Automated Testing
* Cypress e2e test: embed-modal.spec.js


templates/vue/src/components/modals/EmbedModal.vue
```diff
@@ -0,0 +1,101 @@
...
+    embed() {
+      return `<iframe width="${this.width}" height="${
+        this.height
+      }" style="border: none;" src="${this.settings.permalink}?iframe"></iframe>${
+        this.showInfo ? this.info : ""
+      }`
+    },
+    info() {
+      // ! using local path for Tapestry icon; change to hosted URL in production
+      return `<div style="margin-top:10px;display:flex;align-items:center;gap:10px;"><img height="40" width="40" src="http://localhost:8888/wp-content/plugins/tapestry-wp/templates/img/TapestryLogo.png" /><b>${this.settings.title}</b> <a href="${this.settings.permalink}" target="_blank">Open this Tapestry</a></div>`
```
wynnset, line 83: This path should not be hardcoded in. It should be generated using the real URI to the image. You can pass that in from tapestry.php, just like how we are passing in [vue-uri](https://github.com/tapestry-tool/tapestry-wp/blob/master/tapestry.php#L167).

templates/vue/src/components/modals/EmbedModal.vue
```diff
@@ -0,0 +1,101 @@
...
+    ...mapState(["settings"]),
+    embed() {
+      return `<iframe width="${this.width}" height="${
+        this.height
+      }" style="border: none;" src="${this.settings.permalink}?iframe"></iframe>${
+        this.showInfo ? this.info : ""
+      }`
+    },
+    info() {
+      // ! using local path for Tapestry icon; change to hosted URL in production
```
wynnset, line 82: You can remove this comment once you make this dynamic.

templates/vue/src/components/modals/EmbedModal.vue
```diff
@@ -0,0 +1,101 @@
...
+            <b-form-input v-model="height" type="number"></b-form-input>
+          </b-form-group>
+        </b-col>
+      </b-form-row>
+      <b-form-group>
+        <b-form-checkbox v-model="showInfo">
+          Show information below the iFrame
+        </b-form-checkbox>
+      </b-form-group>
+      <b-form-group
```
wynnset, line 29: I think there might be a way to also copy the code automatically to their clipboard. We should give them the option to click a button to copy to clipboard.

wynnset (changes requested):
This looks good overall, just a few changes needed as per comments, in addition to these:
- [ ] make showing or hiding the sidebar an embed option
- [x] reduce size of image to 80px x 80px

zifgu (approved):
The iframe mode and embed modal both work as described, and tests are passing locally. Tested the embed code by running another server, adding the embed code to an HTML file, and viewing the embedded Tapestry. This looks good to me!

templates/single-tapestry.php
```diff
@@ -8,8 +8,39 @@
...
+        }
+        <?php endif; ?>
+    </style>
+</head>
+<body <?php body_class(); ?>>
+<div id="page" class="site">
+<div id="content" class="site-content">
 
-get_header();
+<?php
```
wynnset, line 41: Rather than define the header content like this, it might be more eloquent to create an alternate header (maybe called iframe or frameless) and use the get_header function to call it. See the [Wordpress documenation](https://developer.wordpress.org/reference/functions/get_header/).

templates/single-tapestry.php
```diff
@@ -72,4 +103,17 @@
...
+</div><!-- #content -->
+</div><!-- #page -->
+<?php wp_footer(); ?>
+</body>
+</html>
+
+<?php
+else: get_footer();
+endif;
+?>
```
wynnset, line 119: Same as above, I believe you should be able to use the get_footer function to return an alternate footer.

templates/vue/src/components/modals/EmbedModal.vue
```diff
@@ -0,0 +1,137 @@
...
+        <b-form-checkbox v-model="showInfo" data-qa="show-info-toggle">
+          Show information below the iFrame
+        </b-form-checkbox>
+      </b-form-group>
+      <b-form-group label="iFrame Code">
+        <b-form-textarea
+          ref="code"
+          class="embed-code"
+          data-qa="embed-code"
+          readonly
```
wynnset, line 38:
```suggestion
          readonly
          rows="2"
          max-rows="10"
```

wynnset (changes requested):
Looks good. Just a couple of minor things. I'll probably also want to play with it myself after for minor adjustments.

templates/single-tapestry.php
legendword: The `get_header` function only looks at files in the **theme**, and after a bit of research, there is no way for it to use a custom header file in the **plugin**. Thus, I have changed to using the more low-level `load_template` which works similarly.

wynnset (approved):

wynnset:
I found an issue where if the Tapestry is imported from somewhere else, then the URL for the tapestry is undefined so the embed code doesn't work:
image

There might be a bigger underlying issue that's causing this. Maybe @zifgu could look into this as part of #652

zifgu:
> There might be a bigger underlying issue that's causing this. Maybe @zifgu could look into this as part of #652

It does look like a bug in the import functionality. During the tapestry export, the `settings->permalink` field is being removed, and is never set back when imported.

https://github.com/tapestry-tool/tapestry-wp/blob/e94c377e524e2869ca5102b32ab9ba4a00c1e1bf/classes/class.tapestry.php#L450

Since #652 is closed, I can fix the bug in #1197 if it's what you meant.

legendword:
image

The tapestry logo is using a hard-coded absolute path; since on my local server the tapestry plugin is not located under the directories used in the code, the logo is not displaying properly.

wynnset (approved):

PR #1177, opened 2022-06-06
1176 - Bug fix: Creating cycles with Multicontent causes Tapestry to freeze

### LINKED ISSUE #1176: Bug: Creating cycles with Multicontent causes Tapestry to freeze
**Steps to reproduce**:
- Create a Tapestry with a root node.
- Create a new node off the root.
- Create a new Multicontent node off this new node.
- Create a link between the Multicontent node and the root node. Click OK on all the dialogs that appear.
- Try to edit the Root node. At this point, the screen freezes for me and the Tapestry can no longer be interacted with at all.

This may not be the only way to trigger the bug. It may be related to cycles in the Tapestry and infinite looping. From preliminary exploration, it looks like the bug only happens when a multicontent node is involved in the cycle, but not 100% sure about this.

https://user-images.githubusercontent.com/44929104/172236644-11acb885-d0e1-4a7e-ac64-88aaed28f6a2.mp4

(It may not be visible in the video, but after 0:39, I am attempting to click.)

legendword:
## Changes
* Add check for parent mediaType
## Issue Linkage
Closes #1176 
## PR Dependency
Depends on: N/A
## Automated Testing
N/A

### DIFF
--- a/templates/vue/src/components/modals/NodeModal/index.vue
+++ b/templates/vue/src/components/modals/NodeModal/index.vue
@@ -452,7 +452,7 @@ export default {
       let parents = []
       let parentId
       let parent = this.parent
-      while (parent != null) {
+      while (parent != null && parent.mediaType == "multi-content") {
         parents.unshift(parent)
         parentId = this.getParent(parent.id)
         parent = parentId ? this.getNode(parentId) : null


zifgu (approved):
Tested the changes and the bug looks fixed. Tested with additional multicontent children, and the breadcrumb in the editing modal still works.

PR #1178, opened 2022-06-06
1175 - Bug fix: Changing video node to other type retains the media duration

### LINKED ISSUE #1175: Bug: Changing video node to other type retains the media duration
**Steps to reproduce**:
- Create a video node. When the node is created, it calculates the video duration, which is displayed below the node title.
- Then change the node to a different type. The bug seems to occur for all node types except H5P, possibly because H5P is processed as a kind of media and overwrites the `mediaDuration` field.

**Expected behaviour**: The `mediaDuration` field should be set to 0 and be hidden, since the node is no longer a video.

**Actual behaviour**: The `mediaDuration` field remains set and is still displayed.

https://user-images.githubusercontent.com/44929104/171795961-82e8be90-dd6e-487a-9922-f78adfe91b7c.mov

### LINKED ISSUE #1181: Bug: Changing from H5P to YouTube video does not trigger duration reload
**Steps to reproduce:**
1. Create a video node with a YouTube URL.
2. Change the node to an H5P node that is not an interactive video.
3. Change the node back to a video node, and provide the same YouTube URL as before.

**Expected behaviour:** The duration of the YouTube video should be displayed.
**Actual behaviour:** No duration is displayed.

https://user-images.githubusercontent.com/44929104/172493764-c86280e0-645f-462d-b12d-f7f7e6664367.mp4

legendword:
## Changes
* Unset mediaDuration when mediaType is no longer video/h5p.
* Trigger duration reload when mediaDuration does not exist, even when the link URL is the same.
## Issue Linkage
Closes #1175; Closes #1181
## PR Dependency
Depends on: #1167 
## Automated Testing
N/A


templates/vue/src/components/modals/NodeModal/index.vue
```diff
@@ -1099,6 +1099,15 @@ export default {
...
       }
+
+      if (
+        this.node.mediaDuration &&
+        this.node.mediaType !== "video" &&
+        this.node.mediaType !== "h5p"
+      ) {
+        this.node.mediaDuration = undefined
+      }
+
```
zifgu, line 1110: For clarity, would it be possible to separate this from the `validateNode` method? Since `validateNode` focuses on checking the node inputs, maybe the code to clear the duration could be put in `handleSubmit`, for example.

zifgu (changes requested):
Tested and the mediaDuration is being successfully unset when the node type is changed. 

However, I am currently encountering another bug where changing the node back to a video does not make the duration reappear. Does the bug reproduce for you?


https://user-images.githubusercontent.com/44929104/172272514-272149be-f6c2-4b60-b609-da2eec8e11f4.mov

legendword:
Using a different approach. Let me know what you think.

zifgu (approved):
Ideally, we would change the underlying mediaDuration field, but that would be much more difficult to keep consistent. I think this does make sense from a display perspective, so I will go ahead and approve it.

legendword:
The reason I did this is because when we change a node from video to some other type, the video URL information is still saved, so when we switch back to video that same URL will be there and it will not trigger the `shouldReloadDuration` method. This is why the mediaDuration stays undefined when we switch back to video. So I just implemented the simplest solution to fix this issue. In the future, we may consider triggering `shouldReloadDuration` when changing mediaType.

legendword:
Reverted the "other approach", and fixed (which also fixes bug #1181):

> Tested and the mediaDuration is being successfully unset when the node type is changed.
> 
> However, I am currently encountering another bug where changing the node back to a video does not make the duration reappear. Does the bug reproduce for you?
> 
>  bug-media-duration-2.mov

zifgu (changes requested):
The original bug looks fixed and the current code changes look good. However, changing from a YouTube video > H5P video > the same YouTube video still doesn't work as expected for me.


https://user-images.githubusercontent.com/44929104/172709484-b084adeb-6e0d-4129-b5c4-22496275b3d0.mp4



When the node is changed back to a YouTube video, `shouldReloadDuration` checks if the YouTube ID has changed. However, the old YouTube id is not unset in the background when the node type changed to H5P. One way to circumvent this might be checking whether the media type has changed.

legendword:
Implemented the automatic unsetting of mediaURL / youtubeID when changing node type.

zifgu (approved):
Tested switching to and from "video" type from all other node types. All bugs mentioned above are no longer occurring. I think this looks good.

wynnset (approved):

PR #1191, opened 2022-06-09
1190 - Tapestry 3.0 - Zoomable interface

### LINKED ISSUE #1190: Zoomable interface
This issue is related to the draft branch implementing the zoomable interface and gaining feedback.

### LINKED ISSUE #1263: Make zoom in/out and panning gestures more intuitive
Match zoom in/out and panning gestures to whatever Google Map does (unique to Windows vs. Mac)

legendword:
## Changes

*For testing: Edit node levels in the node modal, above the node title.*

* Drop shadow on nodes and lines.
* Darkened node color on lower levels.
* Zoom in using one of the following methods: Ctrl + Scroll (Windows & macOS), Cmd + Scroll (macOS), pinch scroll (macOS trackpad & mobile), double click.
* Nodes become further apart from each other while increasing in size when zooming in.
* Pan view by dragging on empty space. The user should not be able to pan too far away from the Tapestry --- the precise amount they can pan to each side is the border of the Tapestry nodes plus a margin of the radius of the currently largest node.
* Nodes are hidden based on scale and depth. Depth determines how many levels down the user can see. As the user zooms in, the "current level" will be changed and more nodes will be visible based on the depth. The nodes at the visible boundary is styled as a "grandchild" node, similar to how it used to show.
* The depth slider now functions to change the depth (how many levels down the user can see). The minimum depth is 1 and the maximum depth is (maxLevel - 1).
* Buttons on a node are hidden if node is too small (mainly for authoring).
* Click on a node to zoom in on that node such that the node is the default size (*default size = size of Level 1 node at initial scale) and also pan the view towards that node. If the user has manually zoomed in, then clicking on a node will zoom in further than the normal amount.
* Minimap is shown on the bottom right and can be toggled on / off. It shows the nodes (with their colors), links, and the current view position. Dragging the view box or clicking anywhere on the minimap will pan the view to the desired position.
* The minimap also displays depth cues including node size and shadow.
* (#1250) Adds a solid blue halo around the selected node.
  * The halo is "outside" the progress bar, but "inside" the outline from a draft/review node. The halo should fade in, and animate smoothly when the node radius changes, at the same rate as the node.
  * The halo has a width of 10% of the node radius.

**Bug fixes & other changes:**

* Fixed a bug where depth slider is hidden for a user that is not logged in.
* Removed event listeners on `$root` when component is about to be destroyed to avoid redundant event listeners that are added via `this.$root.$on`.

## Screenshot
image

## Issue Linkage
Closes #1190 
Closes #1263 
## PR Dependency
Depends on: #1220 
## Automated Testing
TBD


legendword:
Refer to the video below for the problems with [MDN's pinch zoom detection algorithm](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events/Pinch_zoom_gestures).

*First few seconds show that the algorithm works in their demo and also stopping the actual zoom at the same time.* However, this does not happen when I use the same algorithm in TapestryMain. As can be seen from the console logs, the pinch does not always get captured; and, it does not prevent the default behaviour of zoom on mobile.

Will implement an alternative using [touch events (not supported by IE / Safari)](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events#browser_compatibility) and [gesture events (ONLY supported by Safari)](https://developer.mozilla.org/en-US/docs/Web/API/GestureEvent#browser_compatibility).

https://user-images.githubusercontent.com/12668055/173462006-3a82a2dd-627b-44c4-bb92-71c87026a8f7.mp4

legendword:
Wheel zoom "hack" source: https://kenneth.io/post/detecting-multi-touch-trackpad-gestures-in-javascript

legendword:
Very strange: touch events are working on Safari mobile, yet on the support tables by MDN & caniuse, it is not supported. https://caniuse.com/?search=touchstart

legendword:
The failing test has to do with the old depth slider mechanism which was replaced by the new mechanism based on preset node levels. Not a priority but it needs to be updated. We may also need more tests for the zoomable interface once the specs are set in stone.

legendword:
Note: we can just revert [9adf44d](https://github.com/tapestry-tool/tapestry-wp/pull/1191/commits/9adf44d05dcc48edb4c993570b1beb61085fcf4b) to get rid of the developer settings for scaleConstants.

legendword:
Note: this PR should have its base branch changed to the PR branch of #1220 once #1203 is merged.

Bitajkr:
- [ ] The mini-map isn't always visible, such that I need to scroll down to see the mini-map. Whereas, the mini-map should be visible to the user during the entire time when they're navigating the tapestry. 

https://user-images.githubusercontent.com/113054791/189769224-188f1185-1014-4038-a4c0-5a457002f802.mov

Bitajkr:
- [ ] When I tried adding the 6th level (as seen in the image below), the node was positioned in a very random, far position. 

Screen Shot 2022-09-12 at 3 33 42 PM

Bitajkr:
- [ ] It's quite hard to see the progress bar for some of the deeper nodes as the colour for the progress bar and background is the same. 

Screen Shot 2022-09-12 at 3 56 09 PM

Bitajkr:
- [ ] My level 7 node has no shadows no matter how much I zoom in: https://test.tapestry-tool.com/1190-zoomable-interface/tapestry/testing-zoomable-interface-september-12-2022/#/nodes/16?scale=6.77&x=2677.5767&y=6487.0388

Screen Shot 2022-09-12 at 3 58 49 PM

PR #1193, opened 2022-06-15
1192 - Replace deprecated Cypress commands

zifgu:
## Changes
* Replaces deprecated `cy.server` and `cy.route` commands in the Cypress tests with `cy.intercept` (see [here](https://docs.cypress.io/guides/references/migration-guide#Migrating-cy-route-to-cy-intercept))
* Addresses some interceptions whose results are never used in the tests, by either deleting or `cy.wait`-ing on them

## Issue Linkage
Closes #1192
## PR Dependency
Depends on: N/A
## Automated Testing
N/A

PR #1195, opened 2022-06-21
1194 - Option to hide locked nodes

### LINKED ISSUE #1194: Option to hide locked nodes
# Description

When adding a condition to lock a node, add the option to also hide the node until it is unlocked. This would be done in the frontend (backend still returns the nodes in their locked form) for simplicity.

legendword:
## Changes
* Added a toggle in the Access tab in NodeModal to hide the node when locked.
* When this is enabled, the node will show an icon to indicate to the admin that it's hidden when locked, and its links will show up as half transparent to indicate that it will be hidden when in authoring view.
* When this is enabled, the node and its links will not show up at all for the viewer unless the node has been unlocked.
    * This includes showing up in the Tapestry, inside a Multi-Content node, inside the side menu of a Multi-Content page, and in the drop-down menu of a Multi-Content unit.
* This WILL result in some nodes being isolated (having no visible links) if the linked nodes are locked and hidden.

#### New Multi-Content display behaviours:
* If all pages of a Multi-Content unit are hidden, or if all children of a Multi-Content page / accordion are hidden, a text is displayed instead.
* If the first page(s) of a Multi-Content unit is/are hidden, the first non-hidden page will be shown instead; if the first non-hidden page happens to be locked, it will display a locked text describing the conditions of unlocking the page.

#### New Multi-Content page menu behaviours:
* The "Add navigation menu" option in the Appearances tab of Multi-Content pages will only control whether or not the page section navigation of the page menu is displayed (navigating between the children of the page).
* The unit navigation of the page menu will **always** be displayed when viewing a page in a Multi-Content unit, regardless of the option set for any specific page.
* The ability to toggle the page menu, therefore, depends on whether at least one of the two navigation features of the page menu should be displayed. It remains that units with only one page does not require the unit navigation.
* As a summary, the page menu component will not be displayed at all when:
    * There is only one page in the unit AND the page's show navigation menu option is OFF; or,
    * The page is not part of a unit AND the page's show navigation menu option is OFF.
    * In all other cases, the toggle for the page menu is visible and the user will be able to open the page menu, although the availability of each of the two types of navigation provided in the page menu depends on the aforementioned factors.
* The dropdown menu of the unit selection will now highlight the currently selected page.

#### Multi-Content Media navigation buttons:
* Added "Previous" and "Next" buttons below the Multi-Content media when browsing a page of a Multi-Content unit.
* The navigation buttons should not show for non-page Multi-Content, nested Multi-Content, and page that is not part of a unit.

## Screenshot
image

## Issue Linkage
Closes #1194 
## PR Dependency
Depends on: #1193 
## Automated Testing
* Added Cypress test to check the option for "hide node when locked" as administrator, view Tapestry as subscriber (node should be hidden), and unlocking as subscriber. [node-locked.spec.js]


zifgu:
Based on our discussion: Nodes that are children of multicontents and "hidden when locked" should not be shown inside the opened multicontent, or in the multicontent sidebar.

zifgu (commented):
Tested the behaviour of hidden nodes as regular children, and as children of multicontent accordions, pages, and units. Hidden nodes are completely hidden in the Tapestry, menus, and views. The code changes also look good to me.

I just added a sentence to the PR to more explicitly state all involved changes. Please feel free to add anything I missed!

Currently, children of a locked multi-content are styled regularly even if they are locked or hidden. Would it be a good idea to override this?

image

Also, I believe this was an edge case before this PR, but if the first Page under a Multicontent Unit is locked, it would still be opened to the locked page, thus bypassing the locking. Now that we've added hidden nodes, it is possible to open a Multicontent Unit and see a hidden page before it is unlocked. Is this something we'd want to address in the current PR?

wynnset:
> Currently, children of a locked multi-content are styled regularly even if they are locked or hidden. Would it be a good idea to override this?

I think it's a good idea to show the children as locked / hidden as well.

legendword:
@zifgu Please review the latest 2 commits involving the hide logic of multicontent children nodes. They should behave as expected now.

legendword:
@zifgu Please review the new changes as well as test them against the updated description at the top of this PR for how they should work.

templates/vue/src/store/getters.js
```diff
@@ -87,7 +87,10 @@ export function isVisible(state, { getNode, hasMultiContentAncestor }) {
       return false
     }
     if (!Helpers.hasPermission(node, "edit", showRejected)) {
-      return !hasMultiContentAncestor(node.id)
+      return (
+        (node.accessible || !node.hideWhenLocked) &&
```
zifgu, line 91: Similarly, would it be better to use `page.unlocked` here? I am not sure of the exact meaning of `accessible` vs `unlocked`, but this would increase consistency.

templates/vue/src/components/Lightbox/media/MultiContentMedia/index.vue
```diff
@@ -108,6 +149,20 @@ export default {
       "isMultiContent",
     ]),
     ...mapState(["favourites", "rootId"]),
+    parentNode() {
+      const parentNodeId = this.getParent(this.node.id)
+      return this.getNode(parentNodeId)
+    },
+    filteredPages() {
+      return this.pages
+        ? this.pages.filter(page => page.accessible || !page.hideWhenLocked)
```
zifgu, line 158: Since all the other checks of this form look like `node.unlocked || !node.hideWhenLocked`, would it be better to use `page.unlocked` here for consistency?

zifgu (changes requested):
All tests are passing locally. Tested the new Multi-Content display behaviours and navigation buttons. All is working:
- When all children of a multi-content accordion / page are hidden, a text is shown
- When a Multi-Content unit is opened, the first non-hidden (locked or visible) page is shown
- When "Add navigation bar" is checked, a menu is shown to navigate through the page's children. When unchecked, this menu is not shown.
- Multi-Content units with 1 child do not have the unit menu
- The currently selected page is highlighted in unit dropdown
- Previous and Next buttons appear only in Multi-Content units, and are disabled as appropriate.

The code looks good to me. I am still not too sure of the meaning of `accessible` vs `unlocked`... but based on consistency alone, I left some comments that may or may not be useful.

Lastly, when a Multi-Content has 0 children, it still says "Content in this node has not been unlocked yet." I think it may be good to show different messages for Multi-Contents that are truly empty vs empty because the content is still locked, to avoid confusion.

@legendword also discovered that the new page menu does not work in fullscreen mode, which would be good to fix too.
image

legendword:
I think it would be best to fix the full screen page menu bug in #1203 since that PR is based on this one and introduces lots of changes to the page menu component.

zifgu (approved):
> I think it would be best to fix the full screen page menu bug in #1203 since that PR is based on this one and introduces lots of changes to the page menu component.

Sounds good.

I tested the changes and the new text shows up for an empty Multi-Content unit. I will mark this as ready for final review.

templates/vue/src/components/Lightbox/media/MultiContentMedia/index.vue
```diff
@@ -17,17 +17,44 @@
...
-      :level="level"
-      @load="$emit('load')"
-      @change-row="changeRow"
-      @complete="complete"
-    />
+    <template v-if="node.presentationStyle">
+      <div v-if="node.presentationStyle === 'unit'">
+        {{
+          node.childOrdering.length
+            ? "The content of this unit has not been unlocked yet."
```
wynnset, line 24: Rather than this generic message, we should tell the user what they need to do to unlock the node (I think you can use the "locked-content" component for this).

templates/vue/src/components/modals/NodeModal/forms/ConditionsForm.vue
```diff
@@ -9,6 +9,11 @@
         logged in.
       </b-form-invalid-feedback>
     </b-form-group>
+    <b-form-group v-if="lock" class="mb-3">
+      <b-form-checkbox v-model="hideWhenLocked">
+        Also hide this node when locked
```
wynnset, line 14: Rather than a checkbox, can we have a toggle on the right side of the "Prevent access..." with 2 options: "Grey out" and "Hide"

templates/vue/src/components/Lightbox/media/MultiContentMedia/MultiContentRows.vue
```diff
@@ -1,6 +1,9 @@
 <template>
   <div data-qa="multi-content-rows" class="multi-content-rows">
-    <b-row :class="(node.presentationStyle === 'page' ? 'mr' : 'ml') + '-0'">
+    <div v-if="allHidden">
+      Content in this node has not been unlocked yet.
```
wynnset, line 4: Same here - let's please tell the user what they need to do to unlock this node.

templates/vue/src/components/Lightbox/media/MultiContentMedia/index.vue
```diff
@@ -17,17 +17,44 @@
...
+            : "There is no content in this node."
```
wynnset, line 25: Let's avoid using the word "node" for our user-facing (i.e. not authoring) side:
```suggestion
            : "There's currently no content here."
```

wynnset (changes requested):
Looks really good! Just a couple of comments and one other thing from my testing: Let's avoid using opacity for hidden nodes since it shows the link underneath and that doesn't look super polished. Instead, let's maybe just grey out those nodes (maybe a bit more than the regular locked nodes). 
image

Once done, please mark as "ready for merge" and assign back to me.

templates/vue/src/components/Lightbox/media/MultiContentMedia/index.vue
legendword: This text shows up only if the user opens an unlocked unit node with all its children locked and hidden (even if just one of the children is locked but not hidden, this text will not be shown and instead the locked-content prompt of that node will be shown). If we display the locked-content prompt, it will essentially let the user know that there are child nodes contained within the unit and let them see the conditions for locking the node, making one of the nodes "not hidden" when they should be. There are two issues here:

1. The behaviour would become the same as the "locked but not hidden" behaviour, since for locked nodes we will display the conditions. But for a hidden node, wouldn't it also be hidden when viewed from the parent, such that to the viewer the child would essentially not exist at all? So even just showing a locked-content prompt reveals some information that might not be what the author want when they decided to make the node hidden. (Also having all children of a multi-content node hidden while not hiding the multi-content node itself would be a strange decision on the author's side, so usually authors should just hide the multi-content node itself to achieve the desired result.)
2. If we do decide to use a locked-content prompt to let the user know what they need to do to unlock the node, which node's conditions should we display here? Since a unit can have multiple children that are all hidden, and each may have a different set of conditions, there may be some ambiguity as to which condition should be shown.

legendword: After discussion, we will use the text for the "has no content" situation when all children are hidden so that they behave as if there are no children attached to the multi-content node.

legendword:
When all children of a multi-content node are hidden, the user will perceive it the same way as when there are no children attached to a multi-content node, as per discussed. This ensures that hiding a node effectively removes the node from every  possible case of viewing the node and render it as if it does not exist.

wynnset:
Upon testing this, whenever I set the toggle in the Access tab to “Hide”, it shows as if it’s saved but when I refresh, it’s not saved. It also doesn’t seem to work on the user side. The data being loaded is consistent with this issue as there is no `hideWhenLocked` property under the node. This was tested on an existing tapestry by the way, but it should be backwards compatible.
image
image

wynnset (approved):

PR #1201, opened 2022-06-29
1200 - Bug fix: Empty H5P dropdown

### LINKED ISSUE #1200: Empty H5P dropdown
**Bug description**:

Although there is H5P content on WordPress, they are not detected in the H5P dropdown.

image

The error in the JavaScript console is:
image

This error was observed in production sites and recently reproduced locally. It may be caused by this line https://github.com/tapestry-tool/tapestry-wp/blob/062aa3195a8ad96d54bd9a520bab16523dbfca70/templates/vue/src/services/H5PApi.js#L23 when the `details` field is an empty string.

zifgu:
## Changes
* If the `details` field of the H5P is empty or otherwise invalid JSON, set `details` to an empty object by default rather than throwing an error.

**Testing**:
The issue seems to occur when at least one H5P has an empty `filtered` attribute in the `wp_h5p_contents` database. See the difference below:

image

To reproduce the bug, you could try manually deleting this attribute in the database, then see if this fixes it.

## Issue Linkage
Closes #1200 
## PR Dependency
Depends on: N/A
## Automated Testing
N/A

### DIFF
--- a/templates/vue/src/services/H5PApi.js
+++ b/templates/vue/src/services/H5PApi.js
@@ -16,11 +16,18 @@ export default {
 function parse(response) {
   if (response && response.length) {
     return response.map(({ id, title, library, details }) => {
+      let detailsObject = {}
+      try {
+        detailsObject = JSON.parse(details)
+      } catch (error) {
+        // ignore
+      }
+
       return {
         id,
         title: Helpers.decodeHTMLChars(title),
         library,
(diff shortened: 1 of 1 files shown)


legendword (approved):
Reproduced the bug, and after switching to this branch the bug no longer exists and I'm able to select H5P content. Looks good to me.

PR #1203, opened 2022-07-04
1202 - PageMenu redesign

### LINKED ISSUE #1202: PageMenu redesign

legendword:
## Changes
This PR aims to remove the padding still displayed by the PageMenu component while it is closed, so that content and interactive elements on the left edge of the Multi-Content page will not be covered by the closed PageMenu.

* Move page menu toggle button out of the page menu content, to allow the page menu content to be completely hidden when closed.
* Better transitions for opening the page menu (fade and slide in, text will no longer move around while transitioning).
* Fix the bug introduced in #1195 where full screen page menus are not shown correctly.
* More consistent styling of the logout link.

## Screenshot

During the transition, text slides smoothly into view along with the background:
image

Better logout link styles:
image

Can now interact with the content in the area highlighted in red, while the menu is closed (was not possible before this PR):
image

## Issue Linkage
Closes #1202 
## PR Dependency
Depends on: #1195 
## Automated Testing
* N/A


templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
```diff
@@ -217,103 +235,165 @@ export default {
...
     }
+  }
+}
 
-    @media screen and (min-width: 960px) {
-      font-size: calc(14px + (2 * (100vw - 960px) / 1280px - 960px));
+$slide-fade-speed: 0.3s;
+
+.slide-fade-enter-active {
+  white-space: nowrap;
```
zifgu, line 320: I think it may be better to remove this line (and the one in `.slide-fade-leave-active`). When the node title is too long to fit on the side menu, I am encountering a small stutter in the animation when the text re-wraps at the end of the animation.

With this line:


https://user-images.githubusercontent.com/44929104/177886742-3bb7eae9-7f48-41f0-befa-3e969eb86331.mov


Without this line:

https://user-images.githubusercontent.com/44929104/177886723-c3b3c480-5b3f-471e-ae13-0d6996358165.mov

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
```diff
@@ -217,103 +235,165 @@ export default {
...
+/*
+.slide-fade-enter-active,
+.slide-fade-leave-active {
+  transition: all 3s ease;
+}
+.slide-fade-enter,
+.slide-fade-leave-to {
+  min-width: 5px;
+  max-width: 5px;
+  opacity: 0;
```
zifgu, line 365: Could this commented code be removed, if safe?

zifgu (changes requested):
Tested the page menu, and the animation displays smoothly in both lightbox and fullscreen mode. The logout link is also much more visible now. I have just left some comments to do with refactoring.

Clicking an item in the page menu used to scroll to the corresponding row, but this behaviour is no longer working for me locally. If you are able to reproduce this issue, it might be good to fix.

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
legendword: On mine it does this thing while transitioning if without that line:

image

The text "Normal Unit" and "Body A" all fit in one line when opened, but since we're changing the width, they would not fit while transitioning. As to the situation you described, that is indeed problematic but it is a result of fixing this. I'm not sure how we can have both problems fixed....

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
```diff
@@ -217,103 +235,165 @@ export default {
...
+    }
 
-      &.closed {
-        background: transparent;
-        min-width: 20px;
-        max-width: 20px;
-      }
+    .page-nav-container {
+      text-align: left;
     }
```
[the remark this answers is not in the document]
legendword: Resolved: should be `.page-nav`

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
```diff
@@ -1,87 +1,97 @@
...
+            <h5 class="pl-2 py-1 mb-4">{{ node.title }}</h5>
+          </div>
+          <div
+            v-if="pageMenuVisible"
+            :class="[
+              'page-nav-content',
+              'mb-auto',
+              {
+                fullscreen: node.fullscreen,
+              },
```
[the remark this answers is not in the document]
legendword: You're right, thanks for noticing this!

templates/vue/src/components/Lightbox/media/MultiContentMedia/PageMenu/index.vue
legendword: Removed.

legendword:
We may want to discuss what approach to take for solving [this problem](https://github.com/tapestry-tool/tapestry-wp/pull/1203#discussion_r916344432), as I've noted fixing this will result in the text being wrapped during transition.

zifgu:
> We may want to discuss what approach to take for solving [this problem](https://github.com/tapestry-tool/tapestry-wp/pull/1203#discussion_r916344432), as I've noted fixing this will result in the text being wrapped during transition.

Could we animate the CSS transform instead, and have the sidebar translate in from the left so we don't have to worry about changing width? Or can we use different wrapping styles for the node title and the page menu items (wrap the node title, while not wrapping the rest)?

If those would be too complicated, we could go with the option that causes the least disruption visually. Without `nowrap`, the wrapping can get quite extreme while the side menu is expanding, so we could keep the current styles.

zifgu:
The new behaviour we should aim for:
- In lightbox mode, or in fullscreen mode with width <= 800px, the page menu should be collapsed by default and animate in from the side when opened.
- In fullscreen mode, if the screen width is greater than 800px, the page menu should always be visible and not animate.

legendword:
The new changes work great and the code looks good to me. Using `:css` to disable the transition is a smart choice.

I've brought back a bit of css that somehow got deleted. It is the new styling of the unit dropdown and it was introduced in #1195 so it was not supposed to be removed in this PR.

wynnset (approved):

PR #1208, opened 2022-07-11
1207 - Bug fix: Flaky test due to node covered by toolbar

### LINKED ISSUE #1207: Bug: flaky test due to node covered by toolbar
Refer to the tests of [this commit](https://github.com/tapestry-tool/tapestry-wp/pull/1199/commits/496a6f82018cbbb4e5c295f554fd06d354246934) for the flaky test.

legendword:
## Changes
* Hide the toolbar in the test for Answers node.

## Issue Linkage
Closes #1207 
## PR Dependency
Depends on: #1191 
## Automated Testing
* N/A

### DIFF
--- a/templates/vue/cypress/integration/lightbox/answer.spec.js
+++ b/templates/vue/cypress/integration/lightbox/answer.spec.js
@@ -53,11 +53,8 @@ describe("Answers", () => {
       cy.wait("@addNode")
     })
 
-    // hide the toolbar so the newly created node will never be blocked
-    cy.getByTestId("tapestry-toolbar").invoke("attr", "style", "display: none;")
-
     cy.getNodeByTitle("Answer Node").then(answer => {
-      cy.getNodeById(answer.id).click()
+      cy.getNodeById(answer.id).click({ force: true })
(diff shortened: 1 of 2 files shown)


zifgu (approved):
The test in question consistently passes locally and the solution makes sense. Everything looks good to me.

zifgu:
One thing to note: In other tests, `click({ force: true})` seems to be used to bypass the same problem. For example: 
```
    cy.getNodeByTitle(node.title).then(node => {
      cy.getNodeById(node.id).click({ force: true })
    })
```
in `review.spec.js` here: https://github.com/tapestry-tool/tapestry-wp/blob/5b27276e91146fcaedb7c92a99f25300f0469f36/templates/vue/cypress/integration/review.spec.js#L139

This solution works too, but I am wondering if it would be better to follow what the other tests do, for consistency.

zifgu (approved):
Everything looks good to me! The test still passes locally. I removed the `tapestry-toolbar` test ID for clarity - if it's needed in another PR, we can add it back.

PR #1210, opened 2022-07-12
1209 - Bug fix: Unable to scroll on mobile

### LINKED ISSUE #1209: Bug: Unable to scroll on mobile
The user will trigger dragSelect when performing a scroll when on touch screen. Disable dragSelect on mobile to fix the issue for now. In the zoomable interface, provide a dedicated tool for group selection so dragSelect can work on mobile and on the zoomable interface (along with the panning).

legendword:
## Changes
* Disable dragSelect on touch enabled devices so users can scroll the page normally.

## Screenshot
* N/A
## Issue Linkage
Closes #1209 
## PR Dependency
Depends on: N/A
## Automated Testing
* N/A

### DIFF
--- a/templates/vue/src/components/TapestryMain/index.vue
+++ b/templates/vue/src/components/TapestryMain/index.vue
@@ -13,7 +13,7 @@
           :target="nodes[link.target]"
         ></tapestry-link>
       </g>
-      <g v-if="dragSelectEnabled && dragSelectReady" class="nodes">
+      <g v-if="!dragSelectEnabled || dragSelectReady" class="nodes">
         <tapestry-node
           v-for="(node, id) in nodes"
           :key="id"
@@ -82,7 +82,7 @@ export default {
       return Number(this.$route.params.nodeId)
     },
     dragSelectEnabled() {
-      return !this.settings.renderMap
+      return !this.settings.renderMap && !Helpers.isTouchEnabledDevice()
     },
     editableNodes() {
       return this.nodes.length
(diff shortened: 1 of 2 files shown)


zifgu (approved):
This successfully disables drag select on my device, which supports touch. Everything looks good to me.

templates/vue/src/components/TapestryMain/index.vue
```diff
@@ -13,7 +13,7 @@
           :target="nodes[link.target]"
         ></tapestry-link>
       </g>
-      <g v-if="dragSelectEnabled && dragSelectReady" class="nodes">
+      <g v-if="!dragSelectEnabled || dragSelectReady" class="nodes">
```
wynnset, line 16: By negating the `dragSelectEnabled` check, I think you are also unintentionally causing the nodes to show up if `this.settings.renderMap` is true. i.e. `!!this.settings.renderMap === this.settings.renderMap`

I think this issue would be solved if you just add `!this.settings.renderMap` to this `v-if`.

```suggestion
      <g v-if="!this.settings.renderMap && (!dragSelectEnabled || dragSelectReady)" class="nodes">
```

wynnset (changes requested):

templates/vue/src/components/TapestryMain/index.vue
legendword: Actually, I found out that we are rendering TapestryMap instead of the TapestryMain component when `this.settings.renderMap` is true, so `this.settings.renderMap` is guaranteed to be false in TapestryMain. I will remove the check for `this.settings.renderMap` from TapestryMain instead, since it is not needed.

wynnset (approved):

PR #1212, opened 2022-07-13
1211 - Bug fix: nodes not being returned in incognito mode

### LINKED ISSUE #1211: Bug: nodes not being returned in incognito mode
**Steps to reproduce**:
- Create a Tapestry. Add a root node and other nodes.
- Reverse the link from one of the nodes to the root node, so that the root becomes the child.
- Open the Tapestry in Incognito mode.

**Expected behaviour**: All nodes in the Tapestry should be visible.
**Actual behaviour**: The entire subgraph connected to the root through the reversed link is not visible, and its data is not present in the Vuex state.

The cause is a bug in `_filterNodesMetaIdsByAccess` in class.tapestry.php. Only one (directed) traversal was initiated from the root, which excluded any nodes not reachable from the root from the list of nodes sent to the frontend.

zifgu:
## Changes
* `_filterNodesMetaIdsByAccess` now keeps track of visited nodes, and initiates traversals from each non-visited node, instead of just one traversal from the root.
* Passes the parameter `$filterUserId` to `_filterNodesMetaIdsByAccess`, which was previously missing. This has no effect on the titular bug, but does fix an existing issue with the code.

## Issue Linkage
Closes #1211 
## PR Dependency
Depends on: N/A
## Automated Testing
N/A

### DIFF
--- a/classes/class.tapestry.php
+++ b/classes/class.tapestry.php
@@ -643,7 +643,7 @@ class Tapestry implements ITapestry
 
     private function _filterTapestry($tapestry, $filterUserId)
     {
-        $tapestry->nodes = $this->_filterNodesMetaIdsByAccess();
+        $tapestry->nodes = $this->_filterNodesMetaIdsByAccess($filterUserId);
         $tapestry->links = $this->_filterLinksByNodeMetaIds($tapestry->links, $tapestry->nodes);
 
         return $tapestry;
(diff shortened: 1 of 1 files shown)


legendword (approved):
When on master, reversing the link to a node will cause that node and its children to disappear in incognito mode. When switched to this branch, this issue no longer exists. The code changes look good as well.

wynnset (approved):

PR #1225, opened 2022-07-29
1224 - Tapestry 3.0 - User-facing UI redesign

legendword:
## Changes
* Restrict app height as to not have an overflown body, thus achieving a fullscreen experience while keeping the Wordpress header and footer visible.
* Minimap size is now set to 20% of the Tapestry view size.
* Rename toolbar to menubar to differentiate between the bar on the top and the toolbar on the left which will be added in #1227.
* Display title of tapestry on the top left.
* Overlay the menubar and title on top of the main Tapestry view, so content can go below them.
* Move tapestry depth slider to the bottom right, beside the minimap. The depth slider is now vertical.
* Move tapestry filter button to the right, besides the menubar. The filter button and the menubar have separate backgrounds.
* Move view user answers button into the triple dot operations button in the menubar to save space.
* Hide the review notifications button in the menubar if there are no nodes awaiting review.
* Add the depth indicator beside the depth slider, showing layers of levels that are reactive to the user's current zoom level.
* Improve the positioning of the add node button, import button, and the import file dropbox when the tapestry is empty.
* Scrolling the page will zoom the tapestry, without the need to press any modifier keys.

## Screenshot
(outdated)
image

## Issue Linkage
Closes #1224 
## PR Dependency
Depends on: #1259 
## Automated Testing
* N/A

PR #1227, opened 2022-08-03
1226 - Tapestry 3.0 - Add general and contextual toolbars

### LINKED ISSUE #1226: Add toolbar and tool switching
As part of the Tapestry 3.0 redesign, there should be a vertical toolbar on the left, and users should be able to switch between tools by clicking on icons in the toolbar. Tools like Add Node and Select should later go in the toolbar.

zifgu:
### General toolbar:

* Adds a vertical toolbar to the left of the Tapestry view. Users can switch between tools by clicking on toolbar buttons.
* The user can pan the view when using any tool except the Select tool. The user can zoom the view at any time.
* Drag selection is now enabled by the Select tool, rather than enabled by default. In the Select tool, panning by dragging the Tapestry is disabled, but panning with the minimap should still work.
* The general toolbar contains the following tools (with their respective shortcut key bindings):
    * Pan tool (P): Drag anywhere on the background of the main view to pan the view. This is the default selected tool.
    * Select tool (V): Click on a node won't select the node but will instead show the node toolbar. Dragging will select all nodes that are in the drag area. The behaviour is exactly the same as before when panning wasn't introduced.
    * Add node tool (N): Shows a placeholder node below the cursor; clicking will create a new node at the cursor position with a default title "Untitled" and enter the in-place title editing mode for that node; after the title has been edited, the placeholder node shows up again for the user to potentially create more nodes.
      * If user does not have add permission for standalone nodes but draft nodes are enabled, this tool will add draft nodes instead.
    * Add connection tool (L): Click on a node to start adding a link. While selecting the target node, a placeholder link appears; when target node is clicked, the placeholder link freezes and the client makes the request to add the link. Once the link is added, the placeholder link is replaced with the actual link.
    * Settings: Opens the Tapestry settings modal.
    * Undo (Ctrl / Cmd + Z): Undoes the last action. Disabled if no available action to undo.
    * Redo (Ctrl / Cmd + Shift + Z): Redoes the last action. Disabled if no available action to redo.

### Contextual toolbar:

* The contextual toolbar will disappear if:
     * A different element with a contextual toolbar is clicked
     * The "background" is clicked
     * The element is no longer visible on the screen due to zooming out
* Clicking a node opens the **node toolbar** above the element if the user has edit permission on the node.
  * The node toolbar will appear *after* the view has zoomed to the node as a result of the click or keyboard navigation.
  * The node toolbar contains the following functionalities:
     * Delete
     * View node (only when `node.hideMedia` is true)
     * Change title text color (when set to "transparent", the title will be hidden)
     * Change thumbail image / color / remove thumbnail image (if applicable)
       * While this submenu is opened, drag & drop a file anywhere on the screen to upload the file as the thumbnail image.
     * Change node level (up / down)
     * Other node operations, including:
         * Node permissions (opens up a modal for editing the node permissions)
         * Lock node (opens up a modal for editing the conditions for unlocking the node)
* Clicking a link opens the **link toolbar** above the element if the user has edit permission on the link.
  * The link toolbar contains the following functionalities:
    * Delete
    * Display of the titles of the source and target nodes and option to reverse the link (only shows when the link is a same-level link).
  * Directionality of cross-level links are automatically determined by the levels of the connected nodes, so there is no option to set the direction manually for these links.
* When nodes are selected / de-selected by either clicking them with modifier keys or using drag select, the **multi-node toolbar** will appear at the center of all selected nodes if the user has edit permission on all of them.
  * The multi-node toolbar contains the following functionalities:
    * Delete the selected nodes
    * Change node levels (up / down) of the selected nodes
  * Introduces 3 new endpoints (batchUpdateNodes, batchDeleteNodes, batchRestoreNodes) to improve the experience when changing levels / deleting multiple nodes at once.

### In-place node title editing:

* Double click on the title of a node to enter the in-place title editing mode, which will immediately select the title text and allow directly editing it via keyboard.
* The cursor will be a text cursor when hovering over the node in in-place editing mode to indicate that the text is editable.
* The title will be saved when the title text element loses focus by any means (including interacting with other UI elements, pressing Enter, pressing Escape, etc.).
* No longer hide the title when the user hovers on a node with a thumbnail, if the user has edit permissions on that node; this is to allow the user to double click on the title.

### Other changes:

* Removed 3 of the colors in the color palette for the color picker in favor of white, black, and gray (the default gray color used as node background), to allow users quick access to these default color values.
* In the case when "transparent" color is allowed (such as the title text color of a node), the last color in the color palette is replaced with the transparent color.
* Removed the link modal. All actions on links can be performed in the link toolbar. Removed the relevant keyboard shortcuts to open the link modal as well.
* Upgraded the `dragselect` package.
* Added the `@popperjs/core` package for positioning the node and link toolbars.
* Increased the width on the parent side of links to make parent-child links more apparent.
* Reduced the duration of the click-to-zoom animation so the zoom effect is faster.
* Will pan view more towards the center if clicking on a node that is near the edges of the view (if node is within 60% of the center of view, this does not apply).
* Introduced the `--border-color` themed css variable and removed the unused `--button-color` themed css variable.
* Menubar buttons are now square-sized, and the menubar uses the same border as the toolbars.

### Bug fixes and improvements:

* Now using the "#ffffff" hex value for the default node title text color instead of the word "white" for better compatibility with the color picker.
* Always maintain the correct maxLevel value for all kinds of actions that may potentially change the value (addNode, editNode, deleteNode).

## Screenshot
Overview (with the general and node toolbar shown):
image

Link toolbar (same-level links):
image

Link toolbar(parent-child links):
image

Multi-node toolbar:
image

Hovering a file for drag & drop when node toolbar's background submenu is opened:
image

## Issue Linkage
Closes #1226

## PR Dependency
Depends on: #1225 

## Automated Testing
* Should be able to add a link using the link tool.
* Should be able to edit a node's title in-place by double clicking.


zifgu:
The current drag selection stops at the edge of the minimap and doesn't detect motion over the minimap. I believe this is because the mousemove event is not triggered on the part of the Tapestry underneath the minimap.

https://user-images.githubusercontent.com/44929104/182760197-ee3be8be-9462-4c16-9755-b5c2d8f06e53.mp4

This also affects panning, although it is much less visible; currently, panning stops when you drag the mouse over the minimap.

zifgu:
Note: The drag select class we are currently using has some strange behaviour. For example, before ea4c667, the drag select would start off properly disabled, but if you clicked a node, drag selection would be permanently on even when the select tool is off. This could be because nodes are considered selectable items, so this somehow re-enabled the drag select.

I don't know what is causing this and couldn't find a way to prevent it, so in ea4c667, I resorted to re-initializing the drag select every time the select tool is enabled. 

It is possible that upgrading to the latest version of the dragselect package (https://www.npmjs.com/package/dragselect) can solve this.

Bitajkr:
- [ ] I imagine you're already aware of this, but none of the features in the toolbar are working.

Bitajkr:
If I add a new node using the "plus" button in the authoring toolbar, and then change the level to level 2 and then try to move the node around, the page starts to glitch and two nodes appear (see video). 

Video on Notion as it's too big: https://www.notion.so/tapestrytool/Dev-Meeting-December-8-2022-1-2-pm-e1345ee40c6a48f08ec51143f38e5c7f#a536497d1aec4404b5ca774590e281d6

legendword:
> If I add a new node using the "plus" button in the authoring toolbar, and then change the level to level 2 and then try to move the node around, the page starts to glitch and two nodes appear (see video).
> 
> Video on Notion as it's too big: https://www.notion.so/tapestrytool/Dev-Meeting-December-8-2022-1-2-pm-e1345ee40c6a48f08ec51143f38e5c7f#a536497d1aec4404b5ca774590e281d6

Fixed in https://github.com/tapestry-tool/tapestry-wp/pull/1227/commits/f3c2e93ff5bd763306d5efa8b0c9490369b9be16!

PR #1229, opened 2022-08-05
1228 - Bug fix: Lightbox height becomes 0 for some H5P nodes

### LINKED ISSUE #1228: Bug: Lightbox height becomes 0 for some H5P nodes
The issue is that when viewing some H5P nodes, the lightbox height is being set to 0. This turns out to be due to the H5P embed having some other element instead of the content as the first element in the iframe body. Adjusting the algorithm to look for `.h5p-content` (while still falling back to the previous behaviour of first element in the iframe body when the `.h5p-content` element is not found) and adding a guard to only update dimensions if not trying to set height to 0 will effectively fix the bug.

legendword:
## Changes
* Adjust the h5p height detection algorithm to look for the `.h5p-content` element (while still falling back to the previous behaviour of first element in the iframe body when the `.h5p-content` element is not found)
* Add a guard so that lightbox dimensions will only be set if not trying to set the height to 0.
* Additional bug fix: `console.error("Cannot scroll to non-existing element ", el)` was placed in the wrong conditional, causing the error to be printed to console every time the sidebar is opened.

## Screenshot
N/A
## Issue Linkage
Closes #1228 
## PR Dependency
Depends on: N/A
## Automated Testing
N/A

### DIFF
--- a/templates/vue/src/components/Lightbox/index.vue
+++ b/templates/vue/src/components/Lightbox/index.vue
@@ -272,7 +272,10 @@ export default {
       }
     },
     updateDimensions(dimensions) {
-      if (dimensions.height <= this.lightboxDimensions.height) {
+      if (
+        dimensions.height > 0 &&
+        dimensions.height <= this.lightboxDimensions.height
+      ) {
         this.dimensions = {
           ...this.dimensions,
           ...dimensions,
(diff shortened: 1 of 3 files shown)


zifgu (approved):
Tested this fix in debug mode on the site that was experiencing the issue. The H5Ps indeed disappear on master, but are properly sized on this branch. H5Ps inside Multi-Contents and other content types don't seem affected based on a general test. The code changes also make sense to me.

wynnset (approved):

PR #1232, opened 2022-08-11
1230 - Merge Kaltura upload into TYDE

### LINKED ISSUE #1230: Merge Kaltura upload into TYDE

zifgu:
## Changes
* Merges Kaltura upload and Kaltura player functionality into the TYDE branch. Specifically, uses a new branch that squashes the changes #1169 makes on top of master into a single commit, and merges it into this branch.
* Fixes warnings caused by updated packages coming from #1169
* Fixes all occurrences of `vue/no-mutating-props` in TYDE-specific code, in accordance with #1166.
* Incorporates the recursive setting of `isDyad` in `getNodesDataForRender` into the new traversal algorithm.
* Fixes a bug where TYDE Youth and Dyad roles are not added upon plugin activation.
* Fixes failing tests as a result of the merge.

**Changes parallel to #1169**:
- Fixes bugs that occur on sites where uploads are organized into month and year-based folders:
     - Videos in folders from earlier months should now be included in the list of uploadable videos.
     - The local path should now be computed correctly from the video URL for videos in folders from earlier months.
- No longer clears the Kaltura ID and media URL when "Use Kaltura Player" is checked or unchecked; instead switches video type based on the node's `mediaFormat`
- Save the partner ID, service URL, and unique configuration for each Kaltura video in `typeData.kalturaData`, and show Kaltura media player depending on the availability of these fields instead of the current Kaltura status. This prepares us to support different Kaltura accounts in the future.

## Issue Linkage
Closes #1230 
## PR Dependency
Depends on: N/A (`tyde2` branch)
## Automated Testing
N/A


zifgu:
[Note]

According to this comment, children of dyad nodes should be marked as dyad themselves.
https://github.com/tapestry-tool/tapestry-wp/blob/0dc4d8db9a64e5d12f25b22392d431a0fa6a2da3/classes/class.tapestry.php#L460

However, after moving the logic for setting `isDyad` into the new traversal algorithm introduced in bc5a85540ba3643c5fe2305b1f8d4354153c9238, I noticed that children of a dyad node may not be marked as dyad if the child has multiple parents. If the child is not dyad to begin with and is visited once through a non-dyad parent, it cannot be visited a second time, even if it does have a dyad parent.

The [old code](https://github.com/tapestry-tool/tapestry-wp/blob/0dc4d8db9a64e5d12f25b22392d431a0fa6a2da3/classes/class.tapestry.php#L442-L491
) seems to have this problem too, so as long as this PR introduces no new bugs, I think it should be OK.

Still, is the current behaviour intended?

legendword (approved):
Everything looks good to me. Below are some potential changes we can make, but they should not be part of this PR and are not a priority.

I noticed the spacing is off when the user settings button isn't the only child. Seems like the css selector is applied to the button which is always the only child of its parent div. Perhaps we want to fix that? (minor styling issue, unrelated to this PR)

image

Also that there are two versions of AddConnectionForm.vue under CircleOfSupport, I believe only the one in the ConnectionsTab subfolder is being used. Again unrelated to this PR, just saw that you made changes to both the components but only one is actually being used.

Lastly, by clicking on the "circles" button on the top left of the COS page, I arrived at this view which covers all the UI below it, making it impossible to exit this view without refreshing the page.

image

zifgu:
> Below are some potential changes we can make, but they should not be part of this PR and are not a priority.

Sounds good. Can confirm that these issues were also occurring locally on my end.

zifgu:
The number of available video resolutions has been increased on the Kaltura player for TYDE.

For instructions on how to change the available video resolutions in the Kaltura media player, see this Notion page: https://www.notion.so/tapestrytool/Kaltura-Integration-5189419bda4b4448823e46c85270b345

wynnset:
Superceded by #1283

PR #1199, opened 2022-06-28
1198 - Tapestry 3.0 - Better accessibility

legendword:
#### Styling

* Remove `outline: none;` CSS styles such that outlines for keyboard navigation users shows up for better accessibility.
* Elements should have `:focus` styles similar to their `:hover` styles, if they have any.

#### Descriptive text for screen readers

* All interactive elements have descriptive text labels for screen readers (`aria-label`), and should be focusable and announced as the correct type of element (by explicitly setting the `role` attribute on some interactive elements).
* Describe modals gracefully and ensure everything in the modal can be accessed by the screen reader.
* Describe nodes gracefully and provide information on how to navigate the nodes for screen readers via `aria-label`.
  * Example text description for a node named Introduction: "Introduction. You are on a level 1 node. This node has 3 children. To view this node, press Enter. To go to its first child, press the Down Arrow Key. To go back up, press the Up Arrow Key. To go to its siblings, press the Left or Right Arrow Key."
* Use `aria-live` for error messages so that they are immediately announced by the screen reader.

#### Focus management

* Improve tabbing behaviour when modal is open so that focus is trapped inside the modal; tabbing beyond the last focusable element will return the focus to the first focusable element and vice versa.
* Focus will return to the element that activated the modal upon closing.
* When a button that adds some elements onto the view is pressed (such as the "Add Condition" button in NodeModal access tab, and the "Add Question" button when editing an Activity node), the newly added element should receive focus automatically.
* Use the [roving tabindex method](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_roving_tabindex) for managing focus within the RichTextForm toolbar. When on an item in the toolbar, press Left / Right arrow keys to navigate to other menu items as outlined in the [W3 WAI-ARIA toolbar keyboard interaction guide](https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/#keyboard-interaction-22).
* Also use the [roving tabindex method](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_roving_tabindex) for the main Tapestry view such that only one node or link is focusable at a time. This makes it easier to navigate into and out of the main Tapestry view, since the user only needs to press Tab **once** to focus on the next element in the document flow.
[...]

PR #1240, opened 2022-08-16
1236 - Tapestry 3.0 - Link-based node navigation

### LINKED ISSUE #1236: Link-based node navigation
## The Problem

The node navigation implemented in #1199 assumes a tree structure, but Tapestries are not restricted to a tree structure. The possibility of multiple parents make the arrow-key-based node navigation a suboptimal solution. With the solution implemented in #1199, a user using just their keyboard cannot navigate to some nodes if one cannot reach that node from the root node following the direction of links.

## The Solution

An alternative way of navigating the nodes in a tapestry by keyboard that follows links to go to other nodes. This includes parent-child links in any direction, or associative links that are not parent-child links (for better compatibility with the new authoring workflow).

legendword:
## Changes
* Implemented link-based node navigation by changing the existing node navigation code in PR #1199.
* While on a node:
  * `Enter` to view the current node.
  * `E` to edit the current node.
  * `Backspace` to go to the previous node (the node that the user was previously on, before navigating to this node).
  * `Q` or `Esc` to leave the main Tapestry view.
* Press `L` while on a node to enter link mode. While in link mode: 
  * `Tab` to go to next link, `Shift + Tab` to go to previous link. These will wrap around when reached the end of the list of links.
  * `Esc` to exit link mode and return to the node the user was on before.
  * `Enter` to follow the link to the other end, reaching a parent / child node, and leaving link mode as a result.
  * `E` to edit the current link.
* The aria-label descriptions for a node and a link are updated to reflect the new method of navigation.

## Screenshot
N/A
## Issue Linkage
Closes #1236 
## PR Dependency
Depends on: #1199 
## Automated Testing
To be added.


legendword:
Note: because links on deeper levels are hidden, they are also inaccessible to the screen reader. With a default depth of 3 and a Tapestry with more than 3 levels, this prevents the user from navigating to any link between deeper levelled nodes, since those have `display: none`. This should be fixed when the "center node on click / focus" feature is implemented in #1191), which would zoom the Tapestry view to the appropriate size that enables the links to be shown.

templates/vue/src/store/actions.js
```diff
@@ -405,68 +405,82 @@ export function setTapestryErrorReporting({ commit }, isEnabled) {
...
+export function setNavLinkMode({ commit, state }, linkMode) {
+  if (linkMode && state.nodeNavigation.currentLinked.type === null) {
+    return false
   }
-  return false
+  commit("setNodeNavigation", {
+    linkMode: linkMode,
+  })
+  return true
 }
```
zifgu, line 424: Would you be able to explain why this function has to return true or false? I see the return value is not being used in our code so far.

zifgu (approved):
Keyboard navigation works smoothly when tested both with and without a screen reader. The aria-label descriptions on nodes and links sound natural and are accurate.
* On a node, opening the lightbox and the NodeModal still work. Pressing L activates the link mode.
* In the link mode, I can use `Tab` and `Shift + Tab` to rotate between all links to/from a node, with wrapping. `Enter` works to go to the other side of the link, and `Esc` works to stay on the current node. `E` opens the LinkModal.
* `Q` and `Esc` exit the Tapestry view and focus the minimap button.

I also experienced the bug you described where links at hidden depths are still read by the screen reader, but this is as expected.

Overall, code and functionality both look great to me!

templates/vue/src/store/actions.js
legendword: Indeed the result is not being used as of now. I just put the return value here in case we want to inform the user that they cannot enter link mode when there are no links attached to the node. If we don't need that functionality then I can remove the return value from this function without any issues.

zifgu: Got it, thanks! Since this PR is not ready for final review yet, I think there's no harm in keeping it, in case it will be used in the future.

PR #1243, opened 2022-08-22
1241 - Persist log of Kaltura uploads

### LINKED ISSUE #1241: Persist log of Kaltura uploads
In this feature, we want to start keeping a permanent record of past Kaltura uploads.

Changes to make:
- Change the existing upload log to be a permanent log. With each new upload, add entries to the end of the log instead of replacing it.
    - This means the same node can appear multiple times in the log (e.g. if uploaded unsuccessfully one or more times, or if the video was changed and re-uploaded).
~~- Add an option to the Settings modal to hide older completed uploads.~~
- Make upload log table paginated, and allow user to choose how many entries to show per page.

zifgu:
## Description
In this PR, we want to start keeping a permanent record of past Kaltura uploads.

Previously, if the user performed 2 uploads (from the same Tapestry), [Entries in upload 1] would be replaced by [Entries in upload 2].
Now, the upload log should show [Entries in upload 2] [Entries in upload 1], i.e. a concatenation of the final logs for each upload, most recent first.

## Changes
* Change the existing upload log to be a permanent log. With each new upload, add entries to the end of the log instead of replacing it.
    * This means the same node can appear multiple times in the log (e.g. if uploaded unsuccessfully one or more times, or if the video was changed and re-uploaded).
* Make upload log table paginated, and allow user to choose how many entries to show per page (options are 10, 25, 50, 100, and all).
    * Log should display in reverse chronological order (most recent first).
    * Timestamps should be expressed in the Wordpress site timezone (set under Settings > General).
* After running "Clean Uploaded Videos", for each updated video, only its latest entry in the upload log will have the upload status changed.

## Screenshot
image

## Issue Linkage
Closes #1241 
## PR Dependency
Depends on: #1169 
## Automated Testing
N/A


legendword (approved):
Tried performing normal uploads, uploads with multiple nodes, uploads with errors, and cleaning uploaded videos. Everything works as expected. The status of the upload is updated correctly, and new upload entries with the same node ID do not override the old entries. Performing "Clean Uploaded Videos" still functions properly. The code looks great, too!

wynnset (approved):

PR #1246, opened 2022-08-25
1245 - Comment on nodes

### LINKED ISSUE #1245: Comment on nodes
Any logged in user with the Read permission can comment on a node. The comment will be added as a Wordpress comment, and it should show up both in the sidebar when selecting the node and in the Comments page of the Wordpress dashboard.

Administrators should be able to perform the same set of actions to the comments from within the Tapestry sidebar as from within the Comments page of the Wordpress dashboard.

When comment moderation is on, the administrators should be able to see the pending comments in the Tapestry sidebar and perform related actions on them.

legendword:
### Sidebar comments section
* Displays comments associated with the current node in the sidebar.
* Allows logged in users with Read permissions to add comments to the current node in the sidebar.
* The Comments section will be hidden for a draft node.

### Comment list behaviour
* Comments posted by the same author within 1 hour will be collapsed (the header containing the name of the author and the time will be hidden), similar to how the review comments behave.
* Unapproved comments (comments held for review) are displayed with a yellow background and are not visible to users who are not the author of the comment or an admin.
* Hovering over the time of a comment or the content of a comment for a brief moment will reveal the precise date and time the comment was posted.
* Hovering over the comment will display actions the user can take for a comment, this may include: Reply, Unapprove, Approve, Spam, Trash. The available actions differ depending on the current user and the comment itself. Admins have access to all actions, while other users only have access to certain actions like Reply and Trash.
* Replies to a comment is shown immediately after the comment itself, in exactly the same way as the normal comment list described above (also supports collapsing of comments) except the spacing is smaller and the list is indented a bit.
* Replies are in chronological order with newest reply first. Replies to a reply does not get indented further (flattened reply list). If someone is replying to another reply, the username of the person they are replying to will be added to indicate that.

### Importing & exporting

* Adds a toggle in the Import / Export modal to choose whether to export comments when exporting the Tapestry. When the toggle is off, comments will not be included in the export file; however, when the toggle is on, comments will be included in the export file and **will always be added to the nodes** when importing.
* When importing comments:
  * Comments made by existing users (matched by username and user ID) will be imported as the exact copy of the original comment --- this includes the comment author's user ID, name, email, and URL.
  * Comments made by nonexistent users will have a user ID of 0, which is the same as when someone comments on a Wordpress post while not logged in. These comments will be imported with the same author name as the original comment but will not have an email or URL associated with it. Note that since the comments section in the sidebar only display the author name and the comment content, these comments will appear indifferent from before. The differences can only be observed when inside the Wordpress dashboard's "Comments" page.
  * The approval status of comments will be replicated, so unapproved comments will still be unapproved and approved comments will still be approved.

### Bug fixes

* Fixes: review comments get reversed every time the sidebar navigation is clicked (this is caused by the use of `reverse()` in the ReviewNotifications component in the Toolbar, which reverses the reviewComments array **in-place**).

## Screenshot

Full view:

image

Comment actions:

image

Reply authoring:

image

Comment replies:

image

## Issue Linkage
Closes #1245 

## PR Dependency
Depends on: N/A

## Automated Testing

Added the following tests in `comments.spec.js`:

* Add a comment as an admin and see the comment show up, even when not logged in. Also check that users who are not logged in should not be able to add comments or perform actions to comments.
* Add a comment as a subscriber and see the comment show up as the author themselves or as an admin (since a comment made by a subscriber can be held for moderation by Wordpress, they may by default be in an unapproved state and may not be seen by all users).
* Delete a comment made by the user themselves.
* Approve and unapprove a comment as an admin, and check that unapproved comments do not show up for other users.
* Check that the comments section is not visible for draft nodes.


Bitajkr:
- [ ] The node commenting looks great! I just noticed that if you reply to someone's comment, it doesn't look like a reply (see image below). Normally, when you reply to someone's comment, it show's up as a sub-comment of the original comment. 

Screen Shot 2022-09-13 at 9 49 29 AM

Bitajkr:
> Hi @Bitajkr yes we decided to leave this as a future improvement, but if you think it's critical, we can add it into the current PR. We are actually saving which comment they are replying to, so we'd just need to decide how exactly it would appear. The appearance is a bit tricky. For example, what about replies to replies? Would we just infinitely indent them? Because we don't have a lot of width to play with here, it might not look great. I haven't given this a lot of thought further than that, but if you have any ideas on how to solve those issues, let us know!

Hi Aidin! I see what you mean. I think you could make it look like this - what do you think? 
Group 4687

wynnset:
Just confirming what we discussed in person: We will proceed with Bita's idea of nesting all reply comments (or descendents) under each top-level comment, in reverse chronological order (newest replies at the top). Only indent slightly.

PR #1248, opened 2022-08-25
1247 - Override default Kaltura account on individual sites

### LINKED ISSUE #1247: Override default Kaltura account on individual sites
In multi-site (as well as potentially single-site) WordPress setups, owners of a particular site may wish to override the Kaltura configuration variables in `wp-config.php`, which apply to all sites, with a different Kaltura account.

- Add a new section under Settings > Tapestry admin page to enter a custom set of Kaltura credentials. 
- When determining whether Kaltura is enabled, creating Kaltura sessions, and passing variables to the frontend, look for a full set of 4 configuration variables in the overrided settings first. If one or more of the variables is not provided, then fall back to the default account in `wp-config`.

zifgu:
## Description
In this PR, we allow site administrators to override the default Kaltura configuration variables (set in wp-config.php) with a different Kaltura account on an individual site. This should work for both single-site and multi-site WordPress setups.

## Changes
* Add a new section under Settings > Tapestry admin page where users can enter Kaltura credentials to override the defaults. User can also override the value of `TAPESTRY_KALTURA_UPLOAD_MAX_FILE_SIZE`.
* LOAD_KALTURA is now true if the full set of 4 configuration variables are provided through the Settings admin page, OR if the full set is provided through the constants defined in wp-config.php.
* Whenever the configuration variables are read, uses the overridden configuration variables if the full set of 4 are available; otherwise, fall back to the default variables.
* If `TAPESTRY_KALTURA_UPLOAD_MAX_FILE_SIZE` is provided via the Settings admin page, uses that value instead to check video file sizes. Otherwise, falls back to the constant defined in wp-config.php.

## Screenshot
image

## Issue Linkage
Closes #1247 
## PR Dependency
Depends on #1169 
## Automated Testing
N/A

## Testing
If you wish to test this PR with multisite, follow the instructions from WordPress here to set up a multisite network with your local WordPress installation: https://wordpress.org/support/article/create-a-network/

Before creating the network, you will have to deactivate all plugins, so make sure to re-activate the Tapestry plugin.


legendword (approved):
I have tested the changes in a single-site Wordpress setup, and the constants are overridden only when all four configuration values are set in the Settings > Tapestry page. The last field (max. upload size) is effective regardless of whether the other fields are set, and the max. upload size is overridden whenever the field is set in the site settings.

I was not able to create a multi-site on my local dev environment because my site is running on port :8888, which apparently makes it impossible to create a multi-site according to Wordpress. However, I looked through the code carefully and can confidently say that it for sure will work in a multi-site setup. The code changes are straightforward and easy to follow. I think everything looks good.

wynnset (approved):

PR #1252, opened 2022-08-31
1251 - Bug fix: H5P lightbox height is not adjusted after first height change

### LINKED ISSUE #1251: Bug: H5P lightbox height is not adjusted after first height change
There is an issue with the lightbox for H5P nodes: when the height of the H5P changes the first time, the lightbox height is adjusted to fit the content, but this does not happen when the height of the H5P changes again.

legendword:
## Changes
* Changes the condition for updating the dimensions of the lightbox such that dimensions will still update when given height is bigger than `lightboxDimensions.height` (the initial lightbox dimensions height determined by the aspect ratio); the updated height just takes the minimum of the two, to ensure that the new height will never exceed the initial height.

## Screenshot
N/A
## Issue Linkage
Closes #1251
## PR Dependency
Depends on: N/A
## Automated Testing
N/A


zifgu (approved):
Tested using debug mode on the Tapestry experiencing the issue, and the lightbox height is now increased properly when the H5P content height increases. Also, the lightbox does not exceed the height dictated by the browser aspect ratio when the H5P content is tall. The fix looks good to me.

wynnset (approved):

PR #1259, opened 2022-09-27
1258 - Improve permission checks in the backend

### LINKED ISSUE #1258: Improve permission checks in the backend
We want the backend to be doing the same checks as the frontend so that potential malicious users cannot perform some restricted actions by calling backend APIs directly. To do this, we need to change the `TapestryHelpers::userIsAllowed` method in the backend to align with the `Helpers.hasPermission` method in the frontend.

We also want to check "add" permissions on the parent node whenever someone is trying to publish a child node. This is currently not possible since the backend does not receive information about the parent node when adding a node. We want to keep the current endpoint for adding a node and let it be called whenever someone is making a standalone node (in which case there is no parent node), but add a new endpoint for adding child nodes. This will also help with the undo / redo functionality because the child node and the link will be created together in one endpoint call instead of separately.

legendword:
## Changes
* Modify checks in `TapestryHelpers::userIsAllowed` (backend) to align with `Helper.hasPermission` (frontend).
* Change the action names used for node permission checks to lower case to match the frontend action names.
* Introduce `UserActions` constant class for backend which is a replica of the `userActions` constant in the frontend.
* Add "move" to the list of user actions.
* Fix a bug where admins cannot directly publish child nodes to a rejected node due to not passing `showRejected` value in the calls to `Helpers.hasPermission` in node modal.
* When adding a child node, the frontend now only calls the addNode endpoint instead of both the addNode and addLink endpoints; the addNode endpoint in the backend now takes care of adding the link from parent to child as well as doing some permission checks before adding the child node.
* No longer conditionally update childOrdering of parent node when adding a child node. The childOrdering of parent node is always updated when adding a child node.
* Always show multi-content child nodes and hidden nodes to the admin, even when the admin does not have edit permission on those nodes (this can happen for draft nodes submitted by other users).
* Hide the "edit" button in the SubItemTable used for showing video popups and multi-content child nodes when the user does not have edit permission (it's much better to hide the button instead of alert the user that they do not have edit permission when they click on the button) (again, this can happen for draft nodes submitted by other users).

## Issue Linkage
Closes #1258 
## PR Dependency
Depends on: #1222


legendword:
New change: The "Edit" button will not show up when the user has no edit permission on that subitem (this applies to a draft node submitted by other users, such as the "Draft A" and "Draft A Private" nodes in the screenshot):

image

wynnset:
@legendword I think this should be rebased on top of master since it is not 3.0-related and should be available in 2.0 as well! Could you please rebase it?

legendword:
This pull request is replaced by #1303 which is based on the master branch.

PR #1274, opened 2022-11-09
1273 - Disable moving nodes for users without permission

### LINKED ISSUE #1273: Disable moving nodes for users without permission
- Users without move permission on a node cannot move that node at all.
- Add a tapestry settings field to allow users to move a node around without move permission (move is not saved), which is the old behaviour.

legendword:
## Changes
* Users without move permission on a node cannot move that node at all.
* Add a tapestry settings field called `allowMovingAllNodes` to allow users to move a node around without move permission (move is not saved), which is the old behaviour.
## Screenshot
* Todo
## Issue Linkage
Closes #1273 
## PR Dependency
Depends on: #1259 
## Automated Testing
* N/A


wynnset:
@legendword similar to #1258, let's also rebase this so it's available to 2.0. We can rebase this on top of 1258.

legendword:
This pull request is replaced by #1304 which is based on the master branch.

PR #1281, opened 2022-11-14
1280 - Show progress when importing a tapestry

### LINKED ISSUE #1280: Show progress when importing a tapestry
Show the current progress when importing a tapestry, to the admin. This will allow the admin to understand what is happening for the import.

legendword:
## Changes
* Store the current import progress in a meta field of the tapestry post, containing a boolean `inProgress` and a string `message`.
* Update the file upload progress in realtime while when importing a zip file and show a progress bar.
* Fetch the current import progress every 5 seconds while the tapestry is being imported and show the message below the import spinner.
* Delete the H5P cachedassets folder (located in `${Wordpress uploads dir}/h5p/cachedassets` if H5P nodes were imported, to avoid running into potential problems with H5P cachedassets.
* Hide the "add node" button when importing a tapestry.
* Add "Import in progress" next to the spinner inside the button when importing a tapestry, to make it more clear that import is in progress.

## Screenshot
image

image

## Issue Linkage
Closes #1280 
## PR Dependency
Depends on: N/A
## Automated Testing
* N/A


wynnset:
@legendword I believe this PR is ready for testing/review but it doesn't have any labels or assignees. Can you please update that?

legendword:
> @legendword I believe this PR is ready for testing/review but it doesn't have any labels or assignees. Can you please update that?

I will update the description and the screenshot to reflect the newer changes and then assign this to you when done!

wynnset:
@legendword A lot of weird things still happening for me, but I’m not sure if it’s related to the code or the environment.

Locally, I am having trouble testing this branch. Even after restarting MAMP and closing and re-opening my browser, despite using the Chrome dev tools to throttle the upload, it seems as if the file uploads is done instantly, and the server sends me that it's started process the H5P files in the zip file. However, when I look at the response to `import_status`, it tells me that only a small portion of the bytes have been processed (and on subsequent calls to `import_status`, the bytes processed slowly increases).

On the production server, the response from `import_status` never contains the `fileUpload` attribute and after a long time, it just sends back an error 413 ("Entity too large").

legendword:
> @legendword A lot of weird things still happening for me, but I’m not sure if it’s related to the code or the environment.
> 
> Locally, I am having trouble testing this branch. Even after restarting MAMP and closing and re-opening my browser, despite using the Chrome dev tools to throttle the upload, it seems as if the file uploads is done instantly, and the server sends me that it's started process the H5P files in the zip file. However, when I look at the response to `import_status`, it tells me that only a small portion of the bytes have been processed (and on subsequent calls to `import_status`, the bytes processed slowly increases).
> 
> On the production server, the response from `import_status` never contains the `fileUpload` attribute and after a long time, it just sends back an error 413 ("Entity too large").

Regarding the first problem that you encountered locally, that actually is a result of restarting MAMP while an ongoing upload request was not finished. Since an ongoing upload request set the importStatus with a message string, it did not get cleared when the server was forced to shut down (the code is designed such that if an error occurs, the status will be cleared, but not when the server is force quit.) Hence, the message will still be stored there and in the frontend, the status message takes precedence over the import progress. So it is actually working and the throttling makes the bytes processed slowly increase. Once an import finishes normally after restarting MAMP, the importStatus message will get cleared and the next import should work all as expected.

For the second issue, I'm not too sure why that is happening, but the 413 error is related to either the PHP's max file upload size limit or the server (NGINX or Apache) request body max size. If you are sure that the active php.ini configuration uses an upload size limit larger than the file, then perhaps you could try checking the request body size limit for the hosting server.

wynnset:
> the 413 error is related to either the PHP's max file upload size limit or the server (NGINX or Apache) request body max size

I am still puzzled because I've checked that the php post_max_size and upload_max_filesize are both set to 1G, and I've now also added a .htaccess directive of LimitRequestBody 2147483647 which sets the server body max size to 2GB, but still getting the 413 error. This issue is really hard to debug since we can't even see the progress for the file.

I tried uploading a smaller zip file and although the tapestry imported successfully, the file upload progress was still not working and the status message the whole time was "Uploading file...". This means that the file upload progress feature is not working on the server, either, for some reason! I checked the import_status requests and they always returned an object with only the `inProgress` property set to `false`.

wynnset:
Update: I changed the code to use a front-end method for tracking the upload progress. This method doesn't rely on the backend and I tested it on our prod server and it works. Based on what I saw, the upload progress goes all the way up to 100% and that's when the server throws the 413 error. I'm going to contact our hosting provider to see if they know why that's happening and will post an update here once I hear back.

templates/vue/src/components/TapestryMain/RootNodeButton/index.vue
```diff
@@ -143,42 +217,96 @@ export default {
...
+        .catch(err => {
+          this.addApiError(err)
+        })
+        .finally(() => {
+          if (this.isImporting) {
+            setTimeout(
+              () => {
+                this.updateImportStatus()
+              },
+              this.isUploading ? 2000 : 50
```
legendword, line 305: Not sure if the 50 here is intended or not. When testing locally, 50ms is great and the import status updates promptly. However, I'm not sure if we want this short of a timeout between requests to the server.

legendword (commented):
Tested the new user interface and the frontend approach to tracking upload progress, and everything works well. Except the one comment I added below, this PR is ready to be merged.

templates/vue/src/components/TapestryMain/RootNodeButton/index.vue
legendword: After discussion, we have decided to increase the timeout between requests to 2000ms, since 50ms is indeed too small and would cause lots of requests being sent to the server.

wynnset (approved):

PR #1282, opened 2022-11-14
1277 - Bug fix: Kaltura video player has various issues

### LINKED ISSUE #1277: Bug: Kaltura videos don't fit properly inside multi-content nodes

wynnset:
## Changes
* Ensure Kaltura player fits well in multi-content
* Fix issue where the player would not show the user progress and always show it's at 0
* Fix issue where initial user seeking was not working (would always start at 0)
* Fix issue where player would play in the background with pop-up on top if seeked past a pop-up time
## Issue Linkage
Closes #1277 
## Automated Testing
No changes to tests


legendword (approved):
The Kaltura fixes are working and the refactoring looks good. Just waiting to test if loading multiple Kaltura scripts from different Kaltura servers breaks anything (when calling kWidget.embed, the kWidget global variable can only be from the latest loaded kWidget script).

legendword:
> The Kaltura fixes are working and the refactoring looks good. Just waiting to test if loading multiple Kaltura scripts from different Kaltura servers breaks anything (when calling kWidget.embed, the kWidget global variable can only be from the latest loaded kWidget script).

Indeed, after testing, two videos from different Kaltura servers will not be played properly, especially of they are in the same multi-content node. The Kaltura script that is loaded last will overwrite any existing kWidget global variable. Currently we are not planning to fix this, as most Tapestry sites using Kaltura are only using videos from one server. If a situation arises that require more than one Kaltura server to be used, this issue will need to be reconsidered.

legendword:
After reviewing the PR description, there is one item that might have been fixed but is now still a problem:
> - Fix issue where player would play in the background with pop-up on top if seeked past a pop-up time

Some notes after I did a bit of testing (for context, this is a Kaltura video with Kaltura player and a popup node at 10 seconds). Tried seeking from before 10s to after 10s multiple times. The popup does not trigger every time (see below).

Scenario 1: `lastTime` update happens after seeking, meaning the `lastTime` is already updated and it is not considered to have sought over the 10s threshold. The popup does not trigger.
image

Scenario 2: `lastTime` update happens before seeking, so the popup at 10s triggers. The pause command is sent to the Kaltura player, yet the video is not paused.
image

If playing video normally without seeking, the popup always shows up normally and the video is paused normally:
image

PR #1317, opened 2023-05-05
1316 - Improve node completion

### LINKED ISSUE #1316: Improve node completion
Improve the logic and display of completion of nodes

legendword:
## Changes
* Show completion icon in the multi-content page menu and multi-content accordion toggle buttons.
* Show a toast indicating the user's answer has been saved when they submit an activity.
* Fix a bug that prevents child nodes of multi-content rows to be completed.
* Fix a problem in the logic that causes progress of nodes not to be set to 1 when completed.
* Change the color of the node completion icons to white and the color of the inner checkmarks to the page menu background color.

Other changes:
* Show "no answers" text and a logged out alert when submitting an activity while not logged in.
* Show error popup for non-backend errors (e.g. frontend Javascript errors).
* Fix current row not being highlighted in the page menu.

Code refactoring:
* Shorten the code for the `getAnswers` getter method.
* Remove unused component states and methods in multiple components.
* Remove the special case for node titled "Resources" when checking for multi-content completion. This feature should be TYDE-only.

## Screenshot
image

## Issue Linkage
Closes #1316 
## PR Dependency
Depends on: N/A
## Automated Testing
* N/A


templates/vue/src/components/Lightbox/media/MultiContentMedia/MultiContentRows.vue
```diff
@@ -104,12 +104,6 @@ export default {
     complete(rowId) {
       this.$emit("complete", rowId)
     },
-    changeRow(rowInfo) {
-      this.$emit("change-row", { context: this.context, ...rowInfo })
-    },
-    areAllPopup(nodes) {
-      return nodes.every(node => node.popup !== null)
-    },
```
wynnset, line 112: Can you explain why these are removed, and same in the index file too?

legendword: I did a global search for these methods in our code base, and they are only defined but never called. Since they are unused, I just removed them to avoid confusion in the future.

wynnset (approved):

